PR #112, opened 2025-09-04
(chore): Change `CD` workflows to publish only when changed

timjroberts:
This PR updates the `'CD'` workflows to ensure that packages are only published if changes are detected in their associated _product_ folders.


sp-io (approved):

sp-io:
Nice!

PR #125, opened 2025-09-09
feat(midnight-js): Migration to ledger 6 and add Unshielded Tokens

sp-io:
Migrate to latest ledger v6


adamreynolds-io (approved):
Given the scale of the change and the need to get an 3.0.0-alpha.1 into hands of QA I am approving this. I recognise there is technical debt in this PR but that is acceptable.

Really well done.

PR #141, opened 2025-09-16
chore: separate executions for unit tests and property tests

sp-io:
Benefits in CI: 
- faster feedback
- fail fast


sp-io:
Closing for a time being

PR #172, opened 2025-09-29
feat: contracts package separation

sp-io:
divide contracts package into smaller pieces


sp-io:
Closing. Will be used for future reference.

PR #266, opened 2025-10-29
feat: fix esm and cjs packaging

sp-io:
https://shielded.atlassian.net/browse/PM-20246

Problem:

publint complains on packages/* with error: 
```
pkg.exports["."].types types is interpreted as CJS when resolving with the "import" condition. This causes the types to be ambiguous when default importing the package due to its implied interop. Consider splitting out two "types" conditions for "import" and "require", and use the .mts extension, e.g. pkg.exports["."].import.types: "./dist/index.d.mts"
```

Solution:

- Add rollup-plugin-dts for .d.ts bundling
- Generate .d.ts, .d.mts, .d.cts files
- Return array of configs (code + types)
- Support dual package exports pattern


adamreynolds-io (approved):
LGTM

PR #318, opened 2025-11-17
feat: release 3.0.0-alpha.3

sp-io:
Release 3.0.0-alpha.3


timjroberts:
Does this need the API doc files to be re-generated (at least with the updated version number)?

PR #434, opened 2026-01-15
chore: add e2e scoped test for circuit fail

jtsang586:
Add additional test to cover behaviour of scoped tx when circuit call fails


jtsang586:
Error message is confirmed to be a bug and needs to be fixed before this PR is merged.
https://shielded.atlassian.net/browse/PM-21257

PR #470, opened 2026-01-28
feat: add contract address scoping for private state operations

sp-io:
Introduced `setContractAddress` method to scope private state operations, ensuring namespace isolation across contracts. Updated in-memory and level-based private state providers to include scoped key management. Added comprehensive tests to validate scoping behavior and error handling for unset contract addresses.


adamreynolds-io (dismissed):
Recognise this is a draft but this is an elegant solution that reduces the attack surface.

PR #502, opened 2026-02-09
docs: clean up `level-private-state-provider` README

sp-io:
Add risk notice for the private provider


adamreynolds-io (approved):
The wording has been agreed and correctly explains the limitations of the private state *level db* provider.

PR #526, opened 2026-02-18
feat: add signing key export/import APIs and browser storage warning 

sp-io:
## Summary                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                               
  This PR addresses the audit finding that browser storage loss can prevent users from executing private contract operations by adding:                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                               
  - **Signing key export/import APIs** - New `exportSigningKeys()` and `importSigningKeys()` methods to back up and restore signing keys                                                                                                                                                                                                                                                                       
  - **Runtime browser warning** - Console warning shown once per session when using the provider in a browser environment                                                                                                                                                                                                                                                                                      
  - **Comprehensive tests** - 41 new tests covering all export/import scenarios and browser warning behavior                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                               
  ## Changes                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                               
  ### Types (`packages/types/src/`)                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                               
  - Added `SigningKeyExport` interface with format, encryptedPayload, and salt                                                                                                                                                                                                                                                                                                                                 
  - Added `MAX_EXPORT_SIGNING_KEYS` constant (10000)                                                                                                                                                                                                                                                                                                                                                           
  - Added `ExportSigningKeysOptions` and `ImportSigningKeysOptions` interfaces                                                                                                                                                                                                                                                                                                                                 
  - Added `ImportSigningKeysResult` interface                                                                                                                                                                                                                                                                                                                                                                  
  - Added `SigningKeyExportError` class                                                                                                                                                                                                                                                                                                                                                                        
  - Extended `PrivateStateProvider` interface with new methods                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                               
  ### Implementation (`packages/level-private-state-provider/src/`)                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                               
  - Implemented `exportSigningKeys()` with AES-256-GCM encryption (same security as private states)                                                                                                                                                                                                                                                                                                            
  - Implemented `importSigningKeys()` with conflict strategies (error/skip/overwrite)                                                                                                                                                                                                                                                                                                                          
  - Added browser warning that displays once per session via sessionStorage                                                                                                                                                                                                                                                                                                                                    
  - Uses PBKDF2 with 100k iterations for key derivation                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                               
  ### Test Infrastructure                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                               
  - Updated mock in `packages/contracts/src/test/test-mocks.ts`                                                                                                                                                                                                                                                                                                                                                
  - Added full implementation to `inMemoryPrivateStateProvider` in testkit-js                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                               
  ## Security Considerations                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                               
  - Same encryption standard as private states (AES-256-GCM, PBKDF2 100k iterations)                                                                                                                                                                                                                                                                                                                           
  - Single generic error for decryption failures (prevents oracle attacks)                                                                                                                                                                                                                                                                                                                                     
  - Fresh salt per export (different from storage salt)                                                                                                                                                                                                                                                                                                                                                        
  - Password validation (16 character minimum)                                                                                                                                                                                                                                                                                                                                                                 
  - Separate format identifier (`midnight-signing-key-export`) prevents accidental misuse

### DIFF
--- a/packages/level-private-state-provider/src/level-private-state-provider.ts
+++ b/packages/level-private-state-provider/src/level-private-state-provider.ts
@@ -31,8 +31,7 @@ import {
   type PrivateStateId,
   type PrivateStateProvider,
   type SigningKeyExport,
-  SigningKeyExportError,
-  type WalletProvider
+  SigningKeyExportError
 } from '@midnight-ntwrk/midnight-js-types';
 import { type AbstractSublevel } from 'abstract-level';
 import { Buffer } from 'buffer';
@@ -58,7 +57,7 @@ export const MN_LDB_DEFAULT_PRIS_STORE_NAME = 'private-states';
 export const MN_LDB_DEFAULT_KEY_STORE_NAME = 'signing-keys';
 
 /**
- * Optional properties for the indexedDB based private state provider configuration.
+ * Configuration properties for the LevelDB based private state provider.
  */
 export interface LevelPrivateStateProviderConfig {
   /**
@@ -74,30 +73,20 @@ export interface LevelPrivateStateProviderConfig {
    */
   readonly signingKeyStoreName: string;
   /**
-   * Wallet provider used to get the encryption public key for password derivation.
-   * If privateStoragePasswordProvider is not provided, the wallet's encryption public key
-   * will be used as the password.
-   */
-  readonly walletProvider?: WalletProvider;
-  /**
    * Provider function that returns the password used for encrypting private state.
    * The password must be at least 16 characters long.
    *
-   * If not provided, defaults to using walletProvider.getEncryptionPublicKey().
+   * SECURITY: Use a strong, secret password. Never use public key material
+   * or other non-secret values as the password source.
    *
    * @example
    * ```typescript
-   * // Using default (wallet's encryption public key)
-   * { walletProvider: wallet }
-   *
-   * // Using custom password provider
    * {
-   *   walletProvider: wallet,
-   *   privateStoragePasswordProvider: async () => await getUserPassword()
+   *   privateStoragePasswordProvider: async () => await getSecretPassword()
    * }
    * ```
    */
-  readonly privateStoragePasswordProvider?: PrivateStoragePasswordProvider;
+  readonly privateStoragePasswordProvider: PrivateStoragePasswordProvider;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/level-private-state-provider/src/level-private-state-provider.ts
```diff
@@ -588,6 +665,148 @@ export const levelPrivateStateProvider = <PSI extends PrivateStateId, PS = any>(
...
+        if (existingKey !== null) {
+          if (conflictStrategy === 'skip') {
+            skipped++;
+            continue;
+          } else if (conflictStrategy === 'overwrite') {
+            overwritten++;
+          }
+        }
+
+        await this.setSigningKey(address, signingKey);
```
adamreynolds-io: Suggestion: No validation of the `signingKey` value before storing. A crafted export could inject arbitrary values that would be accepted and persisted. 

This is consistent with `importPrivateStates` (which also doesn't validate state values), but signing keys are more security-sensitive -- a corrupted key stored silently could cause hard-to-diagnose transaction failures later.

Consider adding a basic structural check before storing, e.g. verifying `signingKey` is not null/undefined and has the expected shape. This would turn a silent data corruption into an immediate, actionable error.

packages/level-private-state-provider/src/level-private-state-provider.ts
```diff
@@ -289,6 +306,64 @@ const validateSalt = (salt: string): void => {
...
+      if (storage.getItem(BROWSER_WARNING_KEY)) {
+        return;
+      }
+      storage.setItem(BROWSER_WARNING_KEY, 'true');
+    }
+  } catch {
+    // sessionStorage may be unavailable in some contexts
+  }
+
+  console.warn(
```
adamreynolds-io: Should this be debug/dev warning? It's also explicitly providing implementation details.

adamreynolds-io (changes requested):
Mainly to do with that console warning. How a developer uses the provider should be down to them I think. One to discuss!

packages/level-private-state-provider/src/level-private-state-provider.ts
sp-io: @timjroberts thoughts on this? I am unsure what's the optimal option here.

PR #534, opened 2026-02-20
feat: implement consistent salt generation to prevent race conditionsi n private state operations

sp-io:
Added `getOrCreateSalt` utility to synchronize salt initialization for private states and signing keys. Introduced tests to verify salt consistency across sequential operations, after clear operations, and in mixed use cases. Updated relevant `.gitignore` entries.


adamreynolds-io (approved):
Nicely done.

PR #558, opened 2026-02-24
fix: Attach unshielded offers for user-addressed claim unshielded spends

JosephDenman:
This PR is motivated by [this](https://github.com/LFDT-Minokawa/compact/issues/151) issue that was reported in the Compact repository.

When a Compact contract calls `sendUnshielded` to transfer tokens to a user's wallet address, the resulting transaction must include `UtxoOutput` entries in the intent's `UnshieldedOffer`. Without these entries, the ledger rejects the transaction during validation. This change extracts the required outputs from the transcript's effects and attaches them to the correct section (guaranteed or fallible) of the intent before constructing the transaction.

When a contract calls `sendUnshielded(tokenType, amount, userAddress)`, two things happen during circuit execution. The contract's effects record the intent to send tokens: `effects.unshieldedOutputs` tracks the total amount per token type, and `effects.claimedUnshieldedSpends` records each individual spend with its recipient address and amount.
The ledger later validates the transaction by computing `real_unshielded_spends` — the set of actual token movements — and checking that every claimed spend is backed by a real one. For sends to user addresses, "real" means there's a matching `UtxoOutput` entry in the intent's `UnshieldedOffer`. For sends to contract addresses, the recipient contract's own `receiveUnshielded` effects serve as proof instead, so no offer output is needed.


adamreynolds-io (changes requested):
Good change but needs tests.

adamreynolds-io (dismissed):

adamreynolds-io (approved):

PR #633, opened 2026-03-13
feat: [PM-22110] Flow `LedgerParameters` like public state data

timjroberts:
[PM-22110](https://shielded.atlassian.net/browse/PM-22110)

With this PR, `LedgerParameters` are retrieved from the block associated with the public contract state when querying for a contract's Zswap chain state and public state, and then used to initialize the call options when setting up the next circuit call.

The end result of this change is that we no longer default to `LedgerParameters.initialParameters()` when calling circuits.

[PM-22110]: https://shielded.atlassian.net/browse/PM-22110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/http-client-proof-provider/src/test/commons.ts
+++ b/packages/http-client-proof-provider/src/test/commons.ts
@@ -24,10 +24,13 @@ import {
   emptyZswapLocalState,
   sampleSigningKey} from '@midnight-ntwrk/compact-runtime';
 import {
+  LedgerParameters,
   sampleCoinPublicKey,
   sampleContractAddress,
   sampleEncryptionPublicKey,
-  type UnprovenTransaction,  ZswapChainState} from '@midnight-ntwrk/ledger-v8';
+  type UnprovenTransaction,
+  ZswapChainState
+} from '@midnight-ntwrk/ledger-v8';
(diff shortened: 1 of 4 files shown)


packages/types/src/public-data-provider.ts
```diff
@@ -113,7 +114,7 @@ export interface PublicDataProvider {
   queryZSwapAndContractState(
     contractAddress: ContractAddress,
     config?: BlockHeightConfig | BlockHashConfig
-  ): Promise<[ZswapChainState, ContractState] | null>;
+  ): Promise<[ZswapChainState, ContractState, LedgerParameters] | null>;
```
adamreynolds-io: 🤖 AI-assisted review

This is a breaking change for external implementors of `PublicDataProvider` (2-tuple → 3-tuple). Should be called out in release notes.

Side note: the tuple is getting long — three positional elements where order matters is fragile. A named object would be safer, but that's a larger refactor outside this PR's scope.

adamreynolds-io (commented):
Inline comments on specific locations (AI-assisted 🤖). See the summary review comment for the full checklist.

PR #648, opened 2026-03-17
chore: use add calls to build transaction

sp-io:
Use Transaction add calls to construct Transaction

https://shielded.atlassian.net/browse/PM-21556?search_id=dd39553b-e248-4533-af64-54448a777c73

The change is in how createUnprovenLedgerCallTx assembles the transaction.                                                                                                                                                                                                                                                                                                                                   
                                                   
  Current (after reverts):                                                                                                                                                                                                                                                                                                                                                                                     
  - Uses ContractCallPrototype + Intent.addCall() — requires already-partitioned transcripts (PartitionedTranscript = [guaranteed, fallible])                                                                                                                                                                                                                                                                  
  - Manually creates UnshieldedOffer via extractUserAddressedOutputs for user-addressed spends                                                                                                                                                                                                                                                                                                                 
  - Signature: (circuitId, contractAddress, initialContractState, zswapChainState, partitionedTranscript, privateTranscriptOutputs, input, output, nextZswapLocalState, encryptionPublicKey)                                                                                                                                                                                                                   

  Target:
  - Uses PrePartitionContractCall + Transaction.addCalls() — takes pre-partition transcripts (Op<AlignedValue>[] = raw publicTranscript)
  - addCalls() internally handles transcript partitioning and UnshieldedOffer creation — no manual work needed
  - PreTranscript requires a LedgerQueryContext — created via createCircuitContext() from compact-runtime, which reproduces the exact same context used during circuit execution
  - Signature: (circuitId, contractAddress, initialContractState, zswapChainState, publicTranscript, privateTranscriptOutputs, input, output, nextZswapLocalState, encryptionPublicKey, ledgerParameters, coinPublicKey)

  What changes in the caller (unproven-call-tx.ts):
  - Passes raw publicTranscript instead of partitionedTranscript
  - Adds ledgerParameters and coinPublicKey as new parameters
  - Removes extractUserAddressedOutputs and manual offer construction

  Why:
  - addCalls() is the newer, correct API — it manages partitioning and offers internally
  - Eliminates bugs from manually building UnshieldedOffer (e.g., error 139 in e2e tests with NIGHT token transfers)
  - createCircuitContext guarantees the QueryContext in PreTranscript has the exact same state path encoding as the one used during circuit execution (solves the [-] vs [01] encoding mismatch)

### DIFF
--- a/packages/contracts/src/test/utils/ledger-utils.test.ts
+++ b/packages/contracts/src/test/utils/ledger-utils.test.ts
@@ -17,6 +17,7 @@ import {
   type AlignedValue,
   ContractOperation,
   ContractState as CompactContractState,
+  type Op,
   QueryContext
 } from '@midnight-ntwrk/compact-runtime';
 import {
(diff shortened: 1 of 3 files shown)


adamreynolds-io (changes requested):
Migration from `ContractCallPrototype` + `Intent.addCall()` to `PrePartitionContractCall` + `Transaction.addCalls()` looks correct — parameter threading is right, `createCircuitContext` usage is sound, `extractUserAddressedOutputs` removal is clean with no remaining references, and no public API breakage.

One issue: the test coverage dropped significantly. See inline comment.

timjroberts (commented):
The changes looks sound as far as I can see. I agree with @adamreynolds-io in that we likely need some equivalent test coverage over the `publicTranscript` parameter.

adamreynolds-io (approved):

PR #711, opened 2026-03-24
fix: prevent command injection in compact CLI tools

sp-io:
## Summary

- **`fetch-compact.mts`**: Replaced `execSync` / `exec` with template literal interpolation with `spawnSync` / `spawn` using argument arrays for `unzip`, `chmod`, and `docker pull` commands
- **`run-compactc.cjs`**: Replaced `exec()` with string-joined docker command with `childProcess.spawn()` using a proper argument array; removed unused `exec` import

## Problem

Both `fetch-compact.mts` and `run-compactc.cjs` construct shell commands by interpolating variables (version strings, file paths, CLI arguments) directly into template literals passed to `execSync` / `exec`. This is vulnerable to command injection if the interpolated values contain shell metacharacters (`;`, `|`, `&`, `` ` ``).

For example, a `COMPACTC_VERSION` value of `1.0; rm -rf /` would be executed as:
```
unzip /path/compactc-1.0; rm -rf /.zip -d /path/1.0; rm -rf /
```

While these are developer build tools (not runtime services), they pose a risk in CI/CD pipelines where version strings may originate from external inputs (PR metadata, config files, etc.).

## Fix

Replace all `execSync`/`exec` with template literal arguments with `spawnSync`/`spawn` using argument arrays. `spawn`/`spawnSync` bypass the shell entirely, passing arguments directly to the target process, which eliminates command injection.

Notably, the non-docker path in `run-compactc.cjs` already used the safe `spawn()` pattern — this PR aligns the docker path to match.

## Test plan

- [ ] Verify `fetch-compactc` still downloads and extracts compactc correctly
- [ ] Verify `run-compactc` docker path still executes compactc via docker
- [ ] Verify `run-compactc` native path still works (unchanged)
- [ ] `yarn lint` passes
- [ ] `yarn --cwd packages/compact build` succeeds

### DIFF
--- a/packages/compact/src/fetch-compact.mts
+++ b/packages/compact/src/fetch-compact.mts
@@ -133,8 +133,8 @@ const fetchCompact = async (): Promise<void> => {
   fs.rmSync(targetCompactDir, { force: true, recursive: true });
   fs.mkdirSync(targetCompactDir, { recursive: true });
 
-  childProcess.execSync(`unzip ${targetFile} -d ${targetCompactDir}`);
-  childProcess.execSync(`chmod -R +w ${targetCompactDir}`);
+  childProcess.spawnSync('unzip', [targetFile, '-d', targetCompactDir], { stdio: 'inherit' });
+  childProcess.spawnSync('chmod', ['-R', '+w', targetCompactDir], { stdio: 'inherit' });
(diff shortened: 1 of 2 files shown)


adamreynolds-io (approved):
LGTM

PR #798, opened 2026-04-08
feat: web crypto migration

### LINKED ISSUE #828: Enable cross-platform crypto support in level-private-state-provider (browser, mobile, Node.js)
## Problem

`level-private-state-provider` directly imports Node.js `crypto` (`createCipheriv`, `createDecipheriv`, `pbkdf2Sync`, `randomBytes`) in its runtime code. This makes the package unusable in any non-Node.js environment:

- **Browsers** — bundlers fail on `crypto` resolution or externalize the module, breaking the provider before it can even be exercised
- **React Native** — Hermes/JSC don't ship Node.js `crypto` at all
- **WebView-based mobile apps (Capacitor, Ionic)** — same bundler issue as browsers

Any downstream consumer targeting a browser or mobile build pipeline hits this incompatibility.

### Step 1: Migrate from Node.js `crypto` to Web Crypto API

Replace `createCipheriv`/`createDecipheriv`/`pbkdf2Sync`/`randomBytes` with async `globalThis.crypto.subtle` equivalents. This makes the package work in browsers and Node.js ≥15.

Breaking change: all crypto operations become async (`StorageEncryption` constructor replaced with `StorageEncryption.create()` factory).

### Step 2: Add a `CryptoBackend` abstraction with noble fallback

Web Crypto solves browsers but not React Native (Hermes/JSC don't have `crypto.subtle`). Extract crypto primitives behind a `CryptoBackend` interface with two implementations:

- **`WebCryptoCryptoBackend`** — hardware-accelerated, uses `globalThis.crypto.subtle`
- **`NobleCryptoBackend`** — pure JavaScript (`@noble/hashes` + `@noble/ciphers`), works everywhere

Auto-detect at runtime: prefer Web Crypto if available, fall back to noble. Consumer can override via `cryptoBackend: 'webcrypto' | 'noble'` option.

Both backends produce standard AES-256-GCM in the same binary format — data encrypted by one is decryptable by the other. V1 backward compatibility (100k iteration PBKDF2 data) preserved.

## Scope

- Remove all `import ... from 'crypto'` from the package
- New async API for `StorageEncryption` (breaking change)
- New files: `crypto-backend.ts`, `crypto-backend-webcrypto.ts`, `crypto-backend-noble.ts`
- New dependencies: `@noble/ciphers@^2.0.0`, `@noble/hashes@^2.0.0`
- Optional `cryptoBackend` field on `LevelPrivateStateProviderConfig`
- Cross-backend compatibility and V1 backward compat test coverage

## Out of scope

- Full browser packaging for `level`/`abstract-level` stack (`events` polyfill still needed downstream)
- React Native storage backend (IndexedDB not available in RN — separate concern)

## Acceptance criteria

- [ ] No Node.js `crypto` imports remain in the package
- [ ] Package bundles without errors in browser build pipelines (Vite, Webpack)
- [ ] Works in environments without `crypto.subtle` via noble fallback
- [ ] Cross-backend interop: encrypt with one, decrypt with the other
- [ ] V1 encrypted data decryptable under both backends
- [ ] Existing test suite passes with both backends

sp-io:
Resolve: https://github.com/midnightntwrk/midnight-js/issues/828

Migrate level-private-state-provider from Node.js crypto to Web Crypto API (globalThis.crypto.subtle).
                                                                                                                                                                                                   
  - Replace createCipheriv/createDecipheriv/pbkdf2Sync/randomBytes with async Web Crypto equivalents                                                                                               
  - Convert StorageEncryption constructor to async factory (StorageEncryption.create())                                                                                                            
  - Add assertWebCryptoAvailable() guards with clear error messages for non-secure contexts                                                                                                        
  - Fix isDecryptionError for cross-platform Web Crypto support (OperationError DOMException)
  - Cache getScopedNames promise with rejection recovery to prevent unrecoverable state                                                                                                            
  - Preserve V1 backward compatibility (old encrypted data still decryptable)
                                                                                                                                                                                                   
  Breaking changes                                                 
              
  All crypto operations are now async. StorageEncryption constructor replaced with await StorageEncryption.create().

### DIFF
--- a/packages/level-private-state-provider/src/level-private-state-provider.ts
+++ b/packages/level-private-state-provider/src/level-private-state-provider.ts
@@ -336,7 +336,8 @@ const isDecryptionError = (error: unknown): boolean => {
     message.includes('salt mismatch') ||
     message.includes('invalid encrypted data') ||
     message.includes('bad decrypt') ||
-    message.includes('unable to authenticate')
+    message.includes('unable to authenticate') ||
+    message.includes('operation failed for an operation-specific reason')
   );
 };
 
--- a/packages/level-private-state-provider/src/level-private-state-provider.ts
+++ b/packages/level-private-state-provider/src/level-private-state-provider.ts
@@ -38,7 +38,7 @@ import { Buffer } from 'buffer';
 import { Level } from 'level';
 import * as superjson from 'superjson';
 
-import { decryptValue, getPasswordFromProvider, type PrivateStoragePasswordProvider, StorageEncryption } from './storage-encryption';
+import { assertWebCryptoAvailable, decryptValue, getPasswordFromProvider, type PrivateStoragePasswordProvider, StorageEncryption } from './storage-encryption';
 
 /**
  * The default name of the indexedDB database for Midnight.
@@ -133,6 +133,7 @@ superjson.registerCustom<Buffer, string>(
 const ACCOUNT_ID_HASH_LENGTH = 32;
 
 const hashAccountId = async (accountId: string): Promise<string> => {
+  assertWebCryptoAvailable();
   const data = new TextEncoder().encode(accountId);
   const hashBuffer = await globalThis.crypto.subtle.digest('SHA-256', data);
   return Buffer.from(new Uint8Array(hashBuffer)).toString('hex').substring(0, ACCOUNT_ID_HASH_LENGTH);
@@ -330,6 +331,11 @@ interface RotateStorePasswordParams {
 
 const isDecryptionError = (error: unknown): boolean => {
   if (!(error instanceof Error)) return false;
+
+  if ('name' in error && error.name === 'OperationError') {
+    return true;
+  }
+
   const message = error.message.toLowerCase();
   return (
     message.includes('unsupported state') ||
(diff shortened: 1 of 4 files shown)


timjroberts:
## Code Review

Overall this is a well-executed migration — clean async factory pattern, V1 backward compatibility preserved, good test coverage, and CI is fully green. A few suggestions:

### Medium

1. **`timingSafeEqual` — pure-JS only.** _This was added by myself in a previous commit, but wondering whether there is now an alternative available in the new libraries?_

2. **`console.warn` in `decryptValue` for unencrypted pass-through** could be very noisy during migration scenarios since it fires for *every* unencrypted value read. Consider `console.debug` or a one-time warning with deduplication.

3. **`hashPassword` uses a single SHA-256** for `verifyPassword`. This is fine since the hash only lives in the runtime instance (never persisted), but a code comment explaining why a fast hash is acceptable here would help future maintainers.

### Minor

4. **`Buffer.from(this.salt).equals(salt)` in `decrypt`/`decryptWithPassword`** creates a `Buffer` copy on every call. Could use `timingSafeEqual(Buffer.from(this.salt), salt)` for consistency and constant-time comparison.

5. **`invalidateEncryptionCache` swallows errors silently** — if the old cached promise was rejected for a real reason, stale cache entries may never get cleared. A `console.warn` on the caught error would aid debugging.

6. **Missing `assertWebCryptoAvailable()` in `isEncrypted`/`getVersion`** — these don't need Web Crypto themselves, but if they return `true` and then `create`/`decrypt` throws, the caller gets a confusing two-step failure. Consider documenting this.

7. **`setSigningKey` → `signingKeyValue` rename** — an alternative is renaming the destructured binding instead, keeping the public-facing parameter name stable.

### Breaking Changes

The async migration is well-documented in the PR description. Recommend a corresponding semver bump and CHANGELOG entry documenting the migration path (`new StorageEncryption(pw)` → `await StorageEncryption.create(pw)`).

**Verdict: Approve with minor suggestions.** Core migration is solid, tests are comprehensive, backward compat is preserved. 👍

timjroberts (dismissed):
Looks good, just one or two comments for consideration.

sp-io:
Thanks for the thorough review @timjroberts! Addressed suggestions 2, 4, and 7:

- **2** `console.warn` → reverted to `console.debug` to avoid noise during migration
- **4** `Buffer.equals` → replaced with `timingSafeEqual` in both `decrypt` and `decryptWithPassword` for constant-time comparison
- **7** `signingKeyValue` → restored original `signingKey` parameter name, renamed the destructured binding instead (`{ signingKey: signingKeyLevelName }`)

Regarding **1 (`timingSafeEqual` — pure-JS only)**:

Web Crypto API does not provide a timing-safe comparison primitive. The only available operations are encrypt/decrypt, sign/verify, digest, and key management — no constant-time buffer comparison.

There is an HMAC-based workaround (`sign` + `verify`), but the spec does not guarantee that `verify` performs constant-time comparison internally, so it offers no formal improvement over the pure-JS XOR approach.

The current `constantTimeBufferEqual` implementation is used exclusively for comparing fixed-length SHA-256 hashes (in `verifyPassword`), where the practical risk of a timing side-channel is negligible. Keeping the pure-JS implementation is the right call here.

timjroberts (approved):

PR #801, opened 2026-04-09
test(midnight-js): improve indexer-public-data-provider coverage

sp-io:
## Summary
- Add 23 unit tests for pure functions: `isRegularTransaction`, `toTxStatus`, `toSegmentStatus`, `toSegmentStatusMap`, `toUnshieldedUtxos`, `toUnshieldedBalances`, `IndexerFormattedError`
- Export internal pure functions to enable direct testing
- Exclude auto-generated `src/gen/**` files from coverage metrics

Coverage for `indexer-public-data-provider.ts`: 33.78% → 41.44% (stmts), `errors.ts`: 0% → 100%

## Test plan
- [x] All 44 tests pass (`yarn test` in indexer-public-data-provider)
- [x] Lint passes
- [x] Coverage no longer includes `src/gen/` files

### DIFF
--- a/packages/indexer-public-data-provider/vitest.config.ts
+++ b/packages/indexer-public-data-provider/vitest.config.ts
@@ -28,7 +28,7 @@ export default defineConfig({
       enabled: true,
       clean: true,
       include: ['src/**/*.ts'],
-      exclude: ['**/test/**'],
+      exclude: ['**/test/**', 'src/gen/**'],
       reporter: ['clover', 'json', 'json-summary', 'lcov', 'text'],
       reportsDirectory: './coverage'
     },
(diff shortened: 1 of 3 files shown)


packages/indexer-public-data-provider/vitest.config.ts
```diff
@@ -28,7 +28,7 @@ export default defineConfig({
       enabled: true,
       clean: true,
       include: ['src/**/*.ts'],
-      exclude: ['**/test/**'],
+      exclude: ['**/test/**', 'src/gen/**'],
```
timjroberts, line 31: This will remove the TypeScript code generated from the schema from the bundle. Why do that?

sp-io: These are generated files, do we want to measure coverage of it ? 
My approach is to only cover what we develop.

timjroberts: Ah, yeah. Sorry, I was confusing myself over the config file name - thinking this was defining what gets bundled. Ignore me!

timjroberts (approved):

PR #848, opened 2026-04-16
test: cover security-critical gaps from post-merge coverage analysis

jtsang586:
## Summary

- Adds **56 net new tests** across 5 packages, targeting gaps identified in a post-merge coverage analysis of features merged between Feb–Apr 2026
- Focuses on security-critical regressions (per-recipient encryption keys, shell injection), architectural integrity (protocol ACL subpath wiring + ESLint rule enforcement), and error-path coverage (dapp-connector wallet failures, encryption cache invalidation)
- Adds first-ever test suites to two previously untested packages (`protocol`, `compact`)
- Each test was critically evaluated for value — 12 low-value tests (tautologies, stdlib assertions, fake-interleaving) were written and then pruned in the final commit

### Packages touched

| Package | Tests added | What they guard |
|---|---|---|
| `contracts` | 5 | #745 resolver: burn-point guard, DApp-mapping precedence, spy proving per-recipient key selection |
| `dapp-connector-proof-provider` | 5 | Wallet API error propagation, ZKConfigProvider failure, transient-failure retry semantics |
| `protocol` | 45 | Subpath/namespace parity, representative symbol resolution, ESLint import-restriction rule (via real `ESLint.lintText`) |
| `level-private-state-provider` | 1 | `invalidateEncryptionCache` actually forces `StorageEncryption.create` re-derivation |
| `compact` | 4 | Static-source regression: `exec`/`execSync` with template literals cannot be re-introduced (#711) |

### Key findings during investigation

- **LevelDB exclusive file lock** prevents concurrent `set`/`get` from a single provider instance — the "web crypto concurrency" gap was overstated; the real interleaving surface is limited to in-memory cache operations
- **Protocol ACL package** had zero test coverage despite being a new architectural layer — now covered with structural + ESLint integration tests

## Test plan

- [x] `vitest run` passes in all 5 touched packages (contracts: 40, dapp-connector: 12, protocol: 45, level-private-state: 153, compact: 4)
- [x] `tsc --noEmit` clean in all packages
- [x] `eslint` clean in all packages
- [x] No pre-existing tests were modified or removed (verified via baseline count comparison against `main`)

### DIFF
--- a/packages/compact/package.json
+++ b/packages/compact/package.json
@@ -21,12 +21,14 @@
     "build": "tsc -p tsconfig.build.json && cp src/run-compactc.cjs dist/",
     "lint": "eslint ./src",
     "typecheck": "tsc -p tsconfig.json",
+    "test": "vitest run",
     "deploy": "yarn npm publish --tolerate-republish"
   },
   "devDependencies": {
     "@types/node": "^24.0.0",
     "ts-node": "^10.9.2",
-    "typescript": "^6.0.2"
+    "typescript": "^6.0.2",
+    "vitest": "^4.1.2"
   },
   "files": [
     "dist/"
--- a/packages/compact/src/test/shell-injection-regression.test.ts
+++ b/packages/compact/src/test/shell-injection-regression.test.ts
@@ -1,0 +1,62 @@
+/*
+ * This file is part of midnight-js.
+ * Copyright (C) 2025-2026 Midnight Foundation
+ * SPDX-License-Identifier: Apache-2.0
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
+import { describe, expect, it } from 'vitest';
+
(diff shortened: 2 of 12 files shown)


sp-io:
### Code review

Found 2 issues:

1. **Tautological assertion on hardcoded constant.** `BURN_ENCRYPTION_PUBLIC_KEY` is a hardcoded hex string starting with `f5b9fa49...` — asserting it is not 64 zeros can never fail. This does not verify the constant is a valid curve point, only that it differs from a value it was never set to. [CLAUDE.md](https://github.com/midnightntwrk/midnight-js/blob/0ba0597111c8b0689d32406e0003475095b0f1f8/CLAUDE.md) says "Do not write tests just to get coverage, they must be important, valid scenarios."

https://github.com/midnightntwrk/midnight-js/blob/0ba0597111c8b0689d32406e0003475095b0f1f8/packages/contracts/src/test/utils/zswap-utils.test.ts#L295-L298

2. **`toBeGreaterThan` instead of exact count in cache invalidation test.** After `invalidateEncryptionCache()`, the test asserts `createSpy.mock.calls.length` is merely greater than before. If a bug caused `StorageEncryption.create` to be called multiple times on cache miss, this test would still pass. An exact assertion like `.toBe(callsBeforeInvalidate + 1)` would catch that. [CLAUDE.md](https://github.com/midnightntwrk/midnight-js/blob/0ba0597111c8b0689d32406e0003475095b0f1f8/CLAUDE.md) says "Use strict equality: `expect(actual.sort()).toEqual(expected.sort())`" (Common Mistakes #3 — prefer precise assertions over one-directional checks).

https://github.com/midnightntwrk/midnight-js/blob/0ba0597111c8b0689d32406e0003475095b0f1f8/packages/level-private-state-provider/src/test/level-private-state-provider.test.ts#L1676-L1681

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

sp-io (approved):

PR #863, opened 2026-04-24
Augment interface of unproven-call-tx.ts to allow cross-contract calls

christianschmitz:
I've requested some changes to make it easier to do cross-contract calls 

1. `communicationCommitmentRand` is threaded all the way from call options into `ContractCallPrototype`. Default behavior is unchanged when the field is omitted, because the ledger still generates fresh randomness in that case. The difference is that callers can now provide the exact communication commitment randomness when they need the call transaction to line up with a separately-constructed opening of transientCommit of calling circuits.

2. Create a variant of `createUnprovenCallTxFromInitialStates()` called `createUnprovenCallTxFromInitialStatesWithZswapLocalState()`, which also returns the zswapLocalState. This is more convenient than rebuilding the zswapLocalState externally. We need the zswapLocalState to figure out which outputs are created by the tx, so we can merge them as transient coins in cross-contract calls.

### DIFF
--- a/packages/contracts/src/call.ts
+++ b/packages/contracts/src/call.ts
@@ -25,6 +25,7 @@ import {
   type ZswapLocalState
 } from '@midnight-ntwrk/midnight-js-protocol/compact-runtime';
 import {
+  type CommunicationCommitmentRand,
   type EncPublicKey,
   type LedgerParameters,
   type PartitionedTranscript,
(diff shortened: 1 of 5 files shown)


sp-io:
Hi @christianschmitz — first, our sincere apologies for the very late response on this. The team has been heads-down against some tight release deadlines and this PR unfortunately slipped through the cracks for far too long. That's on us, and we're sorry for leaving your contribution waiting.

Since this was opened, cross-contract call support has landed in `main` through a broader effort led by Joseph Denman — primarily **#967 (feat: Cross-Contract Call Support)**, with follow-up hardening in **#1034** and test coverage in **#1035**. That work covers the same use case this PR was targeting (threading communication commitment randomness and exposing the zswap local state so outputs can be merged as transient coins), so the changes here are now superseded.

Because of that overlap, we're going to close this PR. This is in no way a reflection on the quality of your work — the approach here was sound, and the direction matched where we ended up taking it. We genuinely appreciate you taking the time to dig into `unproven-call-tx.ts` and propose these improvements, and we'd welcome future contributions.

Thanks again, and apologies once more for the delay. 🙏

PR #864, opened 2026-04-24
feat(contracts): allow browser imports by removing fs/path debug writes

christianschmitz:
The `fs` and `path` imports made midnight-js-contracts unimportable by browsers, and it looks like they are only needed for debugging in `submit-tx.ts`. I've removed those imports and filesystem-related debugging statements in `logTransaction()`, only leaving some `console.log()` calls.

(Note: this is easily patched in our dapp using `pnpm patch`, but we've amassed a fair amount of patches and they are becoming unmaintainable)


sp-io (approved):

sp-io:
Hi @christianschmitz, thanks for the contribution!

CI was failing here because PRs from forks can't access repository secrets — specifically `MIDNIGHTCI_PACKAGES_WRITE` (the GHCR PAT used by `docker/login-action`), which is required to pull a private container image during the build. The PR-title check also flagged `contracts` as a non-allowed scope (allowed scopes: `release`, `testkit-js`, `midnight-js`, `deps`, `deps-dev`, `config`).

I've re-submitted your change as #869 from an internal branch, with the title scope adjusted to `fix(midnight-js): ...` and you credited via `Co-authored-by`. Once #869 is merged, this PR can be closed.

Original commit preserved with your authorship; only the commit message was reformatted to satisfy commitlint.

christianschmitz:
@sp-io  We're probably going to keep submitting PRs going forward. How can we make this process smoother?

PR #874, opened 2026-04-30
feat: Add bundled transaction support

JosephDenman:
Adds `withMultiContractScopedTransaction` for batching circuit calls to different contracts into a single ledger intent. Same shape as `withContractScopedTransaction`, but the scope is contract-polymorphic — `ctx.for(contract)` widens the type at each call site.

Cross-contract unshielded flows (A.send / B.receive) only validate when both calls share an intent segment and guaranteed/fallible half. The ledger's effects check (`verify.rs:1599-1651`) matches `claimed_unshielded_spends` against `unshielded_inputs` keyed on `(intent_seg, is_guaranteed_half, token_type, recipient, value). merge()` splices independently-partitioned intents, so alignment is on us; `Transaction.addCalls` jointly partitions, so it's on the ledger. We want the latter.

### DIFF
--- a/packages/contracts/src/bundle.ts
+++ b/packages/contracts/src/bundle.ts
@@ -1,0 +1,118 @@
+/*
+ * This file is part of midnight-js.
+ * Copyright (C) 2025-2026 Midnight Foundation
+ * SPDX-License-Identifier: Apache-2.0
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { Contract } from '@midnight-ntwrk/midnight-js-protocol/compact-js/effect/Contract';
+import { type AnyProvableCircuitId, type FinalizedTxData } from '@midnight-ntwrk/midnight-js-types';
+
+import { type ContractProviders } from './contract-providers';
+import type { FoundContract } from './find-deployed-contract';
+import * as Internal from './internal/bundle';
+import {
+  type ScopedTransactionOptions,
+  type TransactionContext
+} from './transaction';
+
+/**
+ * The aggregate return value of {@link withMultiContractScopedTransaction}. Mirrors
+ * the shape of {@link FinalizedCallTxData} but for a transaction containing more
+ * than one contract call. The per-call records carry private data the caller may
+ * need to apply additional state updates beyond the wallet's automatic private-state
+ * persistence.
+ */
+export interface BundledFinalizedTxData {
+  /**
+   * Public on-chain data of the bundled transaction. Status is guaranteed to be
+   * `SucceedEntirely` if {@link withMultiContractScopedTransaction} returned
+   * normally; non-success statuses are reported via {@link CallTxFailedError}.
+   */
+  readonly public: FinalizedTxData;
+
+  /**
+   * Per-call records in submission order. Each record carries the contract address,
+   * private state ID (if any), circuit ID, and the {@link UnsubmittedCallTxData}
+   * the SDK produced for that call.
+   */
+  readonly calls: readonly Internal.InFlightCall[];
+}
+
+/**
+ * The contract-polymorphic transaction context handed to the
+ * {@link withMultiContractScopedTransaction} callback. Structurally a
+ * {@link TransactionContext}<{@link Contract.Any}, {@link AnyProvableCircuitId}>
+ * so that the SDK's per-call merge protocol accepts it as the outer scope; in
+ * addition exposes {@link MultiContractTransactionContext.for} for ergonomic
+ * typed widening at each `callTx` invocation.
+ */
+export interface MultiContractTransactionContext
+  extends TransactionContext<Contract.Any, AnyProvableCircuitId> {
+  /**
+   * Returns this same context typed for `contract`. Required because each
+   * `FoundContract`'s circuit-call interface is statically typed at its own
+   * `(C, PCK)`, and our context is intentionally type-erased so a single
+   * instance may be threaded through calls to heterogeneous contracts.
+   *
+   * Pure widening: no runtime work is performed; the same instance is returned.
+   *
+   * @example
+   * ```ts
+   * await withMultiContractScopedTransaction(providers, async (txCtx) => {
+   *   await sender.callTx.send(txCtx.for(sender), ...);
+   *   await recipient.callTx.receive(txCtx.for(recipient), ...);
+   * });
+   * ```
+   */
+  for<
+    C extends Contract.Any,
+    PCK extends Contract.ProvableCircuitId<C> = Contract.ProvableCircuitId<C>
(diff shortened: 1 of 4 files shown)


sp-io:
### Review — PR #874: Add bundled transaction support

Thanks for the work — the architecture is solid, mirrors the existing `transaction.ts` / `internal/transaction.ts` split correctly, the symbol-keyed protocol surface is preserved, and the motivation (joint partitioning via `addCalls` for cross-contract unshielded flows) is clearly explained. Local verification: lint ✅, `build:core` ✅, unit tests ✅. Coverage: `bundle.ts` 100%, `internal/bundle.ts` 74,11% lines (uncovered: `bundleIntoSingleIntent` core, `[Submit]`, `getInFlightCalls`).

A few items I'd like to see addressed before approval.

#### Blockers

**1. Terminology — "SDK" vs "framework"**
Before this PR, the term "SDK" did not appear in any `.ts` file in the monorepo. This PR introduces 12 occurrences:
- `bundle.ts:45, 54`
- `internal/bundle.ts:60, 74, 182, 260, 281, 303, 305, 380`
- `test/bundle.test.ts:98, 624`

Several are user-visible — line 260 is an error message thrown at runtime, and line 45 leaks via `BundledFinalizedTxData.calls` JSDoc into the public API surface. Please s/SDK/framework/.

**2. Structural cast on `ledgerParameters` is unnecessary — `internal/bundle.ts:268-270`**
```ts
const ledgerParametersBytes: Uint8Array = (
  cached.states.ledgerParameters as { serialize(): Uint8Array }
).serialize();
```
`LedgerParameters` (re-exported from `@midnight-ntwrk/midnight-js-protocol/ledger`, originally from `@midnight-ntwrk/ledger-v8`) declares `serialize(): Uint8Array` natively. `cached.states.ledgerParameters` is already typed as `LedgerParameters` (see `get-states.ts:36`). The cast is bypassing TypeScript without need; please remove so this becomes `cached.states.ledgerParameters.serialize()`.

**3. Silent partial-persistence after on-chain success — `internal/bundle.ts:365-367`**
```ts
for (const [psId, call] of latestByPsId) {
  await psp.set(psId, call.callData.private.nextPrivateState);
}
```
This loop runs after the chain has already advanced. If the second `set()` throws (LevelDB lock, ENOSPC, AES-GCM error), contract A's local private state is persisted, B's is not, and the user gets:
- no log entry,
- no error wrapping,
- no list of which `privateStateId`s remained unpersisted.

Subsequent calls against B will then read stale local state and produce transcripts that disagree with the chain. Please either (a) `Promise.allSettled` and aggregate failures with the unpersisted IDs, or (b) try/catch and rethrow with `{ cause }` plus a "the following private state IDs were not persisted: …" message.

**4. Missing integration test for `bundleIntoSingleIntent`**
`bundleIntoSingleIntent` (`internal/bundle.ts:122-175`) is the function that justifies the entire feature — joint partitioning, cross-package wasm bridging via `LedgerContractState.deserialize`, `PrePartitionContractCall` construction, `Transaction.fromParts(...).addCalls(...)`. The unit suite necessarily stubs it out via `BundleScopedInternalOptions.bundler` because round-trippable wasm bytes can't be produced in unit tests, which means **0% of the joint-partitioning logic is exercised by CI**.

`testkit-js/testkit-js-e2e/test/contracts.scopedtx.it.test.ts` is the analogue for single-contract scoped transactions; please add a parallel test for the multi-contract case that deploys two contracts with a real cross-contract unshielded flow (sender's `claimed_unshielded_spends` against recipient's `unshielded_inputs`) and asserts `SucceedEntirely`. Without this, the central correctness claim of the PR is unverified.

#### Important

**5. `InFlightCall` is `@internal` but exposed via public type — `bundle.ts:47`**
`BundledFinalizedTxData.calls: readonly Internal.InFlightCall[]` leaks `contractStateBytes` and `ledgerParametersBytes` (serialized wasm snapshots) to public callers. Suggest a public `BundledCallRecord` shape exposing only `{ identity, circuitId, callData }`, keeping `InFlightCall` internal.

**6. `identityForCall` ambiguity — `internal/bundle.ts:388-403`**
The doc admits "the most recently cached identity for that `privateStateId` wins". Two contracts with `privateStateId: undefined`, or two distinct contracts sharing one PSID, silently alias to whichever was cached last. Please throw on duplicate matches and lock it with a unit test.

**7. `for(_contract)` impl signature drift — `internal/bundle.ts:322-326`**
The public interface (`bundle.ts:76-79`) requires `FoundContract<C>`, but the impl is typed `_contract: unknown`. Pure widening is intentional, but the laxer impl signature obscures that. Either narrow the impl parameter to match the interface or document explicitly that the argument is for type widening only and is not validated. The test at `bundle.test.ts:144-146` passes `{} as never` — please tighten it with a throw-on-access Proxy to lock the no-runtime-work invariant.

**8. `[Submit]()` exposed as a no-op delegate — `internal/bundle.ts:308-311`**
Comment says "the framework never invokes this. We expose it so a `BundleContextImpl` is structurally a `TransactionContext`." Delegating to `submit()` (which has different return shape and throws `CallTxFailedError` with a circuit-ID **array**) is a footgun if a future code path ever invokes it. Consider throwing `Error('BundleContextImpl[Submit] should not be invoked directly; use bundleScoped')`.

#### Test coverage gaps

- Identity ambiguity: duplicate `privateStateId`, two `undefined` PSIDs.
- Same-identity threading completeness — `bundle.test.ts:380-382` asserts only `privateState.generation`; please also assert `contractState.data` was replaced and `zswapChainState`/`ledgerParameters` were preserved.
- Submission-failure path other than `CallTxFailedError` (e.g., `proveTx` rejecting) — verifies the error wrap with scope name and `cause`.
- Logger side-effects on the two error-wrap paths (`internal/bundle.ts:450, 466`).
- Deprecated `getCurrentStates()` last-touched semantics (`internal/bundle.ts:224-228`).
- `getAdditionalMappings` and `getLastUnsubmittedCallTxDataToTransact` — currently uncovered.

#### Test quality nits

- `bundle.test.ts:142` — bare `rejects.toThrow()` without a pattern; should be `/No calls were submitted/`.
- The test file uses 30+ structural casts (16× `as unknown as Record<symbol, ...>`, 13× `(s: unknown, i: unknown) => void`, 14× merge-call casts, 4× `(i: unknown) => unknown`). Other tests in `packages/contracts/src/test/` use 0–4 such casts. Please factor a single typed helper that exposes the symbol protocol on the context. The repo guideline forbids avoidable `unknown` casts.
- `bundle.test.ts:455-460` uses one-directional `find()`; combined with `toHaveBeenCalledTimes(2)` it's safe today, but per AGENTS.md prefer strict equality on the full sorted call list.
- `getInFlightCalls()` (`internal/bundle.ts:406`) is added as a test seam but never used in the test suite — either use it or delete (YAGNI).

#### Strengths

- File split mirrors `transaction.ts` / `internal/transaction.ts` correctly; protocol surface preserved and asserted in `bundle.test.ts:117-125`.
- Pre-call snapshot capture invariant test with the serialize-counter trick (`bundle.test.ts:277-329`) is excellent — exactly the regression the PR description flags as load-bearing.
- `CallTxFailedError` correctly bypassed at `internal/bundle.ts:457-459` so domain errors aren't double-wrapped.
- Submission-order preservation, latest-write-wins per PSID, missing-`privateStateProvider` guard, and `FailEntirely` + no-`psp.set()` invariant are all properly covered.
- License headers present on all three new `.ts` files; naming consistency (`withMultiContractScopedTransaction` ↔ `withContractScopedTransaction`) is good.
- `submitTx` accepting `circuitId: PCK | PCK[]` is correct (`submit-tx.ts:48`); the array call at `internal/bundle.ts:341-344` is supported.

Issue #811, opened 2026-04-09
Unsafe optional chaining in error handlers


adamreynolds-io:
## Describe the proposed feature:
Fix unsafe optional chaining in error handlers in `unproven-call-tx.ts` and `unproven-deploy-tx.ts`.

## User Need:
Error handling code in two files can itself throw a `TypeError` under certain conditions, masking the original error and making debugging harder.

## Expected Benefit:
Error handlers reliably propagate the original error rather than potentially throwing a secondary `TypeError` that hides the root cause.

## Details:
In `packages/contracts/src/unproven-call-tx.ts:144-146` (and identically in `unproven-deploy-tx.ts:144-146`):

```typescript
if ((error as any)?.['_tag'] !== 'ContractRuntimeError') throw error;
if ((error as any)?.cause.name !== 'CompactError') throw error;
throw new Error((error as any)?.cause.message, { cause: error });
```

Line 145 uses `?.cause.name` — the optional chain stops at `_tag` but then does a non-optional `.cause.name`. If the error has `_tag === 'ContractRuntimeError'` but no `.cause` property, this throws a `TypeError` ("Cannot read properties of undefined") instead of the original error.

**Fix:** Use `?.cause?.name` and `?.cause?.message` consistently in both files.

---
_Migrated from [PM-22204](https://shielded.atlassian.net/browse/PM-22204)_

[PM-22204]: https://shielded.atlassian.net/browse/PM-22204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Kanasjnr:
@adamreynolds-io can i pick this up?

sp-io:
Already fixed by PR #712

Issue #806, opened 2026-04-09
assertDefined / assertUndefined incorrectly reject falsy values


adamreynolds-io:
## Describe the proposed feature:
Fix `assertDefined` and `assertUndefined` in `packages/utils/src/assertion-utils.ts` to correctly handle falsy values.

## User Need:
Any code using `assertDefined` with valid falsy values like `0`, `""`, `false`, or `0n` will incorrectly throw, causing unexpected failures.

## Expected Benefit:
`assertDefined` and `assertUndefined` behave correctly for all values including valid falsy ones, preventing latent bugs from surfacing as the codebase evolves.

## Details:
`assertDefined` at line 25 uses `!value` (truthiness) instead of a null check. This means it incorrectly throws for valid falsy values like `0`, `""`, `false`, and `0n`. The same inverse bug exists in `assertUndefined` at line 38.

These functions are used throughout the codebase (`get-states.ts`, `find-deployed-contract.ts`, `submit-call-tx.ts`, `unproven-call-tx.ts`, `ledger-utils.ts`). Currently latent because all call sites pass objects, but any future use with numeric or string values will trigger the bug.

**Fix:** Replace `if (!value)` with `if (value === null || value === undefined)` in `assertDefined`, and apply the corresponding fix to `assertUndefined`.

---
_Migrated from [PM-22195](https://shielded.atlassian.net/browse/PM-22195)_

[PM-22195]: https://shielded.atlassian.net/browse/PM-22195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Kanasjnr:
@adamreynolds-io can i pick this up?

PR #887, opened 2026-05-15
fix: correct assertDefined/assertUndefined falsy handling + unsafe optional chaining (#806, #811)

BossChaos:
### #806: assertDefined/assertUndefined incorrectly reject falsy values

**Before:** `if (!value)` rejects `0`, `""`, `false`, `0n`
**After:** `if (value === null || value === undefined)` correctly only rejects null/undefined

Same fix applied to `assertUndefined`:
**Before:** `if (value)` misses falsy non-null values
**After:** `if (value !== null && value !== undefined)`

### #811: Unsafe optional chaining in error handlers

**Before:** `error.cause.name` throws TypeError when `cause` is undefined
**After:** `error.cause?.name` safely returns undefined instead of throwing

Applied to both `unproven-call-tx.ts` and `unproven-deploy-tx.ts`.

## Files Changed
- `packages/utils/src/assertion-utils.ts` - fix falsy value handling
- `packages/contracts/src/unproven-call-tx.ts` - add optional chaining
- `packages/contracts/src/unproven-deploy-tx.ts` - add optional chaining

### DIFF
--- a/packages/contracts/src/unproven-call-tx.ts
+++ b/packages/contracts/src/unproven-call-tx.ts
@@ -143,7 +143,7 @@ export async function createUnprovenCallTxFromInitialStates<C extends Contract.A
     };
   } catch (error: unknown) {
     if (!isEffectContractError(error) || error._tag !== 'ContractRuntimeError') throw error;
-    if (error.cause.name !== 'CompactError') throw error;
+    if (error.cause?.name !== 'CompactError') throw error;
     throw new Error(error.cause.message, { cause: error });
   }
 }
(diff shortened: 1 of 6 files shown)


sp-io:
Closing in favour of #900.

Two reasons:
1. No tests — the fix changes behaviour of utility functions used across 15+ call sites without any test coverage.
2. Scope creep — beyond the `assertDefined`/`assertUndefined` fix required by #806, this PR also modifies `unproven-call-tx.ts`, `unproven-deploy-tx.ts`, `indexer-public-data-provider`, and renames `validateSigningKeyExportPassword` in `level-private-state-provider`. Those changes should go through separate PRs.

#900 addresses only #806 with full test coverage.

Issue #825, opened 2026-04-09
validateSigningKeyExportPassword throws SigningKeyExportError


adamreynolds-io:
## Describe the proposed feature:

`validateSigningKeyExportPassword` throws `SigningKeyExportError`, but it's called during an *import* operation. Getting an "export error" while importing is confusing. This mirrors the existing `importPrivateStates` pattern (which throws `PrivateStateExportError` for import password validation), so it's consistently wrong rather than newly wrong.

## User Need:

Developers need consistent, correctly-named error types.

## Expected Benefit:

Less confusing error handling during import operations.

## Details:

Consider a shared `validatePassword` that accepts the error class as a parameter.

Related: midnightntwrk/midnight-js#526

---

_Migrated from [PM-21964](https://shielded.atlassian.net/browse/PM-21964)_

[PM-21964]: https://shielded.atlassian.net/browse/PM-21964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

adamreynolds-io:
Originally reported by @sp-io in Jira

Issue #822, opened 2026-04-09
maybeThrowApolloError discards error context


adamreynolds-io:
## Describe the proposed feature:

In `indexer-public-data-provider.ts:105-110`, `maybeThrowApolloError` replaces the original `ApolloError` (which contains network errors, GraphQL errors, and stack trace) with a plain `Error` using just the message string. This loses the error `cause`, network details, and original stack.

## User Need:

Developers debugging indexer errors need full error context.

## Expected Benefit:

Preserved error context for debugging.

## Details:

Either `throw result.error` directly or use `throw new Error(result.error.message, { cause: result.error })`.

**Affected files:**
- `packages/indexer-public-data-provider/src/indexer-public-data-provider.ts`

---

_Migrated from [PM-22206](https://shielded.atlassian.net/browse/PM-22206)_

[PM-22206]: https://shielded.atlassian.net/browse/PM-22206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

adamreynolds-io:
Originally reported by @adamreynolds-io in Jira

PR #888, opened 2026-05-15
fix(indexer): preserve ApolloError context in withCompleteQueryData (#822)

BossChaos:
## Problem

`withCompleteQueryData` throws `new Error(result.error.message)` which loses:
- Network errors and stack trace
- GraphQL error details
- Original error cause chain

## Fix

Throw the original `ApolloError` directly:
```diff
- if (result.error) throw new Error(result.error.message);
+ if (result.error) throw result.error;
```

This preserves the full error context for debugging.

Fixes midnightntwrk/midnight-js#822

Wallet: 0xdaE5d307339074A24F579dB48e7c639359D94904

PR #890, opened 2026-05-15
fix: preserve ApolloError context + rename validatePassword (#822, #825)

BossChaos:
### #822: maybeThrowApolloError discards error context

`withCompleteQueryData` in `indexer-public-data-provider.ts:106` replaced the original `ApolloError` (containing network errors, GraphQL errors, and stack trace) with a plain `Error` using just the message string.

**Before:** `throw new Error(result.error.message)`
**After:** `throw result.error`

### #825: validateSigningKeyExportPassword throws SigningKeyExportError during import

The function was called during both export and import operations but threw `SigningKeyExportError` in both cases, confusing developers.

**Before:** `validateSigningKeyExportPassword` → throws `SigningKeyExportError`
**After:** `validatePassword` → throws generic `Error`

## Files Changed
- `packages/indexer-public-data-provider/src/indexer-public-data-provider.ts`
- `packages/level-private-state-provider/src/level-private-state-provider.ts`

### DIFF
--- a/packages/indexer-public-data-provider/src/indexer-public-data-provider.ts
+++ b/packages/indexer-public-data-provider/src/indexer-public-data-provider.ts
@@ -103,7 +103,7 @@ const maybeThrowQueryError = <R extends { error?: { message: string } }>(result:
 const withCompleteQueryData = <A>(): Rx.OperatorFunction<ApolloQueryResult<A>, A> =>
   Rx.pipe(
     Rx.filter((result: ApolloQueryResult<A>) => {
-      if (result.error) throw new Error(result.error.message);
+      if (result.error) throw result.error;
       return result.dataState === 'complete';
     }),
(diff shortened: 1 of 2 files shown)


sp-io:
Thanks for the contribution!

A few concerns:

- **No tests.** Per `CONTRIBUTING.md`, all changes require tests.
- **Two unrelated changes in one PR.** `#822` (Apollo error context) and `#825` (password validator rename) should be separate PRs.
- **#825 part is superseded.** `validateSigningKeyExportPassword` has since been replaced by a shared, fully-policied `validatePassword` with a typed `PasswordValidationError` in `packages/utils/password-validation.ts` (merged in #922). The change proposed here — downgrading to a generic `throw new Error(...)` — is a regression on the public error contract.
- **#822 part** duplicates the one-line change from #888.

Closing this PR. The `#822` fix will be re-implemented with tests in a focused PR. Thanks for raising both issues.

Issue #872, opened 2026-04-29
docs(contracts): add privacy warnings to types containing private-data references


sp-io:
## Summary

Several types in `packages/contracts/src/` carry direct or indirect references to data used in zero-knowledge proof generation — most notably the `UnprovenTransaction` field embedded in transaction-result types. Application code that logs, serializes, or transmits these objects without filtering may unintentionally expose private data.

The framework deliberately exposes these references (they are required for retry, replay, debug, monitoring, and observability use cases), but consumers reading the type definitions cannot tell which fields are privacy-sensitive without reading the source.

This issue adds explicit JSDoc `@remarks` warnings on the affected types so that developers and IDE tooling surface the privacy concern at the point of use.

## Files to update

- `packages/contracts/src/tx-model.ts`
  - `FinalizedCallTxData` (~line 133)
  - `FinalizedDeployTxData`
  - any other exported type whose field signatures transitively reference `UnprovenTransaction` or other private-data carriers
- `packages/contracts/src/submit-call-tx.ts`
  - public surface around line 88 — confirm types returned and reference the same warning if applicable
- `packages/contracts/src/submit-deploy-tx.ts`
  - public surface around line 78 — same as above

## Suggested JSDoc shape

```ts
/**
 * Result of a successful contract call transaction.
 *
 * @remarks
 * **Privacy-sensitive type.** Contains direct or indirect references to
 * private data used in zero-knowledge proof generation — notably the
 * `UnprovenTransaction` accessible via the `tx` field, which may include
 * private inputs that the underlying ZK proofs were designed to keep
 * confidential.
 *
 * Application code should not log, serialize, or transmit instances of
 * this type without explicit field filtering. To export only the data
 * required by the application, destructure the public fields
 * (`txId`, `blockHeight`, etc.) explicitly rather than spreading or
 * stringifying the entire object.
 *
 * The framework deliberately exposes these references to support retry,
 * replay, monitoring, and debug workflows that require access to the
 * underlying transaction structure.
 */
export interface FinalizedCallTxData { ... }
```

## Acceptance criteria

- [ ] All transaction-result types containing private-data references carry a `@remarks` block flagging them as privacy-sensitive
- [ ] Each warning identifies the specific field(s) that carry private data
- [ ] Each warning instructs callers to filter fields explicitly when logging / serializing / transmitting
- [ ] Each warning explains why the framework exposes the references (use cases) so consumers understand it is intentional
[...]

Issue #871, opened 2026-04-29
docs(level-private-state-provider): document `invalidateEncryptionCache` as memory hygiene entry point


sp-io:
## Summary

`levelPrivateStateProvider` exposes a public method `invalidateEncryptionCache()` that clears the cached PBKDF2-derived AES key from process memory. The method exists today (`packages/level-private-state-provider/src/level-private-state-provider.ts:1190-1193`) but carries no JSDoc explaining:

- when an application should call it
- what guarantees it provides
- what guarantees it does **not** provide (JS runtime limitations)

Without documentation, consumers do not know this is the intended memory-hygiene entry point for the provider, or how to integrate it with their session/logout/lock flows.

## Files to update

- `packages/level-private-state-provider/src/level-private-state-provider.ts`
  - `invalidateEncryptionCache()` on the object returned by `levelPrivateStateProvider` (~line 1190)

## Suggested JSDoc

```ts
/**
 * Clears the cached encryption key from process memory.
 *
 * @remarks
 * The provider caches the PBKDF2-derived AES key in memory after the first
 * read or write, because re-deriving it on every operation (600,000
 * iterations) would be prohibitively slow. Call this method when the
 * application reaches a logical security boundary — for example:
 *
 * - user logout
 * - session timeout
 * - app lock / screen lock
 * - before persisting application state to disk for backup
 *
 * Subsequent operations will request the password from the configured
 * provider and re-derive the key on first access.
 *
 * **Limitations.** JavaScript does not guarantee immediate erasure of memory
 * due to immutable strings, non-deterministic garbage collection, and V8
 * runtime internals (string interning, JIT artifacts, function call stack
 * copies). This method removes the application-level cache reference, but
 * residual copies of key material may remain in runtime memory until
 * reclaimed by GC. For threat models requiring cryptographic memory
 * hygiene at the OS level, use a hardware-backed key store outside the
 * JavaScript runtime.
 */
invalidateEncryptionCache(): Promise<void>;
```

## Acceptance criteria

- [ ] `invalidateEncryptionCache()` carries the JSDoc above (or equivalent)
- [ ] JSDoc explicitly enumerates use cases (logout, session timeout, app lock)
- [ ] JSDoc explicitly states the JS runtime limitations
- [ ] `yarn build` passes (TypeDoc clean)
- [ ] `yarn lint` passes

## Out of scope

- Behavioral changes to `invalidateEncryptionCache`
[...]

Issue #870, opened 2026-04-29
docs(types): document `null` return semantics on `PrivateStateProvider.get` / `getSigningKey`


sp-io:
## Summary

`PrivateStateProvider.get()` and `getSigningKey()` both return `Promise<T | null>`, but the JSDoc does not say what `null` actually means. Today it can mean either of:

1. **The key is absent** from the underlying store
2. **The key is present but the stored value deserializes to `undefined`**

These are functionally equivalent for callers ("no usable value"), but the type signature alone gives application developers no way to know:

- which conditions collapse to `null`
- which conditions throw instead
- whether `null` is safe to treat uniformly

The result is defensive `try/catch` around `get()` calls, or bugs where developers assume `null` strictly means "absent."

We are deliberately **not** introducing an `Option<T>` monad to disambiguate — that would be a framework-wide API change better handled if/when we migrate to `effect-ts`. The lighter fix is to document the existing behavior precisely so consumers can reason about it.

## Files to update

- `packages/types/src/private-state-provider.ts`
  - `get(privateStateId)` — interface contract (~line 246)
  - `getSigningKey(address)` — interface contract (~line 273)
- `packages/level-private-state-provider/src/level-private-state-provider.ts`
  - `get` / `getSigningKey` returned from `levelPrivateStateProvider` (~lines 759, 791)

The internal helper `subLevelMaybeGet` (~line 473) does not need new JSDoc — it is unexported and its behavior is fully described by the public surface.

## Acceptance criteria

- [ ] `get()` JSDoc enumerates the conditions under which it returns `null` (key absent, stored value deserializes to `undefined`)
- [ ] `get()` JSDoc lists the conditions under which it **throws**:
  - contract address not set (`setContractAddress` was not called first)
  - password validation failure (password returned by the configured provider does not satisfy minimum strength rules — checked lazily on first access)
  - decryption failure (wrong password, salt mismatch, unsupported encryption version, AEAD tag mismatch)
  - rotation lock timeout (concurrent `changePassword` did not complete in time)
  - underlying store I/O errors propagate as-is
- [ ] `getSigningKey()` carries equivalent documentation, but **without** the "contract address not set" throw — it accepts the address as a parameter and does not depend on `setContractAddress`
- [ ] Both interface methods use `@returns` and `@throws` tags
- [ ] Concrete implementation matches the interface contract (or links to it via `{@inheritDoc}`)
- [ ] `yarn build` passes (TypeDoc clean)
[...]

PR #895, opened 2026-05-15
docs: add JSDoc for null semantics, invalidateEncryptionCache, and privacy warnings (#870, #871, #872)

BossChaos:
### #871: document `invalidateEncryptionCache` as memory hygiene entry point
Added 15-line JSDoc to `invalidateEncryptionCache()` explaining when to call it (logout, account switch, idle), what guarantees it provides (clears cached PBKDF2 keys), and what it does NOT provide (JS runtime limitations on memory wiping).

### #870: document `null` return semantics on `PrivateStateProvider.get` / `getSigningKey`
Added null-semantics JSDoc to both methods explaining that `null` means either key-absent or value-deserializes-to-undefined — callers can treat `null` uniformly as "not found."

### #872: add privacy warnings to types containing private-data references
Added ⚠️ **Privacy Notice** to `UnsubmittedTxData` type explaining that `unprovenTx` contains ZK proof inputs (private state, secret keys, witnesses) and that application code must filter/redact before logging, serializing, or transmitting.

## Files Changed
- `packages/level-private-state-provider/src/level-private-state-provider.ts`
- `packages/types/src/private-state-provider.ts`
- `packages/contracts/src/tx-model.ts`

### DIFF
--- a/packages/contracts/src/tx-model.ts
+++ b/packages/contracts/src/tx-model.ts
@@ -24,6 +24,13 @@ import type { CallResult } from './call';
 
 /**
  * Data relevant to any unsubmitted transaction.
+ *
+ * ⚠️ **Privacy Notice:** The `unprovenTx` field contains a `UnprovenTransaction`
+ * which references zero-knowledge proof inputs (private state, secret keys, and
+ * transaction witnesses). Application code that logs, serializes, or transmits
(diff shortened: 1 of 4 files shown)


sp-io:
Thanks for the contribution!

All three issues this PR targets are already closed and were addressed in **#918** (`docs(midnight-js): clarify provider semantics and tx privacy`, merged 2026-05-28).

Closing as superseded.

Issue #819, opened 2026-04-09
Export passwords have weaker validation than storage passwords


adamreynolds-io:
## Describe the proposed feature:

Storage password requires minimum 16 characters at construction, but `exportPrivateStates` and `exportSigningKeys` accept a custom password with weaker or inconsistent minimum length. Exports — designed to be portable and potentially transmitted — can be encrypted with weaker passwords than at-rest storage.

## User Need:

Consistent password strength across storage and export operations.

## Expected Benefit:

Prevent exports from being encrypted with weak passwords.

## Details:

Apply the same 16-character minimum to export passwords. Validate in both `exportPrivateStates` and `exportSigningKeys`.

**Affected files:**
- `packages/level-private-state-provider/src/level-private-state-provider.ts`

---

_Migrated from [PM-22209](https://shielded.atlassian.net/browse/PM-22209)_

[PM-22209]: https://shielded.atlassian.net/browse/PM-22209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

adamreynolds-io:
Originally reported by @adamreynolds-io in Jira

Issue #818, opened 2026-04-09
HTTP allowed without warning for non-localhost indexer/prover hosts


adamreynolds-io:
## Describe the proposed feature:

`IndexerPublicDataProvider` and `HttpClientProofProvider` accept arbitrary URLs but do not warn when plain HTTP is used for non-localhost hosts. Sensitive data (transaction payloads, proof requests, private state queries) can be sent unencrypted.

## User Need:

Developers need to be warned when they misconfigure URLs to use HTTP instead of HTTPS for remote hosts.

## Expected Benefit:

Prevent accidental unencrypted transmission of sensitive data.

## Details:

When the configured URL uses `http://` and the hostname is not `localhost` or `127.0.0.1`, log a warning via the logger provider. Do not block the connection.

**Affected files:**
- `packages/indexer-public-data-provider/src/indexer-public-data-provider.ts`
- `packages/http-client-proof-provider/src/http-client-proof-provider.ts`
- `packages/http-client-proof-provider/src/http-client-proving-provider.ts`

---

_Migrated from [PM-22210](https://shielded.atlassian.net/browse/PM-22210)_

[PM-22210]: https://shielded.atlassian.net/browse/PM-22210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

adamreynolds-io:
Originally reported by @adamreynolds-io in Jira

PR #897, opened 2026-05-15
fix(indexer+private-state): warn on non-localhost HTTP + enforce password validation on all export paths (#818, #819)

BossChaos:
### #818: HTTP allowed without warning for non-localhost hosts
`IndexerPublicDataProvider` now emits `console.warn` when:
- HTTP (non-WSS) query URL is used for a non-localhost host
- Plain WebSocket (ws://) subscription URL is used for a non-localhost host

Both warnings are emitted at construction time before any network activity.

### #819: Export passwords have weaker validation than storage passwords
Both `exportPrivateStates()` and `exportSigningKeys()` now validate the storage-password fallback in addition to custom passwords. Previously, if no custom password was provided, the storage password was used without checking minimum length requirements.

```ts
// Before: storage password used without validation
const exportPassword = options?.password ?? await getPasswordFromProvider(passwordProvider);

// After: storage password is always validated
if (options?.password !== undefined) {
  validatePassword(options.password);
  exportPassword = options.password;
} else {
  exportPassword = await getPasswordFromProvider(passwordProvider);
  validatePassword(exportPassword);
}
```

### DIFF
--- a/packages/contracts/src/tx-model.ts
+++ b/packages/contracts/src/tx-model.ts
@@ -24,6 +24,13 @@ import type { CallResult } from './call';
 
 /**
  * Data relevant to any unsubmitted transaction.
+ *
+ * ⚠️ **Privacy Notice:** The `unprovenTx` field contains a `UnprovenTransaction`
+ * which references zero-knowledge proof inputs (private state, secret keys, and
+ * transaction witnesses). Application code that logs, serializes, or transmits
+ * `UnsubmittedTxData` or any containing type must filter or redact `unprovenTx`
+ * before doing so. See `packages/contracts/src/` for privacy-sensitive field
+ * locations.
  */
 export type UnsubmittedTxData = {
   /**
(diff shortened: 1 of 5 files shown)


sp-io:
Thanks for the contribution!

Per our `CONTRIBUTING.md` and the PR conventions all changes must include tests (TDD: written first, verified red before fix, green after). 
This PR does not include any tests covering:

- the new `console.warn` behavior for HTTP/WS on non-localhost hosts (positive and negative cases, loopback variants, scheme-aware messaging),
- the additional `validatePassword` call on the storage-password fallback path for both `exportPrivateStates()` and `exportSigningKeys()`.

Without tests, we can't verify the fixes behave correctly or guard against regressions.

In the meantime, both issues have been addressed and merged separately:

- #920 — closes #818 (insecure-URL warnings)
- #922 — closes #819 (full password policy on export)

Closing as superseded. Thanks again for raising these — please feel free to open follow-up PRs with tests for any future fixes.

PR #954, opened 2026-06-04
test(midnight-js): guard submit-tx debug logging against disk writes

jtsang586:
## Summary
Adds regression coverage for #271 (security finding, fixed in #869): the debug transaction logging path in `packages/contracts/src/submit-tx.ts` previously wrote serialized transactions — which can contain ZK witness data and private inputs — to `cwd/logs/transactions/`. The fix in #869 removed all filesystem usage from the module so the debug path only emits to the console. This PR locks that in.

- **Source-level invariant** — asserts `submit-tx.ts` imports no `fs` API, catching `from` / `require(` / `import(` of both `fs` and `fs/promises`. This is the most robust guard: it trips on any reintroduced filesystem import regardless of mechanism (mutation-verified against all three variants).
- **Behavioral test** — with `MN_DEBUG=true` (and `__DEBUG__` compiled to `true` in the package's vitest config), the debug-logging branch runs; a serialization failure inside it must not break submission. Self-guarding: if `__DEBUG__` were ever flipped off, `serialize()` wouldn't be called and the assertion fails loudly rather than passing silently.
- **Test hygiene** — `vi.restoreAllMocks()` in `afterEach` so the `console` spies don't leak into subsequent tests (`clearAllMocks` alone does not un-spy).
- **Cleanup** — drop redundant `as AnyProvableCircuitId` casts flagged by SonarQube S4325 (the alias is just `string`).

No production code changes — test-only.

## Test plan
- [x] New tests fail when the vulnerability is reintroduced (mutation-tested: `from 'node:fs'`, `import 'node:fs/promises'`, and `require('node:fs')` all trip the invariant)
- [x] All existing tests pass (contracts suite: 208/208; this file: 9/9)
- [x] Lint clean, typecheck clean (`tsc --noEmit`)


sp-io:
lgtm!

sp-io (approved):

PR #967, opened 2026-06-09
feat: Cross-Contract Call Support

JosephDenman:
This PR adds cross-contract call (CCC) support to `midnight-js`. The transaction builders in `packages/contracts` now assemble a single call transaction that spans a whole call tree and carries one proof per contract involved, and the runtime resolves each callee's on-chain state lazily as execution reaches it, through a new indexer-backed `ContractStateProvider`.

Since a CCC transaction needs ZK artifacts for several contracts at once, I added a `ZKConfigRegistry` that resolves artifacts across multiple compiled-contract sources without any address registration — it derives the binding by matching the deployed verifier key against each source's local key, which is the same check the chain enforces, so it holds up across redeploys and circuit-name collisions. `nodeZkConfigRegistry` is the zero-config version: point it at the artifact root and it discovers every bundle underneath.

I also moved the indexer public-data-provider onto the deployed indexer's as-of `contract(offset: BlockOffset)` query, so contract state is read at a given block rather than via the exact-block variant, with a coherence guard so a mismatched state/zswap pairing fails loudly instead of quietly returning something inconsistent. The dapp-connector and http-client proving providers were updated for the multi-call proving flow.


packages/indexer-public-data-provider/src/provider.ts
```diff
@@ -166,18 +196,26 @@ export class IndexerPublicDataProvider implements PublicDataProvider {
...
-                ? parseHexLedgerParameters(maybeContractStates.transaction.block.ledgerParameters)
-                : LedgerParameters.initialParameters()
-            ] as [ZswapChainState, ContractState, LedgerParameters])
-          : null
-      );
+      .then((queryResult) => queryResult.data?.contract)
+      .then((maybeContract) => {
+        const action = maybeContract?.actions[0];
+        if (!maybeContract || !action) return null;
+        if (maybeContract.state !== action.state) {
```
sp-io, line 203: This guard fires in two situations that aren't caller mistakes:

1. **Race:** `createUnprovenCallTx` pins `blockHash = latestBlock.hash` via `queryBlock()` and then issues this query as a separate round-trip. Any transaction touching this contract that gets indexed between the two makes `actions[0].state` (newest overall — the schema documents `Contract.actions` as newest-first, with no offset filter) differ from the as-of `contract.state`, so call-tx building fails intermittently on any actively used contract, with a misleading "historical offset was likely supplied" error.
2. **Historical reads:** any caller of the public `queryZSwapAndContractState(addr, { type: 'blockHash', blockHash: <older block> })` now always throws once the contract has a newer action, whereas the pre-PR `contractAction(address, offset)` query returned a coherent historical triple from a single action.

Could `zswapState`/`ledgerParameters` be read from an offset-filtered action instead (as the old `contractAction` query did), or the API split so the block-pinned CCC read and the latest-state read are distinct operations? As it stands the coherence invariant lives only in a comment plus a runtime throw that also catches benign races.

packages/contracts/src/unproven-call-tx.ts
```diff
@@ -335,8 +372,14 @@ export async function createUnprovenCallTx<C extends Contract.Any, PCK extends C
     throw new IncompleteCallTxPrivateStateConfig();
   }
 
+  // Pin the transaction to a single block so the root contract and any cross-contract callees are
+  // read from one coherent chain snapshot (and the circuit's `parentBlockHash` is a real block).
+  const latestBlock = await providers.publicDataProvider.queryBlock();
```
sp-io, line 377: This is fetched unconditionally, but `getContractStates` short-circuits on `TransactionContext` cache hits (L259-261) and returns states read/simulated at an earlier block without ever consulting this hash. The context caches only `{ states, identity }` — no block hash — so on a cache hit the callee resolver and `parentBlockHash` are pinned to a newer block than the root's initial state. That violates `CrossContractConfig`'s own documented precondition ("must be the block at which the initial states given to the call were read") and mixes two chain snapshots in one transaction — wrong callee transcripts / on-chain rejection for CCC circuits inside a scoped transaction.

Suggestion: cache the block hash alongside the states and reuse it on cache hits (or invalidate the cache once the chain advances).

Separately, this adds a mandatory indexer round-trip to every call tx — including non-CCC circuits and the cached-context path where the hash is never used. Could it be fetched lazily, only when block pinning is actually needed?

packages/contracts/src/utils/ledger-utils.ts
```diff
@@ -103,51 +105,82 @@ export const extractUserAddressedOutputs = (
...
+          contractAddress: String(call.contractAddress),
+          circuitId: String(call.circuitId),
+          verifierKeyHash: hashVerifierKey(op.verifierKey)
+        })
+      )
+    );
+  }
 
-  const guaranteedOutputs = extractUserAddressedOutputs(partitionedTranscript[0]);
+  const guaranteedOutputs = extractUserAddressedOutputs(rootCall.public.partitionedTranscript[0]);
```
sp-io, line 169: Unshielded offers are assembled from the root call's transcripts only, but every callee's transcript is attached to the intent via its `ContractCallPrototype` above. If a callee circuit records a user-addressed unshielded output, it never reaches `guaranteedUnshieldedOffer`/`fallibleUnshieldedOffer` — the tx would be unbalanced (rejected) or the payout silently dropped. The docstring states the root-only assumption for *shielded* transfers only; nothing documents or enforces it for unshielded outputs.

If the runtime genuinely cannot produce user-addressed unshielded outputs in callee transcripts, could that be asserted here (fail fast) and documented? Otherwise these two extractions should scan all `calls`, not just `rootCall`.

packages/dapp-connector-proof-provider/src/dapp-connector-proving-provider.ts
```diff
@@ -40,6 +41,13 @@ export type DAppConnectorProvingAPI = Pick<WalletConnectedAPI, 'getProvingProvid
  */
 export const dappConnectorProvingProvider = async <K extends string>(
   api: DAppConnectorProvingAPI,
-  zkConfigProvider: ZKConfigProvider<K>,
+  zkConfigProvider: ZKConfigProvider<K> | ZKConfigRegistry,
```
sp-io, line 44: This signature was widened to accept a `ZKConfigRegistry`, but the outer `dappConnectorProofProvider` factory (`dapp-connector-proof-provider.ts`, untouched by this PR) still accepts only `ZKConfigProvider<K>` — and `ZKConfigRegistry` doesn't satisfy `ZKConfigProvider`, so a multi-source registry can't be passed through the documented browser-wallet entry point (the primary path where CCC matters). `httpClientProofProvider` got the widened union; the workaround here (`createProofProvider(await dappConnectorProvingProvider(api, registry), costModel)`) works but shouldn't be necessary — suggest widening the factory signature to match.

packages/types/src/zk-config-registry.ts
```diff
@@ -0,0 +1,134 @@
...
+  private async resolve(keyLocation: string, parsed: ContractKeyLocation): Promise<ZKConfig<string>> {
+    const memoized = this.resolved.get(keyLocation);
+    if (memoized !== undefined) {
+      return memoized;
+    }
+    for (const source of this.sources) {
+      let verifierKey: Uint8Array;
+      try {
+        verifierKey = await source.getVerifierKey(parsed.circuitId);
+      } catch {
```
sp-io, line 121: This blanket catch treats every failure as "circuit not in this source", but `getVerifierKey` can fail for other reasons — notably `ZkArtifactIntegrityError` from the integrity verification (#1015), EACCES, or transient I/O. Those get silently converted into a misleading `ZKArtifactNotFoundError` ("missing or stale") thrown with no `{ cause }`, hiding what may be a security signal.

Suggest collecting the suppressed errors and attaching them to the final `ZKArtifactNotFoundError` (e.g. as `cause`/`AggregateError`). Same pattern exists in `http-client-proving-provider.ts` (`catch { return undefined }`) and in `node-zk-config-provider.ts`'s `isArtifactBundle` (`catch { return false }` — inconsistent with `readManifest` in the same file, which re-throws non-ENOENT errors).

packages/types/src/zk-config-registry.ts
```diff
@@ -0,0 +1,134 @@
...
+        // The source has no circuit by this name; try the next one.
+        continue;
+      }
+      if (hashVerifierKey(verifierKey) !== parsed.verifierKeyHash) {
+        continue;
+      }
+      const config = await source.get(parsed.circuitId);
```
sp-io, line 128: Two avoidable costs here:

1. `source.get(parsed.circuitId)` re-fetches (and re-integrity-verifies) the verifier key that was just fetched and matched two lines above — the base `get()` calls `getVerifierKey` again. Building the config from the in-hand key plus `getProverKey`/`getZKIR` avoids the duplicate disk read + hash.
2. The `resolved` memo (L56) is keyed by the full location string and stores complete `ZKConfig`s — including multi-MB prover keys — unbounded, for the registry's lifetime (proof providers hold it for the app's life). And since nothing memoizes `(source, circuitId) → vkHash`, each new `(address, circuit, vk)` location rescans all sources sequentially. Consider memoizing the vk hash per source/circuit and caching only the location → source binding (or bounding the artifact cache), so long-lived multi-contract apps don't pin every prover key in memory.

sp-io (commented):
Review pass over the CCC changes. The overall architecture (one call tx per call tree, verifier-key-derived artifact binding, lazy callee state resolution) holds up well. I left inline comments; the first two (coherence guard race, cached-context block skew) I'd consider blocking, the rest are diagnostics/API-parity/efficiency.

packages/contracts/src/unproven-call-tx.ts
```diff
@@ -85,54 +103,64 @@ export async function createUnprovenCallTxFromInitialStates<C extends Contract.A
...
+      ? undefined
+      : makeCalleeStateResolver(crossContract.publicDataProvider, crossContract.blockHash);
+  const baseCircuitContext = {
+    address: ContractAddress(contractAddress),
+    contractState: initialContractState,
+    privateState: initialPrivateState,
+    ledgerParameters
+  };
+  const circuitContext =
+    crossContract === undefined || calleeResolver === undefined
```
sp-io, line 121: The second disjunct is dead: `calleeResolver` is defined a few lines above as exactly `crossContract === undefined ? undefined : makeCalleeStateResolver(...)`, so `calleeResolver === undefined` is true iff `crossContract === undefined` and can never fire independently. Two parallel conditions encoding one invariant will desync silently under a future edit — a single `if (crossContract !== undefined)` branch that builds the resolver and the extended context together would leave one branch point.

packages/indexer-public-data-provider/src/provider.ts
```diff
@@ -120,16 +123,37 @@ export class IndexerPublicDataProvider implements PublicDataProvider {
...
     assertIsContractAddress(address);
-    const offset: InputMaybe<ContractActionOffset> = config
-      ? {
-          blockOffset:
-            config.type === 'blockHeight' ? { height: config.blockHeight } : { hash: config.blockHash }
-        }
+    // The deployed indexer resolves `contract(offset:)` "as of" the given block (the latest
+    // contract action at or before it), which is what cross-contract reads require.
+    const offset: InputMaybe<BlockOffset> = config
+      ? config.type === 'blockHeight'
```
sp-io, line 154: The `BlockHeightConfig | BlockHashConfig → BlockOffset` mapping is now hand-rolled inline in five places in this file (`queryBlock` L129, here, `queryZSwapAndContractState` L185, plus the two pre-existing sites at L426/L476), in three different styles. Worth extracting a single `toBlockOffset(config?)` helper so the next as-of query doesn't add a sixth variant.

packages/contracts/src/utils/ledger-utils.ts
```diff
@@ -103,51 +105,82 @@ export const extractUserAddressedOutputs = (
...
+        call.private.input,
+        call.private.output,
+        Option.match(call.communicationCommitment, {
+          onSome: (commitment) => commitment.commCommRand,
+          onNone: () => communicationCommitmentRandomness()
+        }),
+        encodeContractKeyLocation({
+          contractAddress: String(call.contractAddress),
+          circuitId: String(call.circuitId),
+          verifierKeyHash: hashVerifierKey(op.verifierKey)
```
sp-io, line 163: This gives `createUnprovenCallTxFromInitialStates` a new, undocumented precondition: every operation reached here must carry a verifier key (the ledger's `verifierKey` getter throws otherwise — the PR's own test fixtures had to patch `Contract.initialState` output with a `DUMMY_VERIFIER_KEY`). That's a reasonable requirement now that the vk hash is load-bearing for artifact resolution, but it changes behavior for the exported API when fed raw never-deployed initial states — worth a line in the function's docs (and ideally a clearer error than the getter's throw).

sp-io (commented):
Follow-up to my earlier review — four minor items (nits/cleanups), inline below.

sp-io (approved):

sp-io:
Per the release call: this PR is going to be merged as-is and the reported issues are raised as separate issues you will have to resolve @JosephDenman.

I am marking above comments as resolved even though they are not, just to satisfy the merge check.

PR #985, opened 2026-06-19
test(testkit-js): add vitest coverage for check-no-source-leak guard

jtsang586:
## Summary

The source-build pre-commit guard `scripts/check-no-source-leak.js` (added with the issue #185 submodule mechanism in #979) had **zero automated coverage** and only runs locally via `.husky/pre-commit`. A silent regression to always-`exit 0` would let a `file:./.compact-runtime-home` resolution land on `main` and break CI for every runner without that directory — exactly the failure the guard exists to prevent.

This PR adds a behavior suite for the guard and wires it into CI so the safety net is itself monitored.

- **`scripts/check-no-source-leak.test.ts`** — 8 child-process tests against throwaway `git` fixtures:
  - clean staged `package.json` passes
  - staged root `package.json` / workspace `package.json` / `yarn.lock` carrying the marker are blocked (exit 1)
  - no checkable files staged → passes
  - a leak present only in the working tree (unstaged) → passes (the guard reads the **index**, not the worktree)
  - a staged deletion → passes (`--diff-filter=AM` excludes `D`)
  - **fail-closed**: when `git` can't be queried (not a repo), the guard exits 1 rather than silently passing
- **`scripts/vitest.config.ts`** — standalone config (the root `vitest.config.ts` drives package tests via `projects`).
- **`package.json`** — `test:scripts` script.
- **`.github/workflows/ci-testkit-js-build.yml`** — runs `yarn test:scripts` on every build, independent of the `compactc-from-source` opt-in.

Scope is intentionally limited to the source-leak guard. The companion `use-source-compact-runtime.js` inject script is left untested here: its happy path is already exercised by the gated `compactc-from-source` CI opt-in, and as temporary scaffolding (until `compact-runtime` ships on npm) the extra coverage wasn't judged worth the surface area.

## Test plan

- [x] `yarn test:scripts` — 8/8 pass locally (~1.5s, no nix/Docker required)
- [x] Tests are hermetic — each spins up its own `mkdtemp` git repo and cleans up; no reliance on the real repo
- [x] Guard behavior cross-checked against `check-no-source-leak.js` (index-vs-worktree, `AM` filter, `(^|/)package.json$|^yarn.lock$` matcher, fail-closed branch)
- [x] No production/package code touched; new files carry the Apache-2.0 header
- [ ] CI: `Run script tests` step is green on this PR

## Notes

- This workflow is path-filtered to `packages/**` / `testkit-js/**` (plus root `package.json`/`yarn.lock`/workflows). A PR touching only `scripts/` may not trigger it; acceptable for broad regression protection.
- Child-process execution means these tests don't register in V8 line coverage (intentional; noted in the config).

### DIFF
--- a/.github/workflows/ci-testkit-js-build.yml
+++ b/.github/workflows/ci-testkit-js-build.yml
@@ -183,6 +183,12 @@ jobs:
       - name: Run linter
         run: yarn lint -- ${{ inputs.yarn_lint_paths }}
 
+      # Tests for the root build-tooling scripts (e.g. the check-no-source-leak
+      # pre-commit guard). Runs on every build, independent of the submodule
+      # opt-in, since the guard protects `main` for all contributors.
+      - name: Run script tests
+        run: yarn test:scripts
+
       - name: Run tests
         run: |
           yarn test --filter="./testkit-js/*" --cache-dir ./turbo
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "clean-build": "turbo run clean-build",
     "test": "turbo run test",
     "test:unit": "turbo run test --filter='!@midnight-ntwrk/testkit-*'",
+    "test:scripts": "vitest run --config scripts/vitest.config.ts",
     "it": "turbo run it",
     "e2e": "turbo run e2e",
     "e2e-single": "turbo run e2e-single",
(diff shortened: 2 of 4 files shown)


jtsang586:
Closing PR as workaround is now obsolete

PR #970, opened 2026-06-10
chore(deps)!: swap protocol to ledger-v9 + onchain-runtime-v4 (@midnightntwrk)

sp-io:
# Overview

Switches `packages/protocol` to the new `@midnightntwrk` npm scope and bumps the underlying ledger / onchain-runtime to the v9 / v4 release-candidates, then carries the change through the rest of the framework: the structured `{ tag, value }` key/signature model, a v9-compatible compiler, recompiled contract fixtures, and `wallet-sdk@2.0.0`.

**Package re-exports:**
- `@midnight-ntwrk/ledger-v8@8.1.0` -> `@midnightntwrk/ledger-v9@1.0.0-rc.2`
- `@midnight-ntwrk/onchain-runtime-v3@3.0.0` -> `@midnightntwrk/onchain-runtime-v4@4.0.0-rc.2`

**Also bumped to v9-compatible builds:**
- `@midnight-ntwrk/compact-runtime@0.16.0` -> `0.16.103-dev.3a289c2` (resolutions-pinned; dedupes compact-js's transitive copy)
- `@midnight-ntwrk/compact-js@2.5.3` -> `2.5.5-rc.2` -- the ledger-v9 build: sources `MaintenanceUpdate`/`LedgerParameters` from `@midnightntwrk/ledger-v9` and declares `platform-js ^3.0.0`.
- `@midnight-ntwrk/platform-js@2.2.4` -> `3.0.0` (resolutions-pinned; structured `{ tag, value }` `SigningKey`).
- **compactc compiler** `0.31.0` -> LFDT-Minokawa dev build `compactc-dev-73ebfbbff78118e77a83fdc99dca352db0020869` (reports `0.31.108`). All committed contract fixtures recompiled.
- **`@midnightntwrk/wallet-sdk@1.2.0` -> `2.0.0-canary.20260623092110-2f10bcf`** (+ aligned siblings) -- unblocks `testkit-js`.

**Untouched (no v9-compatible build available yet):**
- `zkir-v2@2.1.0` (see Deferred)

> **Status -- green.** Monorepo build **16/16**, ESLint clean, `typecheck:tests` (packages, strict tsc) clean, `midnight-js-contracts` **211/211**, `testkit-js` unit **77/77**. The previous blocker -- `testkit-js` `tsc` failing on `ZswapSecretKeys` v9-vs-v8, waiting on `wallet-sdk@v9` -- is **resolved** by the `wallet-sdk@2.0.0` bump.

### Protocol swap (core)

[...]

Issue #952, opened 2026-06-03
feat(midnight-js): support public contract log emission (MIP-0002)


sp-io:
## Describe the proposed feature:

Add framework-level support in midnight-js for consuming public contract log
events introduced by [MIP-0002 — Public Contract Log Emission for Compact Smart
Contracts](https://github.com/midnightntwrk/midnight-improvement-proposals/blob/main/mips/mip-0002-public-contract-log-emission.md).

MIP-0002 connects the existing-but-disconnected log pipeline (Impact VM `Log`
opcode → ledger `VersionedLogItem` → indexer GraphQL) and defines a fixed set
of standard `LogEventType` variants plus a `Misc` escape hatch. midnight-js
should expose this surface to dApp developers in a typed, ergonomic way.

## User Need:

DApp developers on Midnight currently have no notification mechanism for
contract activity. To track what a contract has done, consumers must either
poll the contract's serialized state and diff it after each action, or replay
every `ContractCall` transaction and interpret deltas. Both approaches are
expensive, introduce polling latency, and force every dApp to re-implement the
same boilerplate.

Once MIP-0002 ships across the stack, midnight-js consumers should be able to
subscribe to or query typed contract events directly, without writing GraphQL
by hand or decoding raw `StateValue` payloads.

## Expected Benefit:

- Typed access to `LogEventType` standard variants (shielded/unshielded
  spend/receive/mint/burn, paused/unpaused, misc) with version-aware decoders.
- Historical querying and real-time subscription of contract events through
  the existing `publicDataProvider` surface — same mental model dApps already
  use for contract state.
- Field-prefix filtering support (per MIP-0002's `FieldPrefixFilter`) so dApps
  can opt into the disclosure/recall trade-off without reaching for the raw
  GraphQL client.
- Eliminates state-diffing as the default notification pattern, reducing
  latency and bandwidth for downstream dApps and wallets.

## Details:

- Likely lands in `publicDataProvider`: a `contractEvents` query plus a
  matching subscription, both typed by `LogEventType`.
- Typed decoders per variant, with graceful fallback to `Misc` / version 0 for
  non-conforming payloads (matches MIP-0002 §VersionedLogItem behavior).
- Blockers (upstream workstreams must ship MIP-0002 first):
  - **Compact compiler** — adds the `log` expression
    ([CoIP](https://github.com/LFDT-Minokawa/compact/pull/442),
    [compact#377](https://github.com/LFDT-Minokawa/compact/issues/377));
[...]

PR #988, opened 2026-06-22
feat(midnight-js): surface MIP-0002 contract events via PublicDataProvider

sp-io:
## Summary

Adds indexer-sourced **contract event** querying and streaming to `PublicDataProvider`. Source of truth is the indexer's server-side-decoded scalar fields (no ledger-v9 binding), so the entire surface is implementable and **fully unit-tested on `main` today** — only end-to-end validation needs a ledger-9-capable indexer.

**Public API (`packages/types`)**
- `ContractEvent` discriminated union (11 variants, keyed on `eventType`), `ContractEventBase`, `ContractEventAddress`.
- Separate `ContractEventQueryFilter` / `ContractEventSubscriptionFilter`; `ContractEventCursor` (`fromId` xor `fromBlock`); `ContractEventsPage`.
- Two new methods: `queryContractEvents(filter, page?)` and `contractEventsObservable(filter, opts?)`.

**Implementation (`packages/indexer-public-data-provider`)**
- `CONTRACT_EVENTS_QUERY` / `_SUB` with `__typename`, `raw`, and inline fragments for all 11 concrete event types (regenerated `gen/*`).
- `events-mapping.ts` — pure `toContractEvent`; fail-fast on unknown `__typename` / missing required field; nullable → `undefined`; `raw` verbatim; compile-time exhaustiveness guard.
- `events-filter.ts` — request build + validation (PascalCase → SCREAMING_SNAKE, per-variant `fieldName` allow-list, empty-`types` and `Misc`+`fieldPrefixes` rejection, cursor → GraphQL args).
- `contractEvents$` observable (replay-from-cursor → live), the two provider methods, `DEFAULT_CONTRACT_EVENTS_PAGE_SIZE`, new typed error variants, and the `getAllContractEvents` async-iterator helper.
- Docs: README "Contract events" section, `llms.txt` surface, full JSDoc.

**Decisions worth reviewing** (resolve the spec against the *real* `schema.graphql`)
1. `sender`/`recipient` → `{ kind: 'user' | 'contract'; value }` — the schema's `AddressOrContract` is an object, not the scalar the spec assumed.
2. `Misc` → opaque `{ name, payload }` — the schema exposes `payload` as a decoded sibling field (not buried in `raw`).
3. Shielded nullable `amount` aliased to `shieldedAmount` in the GraphQL doc to satisfy GraphQL's `SameResponseShape` (shielded `String` vs unshielded `String!`); the public type is unaffected.
4. `recipient`/`ciphertext` kept **out** of the prefix-filter allow-list (spec's conservative default §3.2/§7.5, fail-fast) but **retained in the response type**. One-line reversible in `events-filter.ts` once the indexer team confirms filterability.

## Test plan
- [x] Tests written first (TDD), watched fail before implementing — mapping (22), filter (37), provider (13), helper (5).
- [x] Full package suite green: **184 passed | 2 skipped**.
- [x] `implements PublicDataProvider` conformance verified by build; downstream `contracts` + `midnight-js` barrel build clean.
- [x] Lint clean on all changed files.
- [ ] Gated E2E acceptance suite recorded as `describe.skip` (`testkit-js-e2e/test/contract-events.it.test.ts`) — enable once ledger-9-capable indexer lands (depends on #970).

> Note: a separate, earlier `feat/mip-0002-contract-events` branch exists with a different (codec-based) approach the rev-2 spec supersedes. This PR is the indexer-decoded-fields implementation.

Related: https://github.com/midnightntwrk/midnight-js/issues/952


notsoblack:
## Code Review: feat(midnight-js): surface MIP-0002 contract events via PublicDataProvider

First off — excellent work. The architecture is clean, the error taxonomy is thoughtful, and the TDD discipline (184 passed before the e2e suite) is exactly the rigour the indexer surface needs. After reading through `events-mapping.ts`, `events-filter.ts`, `get-all-contract-events.ts`, and the provider surface, here are three areas of constructive feedback that could improve reliability and maintainability before this lands.

---

### 1. Observable lifecycle: ensure `contractEvents$` teardown propagates to the underlying Apollo subscription

**File:** `packages/indexer-public-data-provider/src/observables.ts` (the `contractEvents$` factory)

**Observation:** `contractEventsObservable` delegates to `contractEvents$(this.client)(variables)`. Under the hood, Apollo subscriptions return a `ZenObservable.Subscription` (or `Observable` in Apollo Client v3). If the RxJS wrapper does not explicitly forward `unsubscribe()` to Apollo's `subscription.unsubscribe()`, the WebSocket subscription will remain open on the server even after the consumer has disposed of the RxJS observable.

**Why this matters:** In long-running dApp sessions, contract-event observables may be created and torn down repeatedly (e.g., on route changes or when the user switches contracts). A leaked subscription holds a WebSocket message handler and prevents garbage collection of the `ApolloClient` cache fragments tied to that subscription.

**Suggested improvement:** Verify that the RxJS `Observable` returned by `contractEvents$` wraps Apollo's subscription with an explicit teardown:

```ts
// Pseudocode for the observable factory
new Rx.Observable<ContractEvent>((subscriber) => {
  const apolloSub = client.subscribe({ query: CONTRACT_EVENTS_SUB, variables }).subscribe({
    next: (result) => { /* map and emit */ },
    error: (err) => subscriber.error(err),
    complete: () => subscriber.complete(),
  });

  return () => {
    apolloSub.unsubscribe(); // ensure this is present
  };
});
```

If this is already handled (I may have missed it in the diff), a brief inline comment at the return site would help future readers trust the teardown path.

---

### 2. `getAllContractEvents`: document the "all events past snapshot tip" edge case

**File:** `packages/indexer-public-data-provider/src/get-all-contract-events.ts`

**Observation:** The snapshot-tip pinning strategy (`snapshotTipId ??= page[0].maxId`) is elegant and prevents duplicates when events are appended mid-iteration. However, there is a subtle edge case that is not mentioned in the JSDoc:

> If events are appended *so rapidly* that the first page already contains events with `id > snapshotTipId`, those events are silently skipped. This can happen when:
> 1. The indexer is under heavy write load.
> 2. The contract emits many events in a single block.
> 3. The page size (`DEFAULT_CONTRACT_EVENTS_PAGE_SIZE = 100`) is smaller than the burst size.

In that scenario, the caller receives only events up to `snapshotTipId`, missing the "future" events that happened to land on the first page. This is technically correct per the snapshot semantics, but it may surprise consumers who expect a "read everything up to now" iterator to include events that were already visible on the first page.

**Suggested improvement:** Add a `@caveat` or `@remarks` block to the JSDoc:

```ts
/**
 * ...existing docs...
 *
 * @remarks
 * The snapshot tip is pinned from the *first* page's `maxId`. If that page
 * already contains events with `id > maxId` (possible under extreme write
 * concurrency or a page-size mismatch), those events are intentionally
 * excluded to preserve snapshot consistency. Increase `limit` or paginate
 * manually if you need strict "read everything visible at invocation time".
 */
```

This turns a potential support question into documented, expected behavior.

---

### 3. Type duplication in `events-mapping.ts` for `AddressOrContractNode`

**File:** `packages/indexer-public-data-provider/src/events-mapping.ts`

**Observation:** You manually define:

```ts
type AddressOrContractNode = {
  readonly kind: AddressOrContractKind;
  readonly userAddress: string | null;
  readonly contractAddress: string | null;
};
```

but the GraphQL codegen (`gen/graphql.ts`) almost certainly emits a matching type for the `AddressOrContract` schema object. Manually redefining it creates a maintenance burden: if the indexer schema later adds a new address kind (e.g., `RELAY`) or renames fields, this local type will drift from the generated source of truth.

**Suggested improvement:** If the generated type is usable directly, replace the manual definition with an import from `./gen/graphql`. If the generated type is awkward (e.g., uses `Maybe<string>` instead of `string | null`), consider a branded type alias rather than a full redeclaration:

```ts
import type { AddressOrContract } from './gen/graphql';

// If the generated type is close but not identical:
type AddressOrContractNode = AddressOrContract;
```

This makes schema changes a compile-time concern rather than a runtime bug.

---

### Minor nit (not blocking)

In `events-mapping.ts`, the `toEventAddress` fallback for an unknown `kind` throws:

```ts
throw IndexerDataError.missingEventField(typename, `${field}.kind`);
```

Semantically, the field is not "missing" — the indexer *did* return a `kind`, but its value is outside the expected set. A new static factory like `IndexerDataError.unknownAddressKind(typename, kind)` would make Sentry grouping and log analysis more precise. This is a nice-to-have, not a blocker.

---

### Summary

| Concern | Priority | Suggested action |
|---------|----------|------------------|
| Apollo subscription teardown | Medium | Verify / document `unsubscribe` forwarding in `contractEvents$` |
| Snapshot-tip edge case | Low | Add `@remarks` to `getAllContractEvents` JSDoc |
| Type duplication | Low | Replace manual `AddressOrContractNode` with generated type |
| Error semantics (unknown kind) | Nice-to-have | Consider `unknownAddressKind` factory |

None of these are regressions — they are polish items that help the next maintainer. The PR is already well above the bar for merge. 👍

sp-io:
Thanks for the thorough read, @notsoblack — appreciate the time. I went back through each point against the code on the branch; here's where things land.

### 1. Apollo subscription teardown — already handled

The premise here is `@apollo/client` v3 / zen-observable, but this package is on **`@apollo/client@4.2.0`**, which is built natively on **RxJS** (rxjs is a direct dependency, and the `dataState` field this code uses is an AC4 API). So `apolloClient.subscribe(...)` returns a real `rxjs.Observable`, `.pipe(...)` is RxJS's own operator chain, and teardown propagates automatically — unsubscribing downstream tears down the source subscription and closes the WS link. No manual `new Rx.Observable(... return () => sub.unsubscribe())` wrapper is needed (it'd actually be a step back from the idiomatic form).

It's also the exact pattern every other subscription in `observables.ts` uses (`contractState$`, balances, blocks) — if it leaked it'd be a package-wide issue, not specific to `contractEvents$`. No change needed.

### 2. `getAllContractEvents` snapshot tip — the skip can't happen on the first page

`maxId` is *"the highest event id the indexer currently knows"* (documented in `public-data-provider.ts`), returned in the **same response** as the page. Since `snapshotTipId = page[0].maxId` and every event on that first page came from that same response, they all satisfy `id <= maxId` — so the `if (event.id > snapshotTipId) return` guard is structurally unable to fire on the first page. It only matters on subsequent pages, which is exactly the intended snapshot semantics.

The "events appended after invocation are excluded" behavior is also already in the JSDoc: *"the highest event id known at the first page (`maxId`) is the snapshot tip, and events appended after it (`id > tip`) are excluded — so a page-boundary insert can neither duplicate nor skip."* So I'll leave the docs as-is.

### 3. `AddressOrContractNode` duplication — fair, will tidy

Agreed it's a hand-maintained shape. Note the generated schema type `AddressOrContract` uses `Maybe<Scalars['HexEncoded']['output']>` rather than the selection-set's `string | null`, so a direct `= AddressOrContract` alias couples to the whole-schema type. The drift-proof option is to derive from the operation type already in use:

```ts
type AddressOrContractNode = Extract<ContractEventNode, { sender: object }>['sender'];
```

Low priority and non-blocking, but happy to fold this in.

### 4. Unknown `kind` error semantics — good catch

You're right that `kind` was present (and `AddressOrContractKind` even includes `'%future added value'`), so `missingEventField(typename, '...kind')` is mislabeled for an out-of-set value. A dedicated `IndexerDataError.unknownAddressKind(typename, kind)` is more accurate for log/Sentry grouping and fits the fail-fast style already there. Will add.

---

Net: 3 and 4 are nice tidy-ups I'll take; 1 and 2 are no-ops (AC4/RxJS handles teardown, and the first-page skip is structurally impossible + already documented). Thanks again. 👍

sp-io:
Points 3 and 4 are addressed in `4d86813f`.

- **3 — type duplication:** dropped the hand-written `AddressOrContractNode` in favour of deriving it from the generated operation node, so a schema change to `AddressOrContract` now breaks the build instead of drifting silently:
  ```ts
  type AddressOrContractNode = Extract<ContractEventNode, { sender: object }>['sender'];
  ```
- **4 — error semantics:** added `IndexerDataError.unknownAddressKind(typename, field, value)` and switched `toEventAddress` to throw it for an out-of-set `kind`. The null-field path still throws `missingEventField`, so the two failure modes stay distinct and Sentry/log grouping is accurate.

Verified: package suite `185 passed | 2 skipped` (the one added negative test), lint clean, `tsc`/dts build green.

Points 1 (Apollo v4/RxJS handles teardown) and 2 (first-page skip is structurally impossible and already documented) are no-ops as discussed above. Thanks again for the review. 👍

packages/indexer-public-data-provider/src/events-filter.ts
```diff
@@ -0,0 +1,173 @@
...
+  return translated;
+};
+
+/**
+ * Enforces: `fieldPrefixes` is only legal when `types` is present and every
+ * filtered type is a standard (non-`Misc`) variant, and each `fieldName` is an
+ * indexed field of *every* filtered type.
+ */
+const validateFieldPrefixes = (
+  types: readonly ContractEventType[] | undefined,
```
VanessaPC, line 86: Should we not remove `undefined`? since we are throwing on it at the start?

packages/indexer-public-data-provider/src/events-filter.ts
```diff
@@ -0,0 +1,173 @@
...
+  fieldPrefixes: readonly { fieldName: string; prefix: string }[]
+): void => {
```
VanessaPC, line 88: So the validation, if correct, comes as void? wouldn't it return the validated valid prefix? or a boolean?

packages/indexer-public-data-provider/src/events-filter.ts
```diff
@@ -0,0 +1,173 @@
...
+  if (filter.fieldPrefixes !== undefined && filter.fieldPrefixes.length > 0) {
+    validateFieldPrefixes(filter.types, filter.fieldPrefixes);
+  }
+
+  return {
+    contractAddress: filter.contractAddress,
+    types,
+    fieldPrefixes: filter.fieldPrefixes ?? null,
+    fromBlock: null,
+    toBlock: filter.toBlock ?? null,
```
VanessaPC, line 136: Can block and txHash and fieldprefix ever be null?

packages/indexer-public-data-provider/src/events-filter.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Builds the variables for `CONTRACT_EVENTS_QUERY`: validated filter plus
+ * inclusive `fromBlock`/`toBlock` bounds and `limit`/`offset` pagination. The
+ * `defaultLimit` is applied when `page.limit` is omitted.
+ */
+export const buildQueryVariables = (
+  filter: ContractEventQueryFilter,
+  page: ContractEventsPage | undefined,
+  defaultLimit: number
+): ContractEventsQueryQueryVariables => ({
+  filter: { ...buildBaseFilter(filter), fromBlock: filter.fromBlock ?? null },
```
VanessaPC, line 151: same here, can it be null? the fromBlock? I would think there is always a block, so null shouldn't be returned?

packages/indexer-public-data-provider/src/events-mapping.ts
```diff
@@ -0,0 +1,165 @@
...
+  id: node.id,
+  maxId: node.maxId,
+  version: node.version,
+  contractAddress: node.contractAddress as ContractAddress,
+  transactionId: node.transactionId,
+  raw: node.raw
+});
+
+/** Asserts a required string field is present, throwing fail-fast otherwise. */
+const required = (value: string | null | undefined, typename: string, field: string): string => {
```
VanessaPC: I know it's a function confined to this file.. but it's fairly vague, something like `requiredFieldIsPresent` will make the code more intuitive. 
Also, why do we allow `null` and `undefined`.. they are not a event field, and I dont think they allow for this. So should it be stricter there and then make sure that the field is an actual string to which fail?

packages/indexer-public-data-provider/src/events-mapping.ts
```diff
@@ -0,0 +1,165 @@
...
+  if (value == null) {
```
VanessaPC: `===`

packages/indexer-public-data-provider/src/events-mapping.ts
```diff
@@ -0,0 +1,165 @@
...
+    throw IndexerDataError.missingEventField(typename, field);
+  }
+  return value;
+};
+
+/** Normalizes an absent (`null`) nullable field to `undefined`. */
+const optional = (value: string | null | undefined): string | undefined => value ?? undefined;
```
VanessaPC: Why do we want to normalise a `null`? But if we do, ` optionalNullableField`. I am just not sure ..about the function need.

packages/indexer-public-data-provider/src/events-mapping.ts
```diff
@@ -0,0 +1,165 @@
...
+ * rather than emitting a half-built address — on an unrecognized `kind`
+ * (`unknownAddressKind`) or an absent `kind`-selected address field
+ * (`missingEventField`).
+ */
+const toEventAddress = (
+  address: AddressOrContractNode,
+  typename: string,
+  field: string
+): ContractEventAddress => {
+  if (address.kind === 'USER') {
```
VanessaPC, line 72: The type of address kinds could be a`type`.

packages/indexer-public-data-provider/src/observables.ts
```diff
@@ -353,3 +357,32 @@ export const blockOffsetToUnshieldedBalances$ =
...
+        query: CONTRACT_EVENTS_SUB,
+        variables,
+        fetchPolicy: 'no-cache'
+      })
+      .pipe(
+        withValidFetchData(),
+        Rx.map((data) => {
+          const node = data.contractEvents;
+          if (!node) {
+            throw new IndexerSubscriptionDataError('contractEvents');
```
VanessaPC, line 384: Do we not have 'contractEvents' on a constant given that it seems a normal error throw or a normal case to come accross ?

packages/indexer-public-data-provider/src/events-filter.ts
sp-io: The `| undefined` is the input contract here — `buildBaseFilter` passes `filter.types`, which is legitimately optional. The early throw is the validation that rejects the `undefined` case. Dropping `undefined` from the signature would just push the same check up to the caller.

packages/indexer-public-data-provider/src/events-filter.ts
sp-io: Intentional — it's a fail-fast assertion: it throws `IndexerProviderConfigError` on anything invalid and returns nothing otherwise (matching the repo's prefer-throw-early convention). There's no validated value to hand back; `fieldPrefixes` is forwarded as-is once it passes, so a boolean/return value wouldn't add anything.

packages/indexer-public-data-provider/src/events-filter.ts
sp-io: Yes — in the generated `ContractEventFilter` they're all `InputMaybe<…>` (nullable inputs). GraphQL distinguishes an explicit `null` from an absent field, and our codegen models "omit this filter" as `null`, so `?? null` is the required idiom to satisfy the generated input type.

packages/indexer-public-data-provider/src/events-filter.ts
sp-io: This `fromBlock` is the query's *lower-bound filter input*, not the event's block height. `null` means "no lower bound" (stream/page from the start). The event's own block is always populated on the response side — different field. So `null` is correct and meaningful here.

packages/indexer-public-data-provider/src/events-mapping.ts
sp-io: Good call on the name — renamed to `requireField` (and `optional` → `optionalField`) in 9146a921. I kept it returning the value rather than a boolean, so `requiredFieldIsPresent` would read as misleading.

On null/undefined: I also tightened the param to `string | null`. The generated result types only ever produce `string` or `string | null` (GraphQL guarantees selected fields are present, just possibly null) — `undefined` was unreachable. The remaining `null` guard is deliberate fail-fast hardening against indexer schema drift, since GraphQL nullability doesn't enforce our decode invariants.

packages/indexer-public-data-provider/src/events-mapping.ts
sp-io: Done in 9146a921 — now that the param is `string | null` (no `undefined` possible) it's an exact `=== null`. For the record the previous `== null` wasn't a bug; it intentionally covered the declared `| undefined`. Narrowing the type is what made `===` the correct exhaustive form.

packages/indexer-public-data-provider/src/events-mapping.ts
sp-io: The normalisation maps GraphQL `null` → TS `undefined` so the public `ContractEvent` type uses idiomatic optional fields (`field?: string`) instead of exposing `| null` to consumers. Renamed to `optionalField` for symmetry with `requireField` in 9146a921.

packages/indexer-public-data-provider/src/events-mapping.ts
sp-io: The `kind` discriminant is already typed — it comes from the generated `AddressOrContractKind` union, and `AddressOrContractNode` is `Extract`-derived from the generated operation type, so a schema change surfaces as a compile error here rather than silent runtime drift. The string literals are the narrowing checks against that union. Happy to extract a named alias if you'd prefer it spelled out.

packages/indexer-public-data-provider/src/observables.ts
sp-io: It's already constrained: `IndexerSubscriptionDataError` takes `IndexerSubscriptionField = 'blocks' | 'contractActions' | 'contractEvents'`, so the literal is type-checked rather than a free-floating string — a typo would fail compilation. Can promote it to a named const if you'd prefer, but it's not a magic string today.

VanessaPC (approved):

Issue #901, opened 2026-05-18
Adopt possibility of ECDSA keys/signatures in Contract Maintenance Authorities


kapke:
There is a need to support ECDSA signatures in Midnight related to availability of MPC and HSM implementations. Ledger 9 introduces possibility of ECDSA keys/signatures for contract maintenance. 

Midnight.js needs to be aware of these because of the interface to contract maintenance actions.

At this moment, supporting documentation is available at internal repositories (https://github.com/shieldedtech/ChiefArchitect/pull/118 and https://github.com/shieldedtech/product/pull/101). Though a MIP will be published based on them.

Acceptance criteria:
- when generating contract maintenance authority key - take an option allowing to specify whether it will be a key for ECDSA or Schnorr
- otherwise - when specifying contract maintenance authority or issuing contract maintenance actions - allow to use ECDSA keys/signatures

DarrenOliveiro-Priestnall:
MIP available at https://github.com/midnightntwrk/midnight-improvement-proposals/blob/main/mips/mip-0003-ecdsa-support.md

sp-io:
Related: https://github.com/midnightntwrk/midnight-js/issues/1000

sp-io:
Verified working: https://github.com/midnightntwrk/midnight-js/pull/999

PR #1014, opened 2026-06-30
test(testkit-js): cover ECDSA contract maintenance actions and key persistence

jtsang586:
## Summary

Extends test coverage for ECDSA contract maintenance authority (CMA) keys, addressing the testing side of #901 ("Adopt possibility of ECDSA keys/signatures in Contract Maintenance Authorities"). The functional fix and the config-level regression guard already landed in #999; the remaining gap was that ECDSA was only exercised end-to-end for the *replace-authority* action, not the verifier-key maintenance actions, and never through the signing-key persistence round-trip.

- **e2e** (`testkit-js-e2e/test/contracts.signing-key.it.test.ts`): proves an ECDSA maintenance authority can authorize the verifier-key actions — `submitRemoveVerifierKeyTx` then `submitInsertVerifierKeyTx` — each asserting `SucceedEntirely`, which proves the ECDSA signature was produced and verified by the node. Folded into the existing signing-key e2e to reuse the already-running devnet (no extra container spin-up).
- **unit** (`level-private-state-provider`): a mixed schnorr+ecdsa signing-key export → clear → import round-trip, asserting each key and its `tag` survive the superjson + AES-256-GCM persistence path.

This is a deliberately minimal set: ECDSA config-key wiring is already guarded by the existing `packages/types/src/test/contract.test.ts` regression test, so redundant unit-level kind assertions were intentionally left out.

## Test plan

- [x] Unit suite green: `level-private-state-provider` (188 passed, incl. the new mixed-kind round-trip)
- [x] `packages/types` rebuilt (refreshed stale `dist` so `contracts` typechecks against current `PublicDataProvider`)
- [ ] e2e (`[@slow]`, Docker devnet required) — not run in this environment; needs a local/CI devnet before merge
- [x] No regressions in adjacent governance unit tests (6 passed)
- [x] Lint clean (pre-commit eslint passed)

Relates to #901.

### DIFF
--- a/packages/level-private-state-provider/src/test/level-private-state-provider.test.ts
+++ b/packages/level-private-state-provider/src/test/level-private-state-provider.test.ts
@@ -1006,6 +1006,30 @@ describe('Level Private State Provider', (): void => {
       expect(await db.getSigningKey(CONTRACT_ADDRESS_2)).toEqual(signingKey2);
     });
 
+    test('exports and imports mixed schnorr and ecdsa signing keys preserving each kind', async () => {
+      const db = levelPrivateStateProvider<PID, PS>(testConfig);
+      const schnorrKey = sampleSigningKey('schnorr');
+      const ecdsaKey = sampleSigningKey('ecdsa');
+      await db.setSigningKey(CONTRACT_ADDRESS_1, schnorrKey);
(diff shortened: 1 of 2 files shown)


sp-io (approved):
LGTM

PR #1008, opened 2026-06-29
build(deps): bump compact-runtime and compactc to 1dcee19f

sp-io:
## Summary
- Bump `@midnight-ntwrk/compact-runtime` to `0.17.104-dev.1dcee19f4367a5cd36b23e81fd917d58844beeb0` (root resolutions + `packages/protocol`)
- Bump `COMPACTC_VERSION` to `1dcee19f4367a5cd36b23e81fd917d58844beeb0` in `.envrc`
- Recompile all managed contracts via `yarn compact` — artifacts now built with compiler `0.32.108` / language `0.24.102` / runtime `0.17.104`
- Adapt the `ledger-utils` `receiveShielded` test (issue #686) to the new runtime API surface:
  - `Contract.initialState` and circuit functions are now async (return Promises) → `await`ed
  - `createCircuitContext` now takes a leading `circuitId` argument
  - `CircuitResults` restructured: proof data moved into `context.callProofDataTrace` (root circuit is the final entry), `gasCost` is a top-level field, and `currentQueryContext` now lives under `context.callContext`

Only this one test references the changed runtime internals; no production framework code touches those APIs directly.

## Test plan
- [x] `yarn compact` recompiles all contracts successfully
- [x] `packages/contracts` `ledger-utils.test.ts` passes (28/28)
- [x] Lint clean on changed files; pre-push typecheck passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1015, opened 2026-06-30
feat(midnight-js)!: verify ZK artifacts against the compactc integrity manifest

sp-io:
## Summary

Implements [shieldedtech/shielded-security-engineering#273](https://github.com/shieldedtech/shielded-security-engineering/issues/273) — ZK artifacts (prover keys, verifier keys, ZKIR) were loaded with **no integrity verification**. This adds SHA-256 verification against the `compiler/contract-manifest.json` now emitted by `compactc`.

- **New shared module** `@midnight-ntwrk/midnight-js-utils/zk-artifact-manifest`: manifest type, parser (asserts `manifest-version === '1'`, one-level flatten to `"<dir>/<file>"` keys, validates hex hash + size), `computeSha256Hex`, `verifyZkArtifactIntegrity`, `assertManifestHash`. Cross-platform hashing via `@noble/hashes` (works in browser — the fetch provider runs there). `types/` untouched.
- **`FetchZkConfigProvider` / `NodeZkConfigProvider`** lazily load + cache the manifest from their base location and verify each artifact's bytes after load.
- **Default mode `require` (fail-closed):** a missing manifest or a digest mismatch throws. Opt out per provider with `{ verify: 'warn' }` (warn on missing, still throw on mismatch) or `{ verify: 'off' }`.
- **Build-time pin:** optional `expectedManifestHash` anchors the whole chain in one hex string the dApp controls — the only mode that defends against a coordinated source-compromise adversary (the co-located unkeyed manifest alone does not; see design notes).

Stacked on #1008 (compactc `1dcee19f` is what emits the manifest).

## ⚠️ BREAKING CHANGES

- **`FetchZkConfigProvider`** constructor: `(baseURL, fetchFunc?)` → `(baseURL, options?)`. Migrate `new FetchZkConfigProvider(url, fetch)` → `new FetchZkConfigProvider(url, { fetchFunc: fetch })`. The bare `new FetchZkConfigProvider(url)` form is unchanged.
- **`NodeZkConfigProvider`** new `require` default: loading artifacts that have **no** `compiler/contract-manifest.json` (older manifest-less `compactc` output, hand-assembled key dirs) now **throws**. Recompile with a manifest-emitting `compactc`, or opt out with `{ verify: 'off' }` / `{ verify: 'warn' }`. (All in-repo artifacts carry manifests; no in-repo production call sites needed migration.)

## Test plan
- [x] TDD: tests written first, verified red before implementation (per-task RED/GREEN evidence)
- [x] All existing provider tests pass (path-traversal, SPA-fallback, 404), migrated to the options shape
- [x] New negative coverage: tamper (deterministic byte-flip) rejects with path + digests; missing manifest throws under `require`, warns under `warn`; pin mismatch + pin-with-absent reject; partial/other-circuit manifest treated as absent; manifest fetched once under concurrent `get()`; transient-network-error retry not memoized; Node small-file Buffer hashed without `byteOffset` mis-hash
- [x] `yarn build` + `yarn test` + `yarn lint` green for the three changed packages (277 tests)

> Note: repo-wide `yarn typecheck:tests` currently fails in `dapp-connector-proof-provider` due to the `ledger-v9` rc.3 `ProvingProvider.lookupKey` vs `dapp-connector-api` incompatibility introduced on the #1008 base — unrelated to this feature; resolves when #1008's rc.3 bump is reconciled.

## Follow-ups
- **AC7:** file an upstream issue to migrate the canonical manifest schema into `@midnight-ntwrk/compact-js` (kept local to `utils` for now to avoid re-blocking on another team's release).
- Optional: make the Node small-file DEV-2 test's negative companion deterministic (force nonzero `byteOffset`).


CjDabrow (commented):
This is a solid, well-tested addition, and I like that it's fail-closed by default. I read the manifest module and both providers against the design notes; the mechanism is correct and the negative coverage is thorough. My main feedback is about making the trust boundary as visible in the API as it is in your head, plus a couple of smaller things.

**1. The default `require` mode invites a false sense of security, and the docs don't correct it at the API surface.** Without `expectedManifestHash`, the manifest is fetched from the same base location as the artifacts and trusted. So `require` genuinely protects against corruption (partial deploy, CDN truncation, a stale or wrong artifact set), but an adversary who can tamper with the artifacts can also tamper with the co-located manifest to match. You know this and call it out in the design notes, but the TSDoc a consumer actually sees doesn't: `verify` just describes the modes, and `expectedManifestHash` says "pinned by the application at build time" without stating it is the *only* mode that resists a coordinated swap. I'd move that point up into the `verify` and `expectedManifestHash` doc comments, something like: "without `expectedManifestHash`, this detects corruption but not a malicious manifest substituted alongside tampered artifacts." The phrase "integrity verification" reads as tamper-resistance to most people, so the caveat should be impossible to miss.

**2. The `require`-by-default breaking change needs a louder runway in the thrown error.** Flipping `NodeZkConfigProvider` to throw on manifest-less artifacts is the right secure default, but every downstream consumer with older `compactc` output breaks on upgrade. The migration is well documented in the PR body, but the *runtime* error a stuck developer hits only says "integrity verification is required." It doesn't mention the two escape hatches (`{ verify: 'off' }` / `'warn'`) or "recompile with a manifest-emitting compactc." Folding that guidance into the error message would save a lot of support round-trips. Separately, worth a sentence on whether `warn` should be the default for one release before flipping to `require`, purely as a softer rollout. I lean toward keeping `require` (secure by default), but it's worth stating the choice deliberately given it's a breaking behavior change.

**3. `size` is validated in the manifest but never verified against the artifact.** `parseZkArtifactManifest` requires each entry to have a non-negative integer `size`, but `verifyZkArtifactIntegrity` only compares the hash. `size` is never read. A hash match subsumes length, so this isn't a security gap, but the field currently implies a check that doesn't happen. Either use it as a cheap pre-hash length check (defense in depth, and a clearer "truncated artifact" error before you hash megabytes), or drop it from the entry validation so it doesn't read as load-bearing.

**4. The parsed compiler/language/runtime versions are unused, which is a missed supply-chain check.** They're parsed into the manifest object but nothing consults them. Since you're already at the integrity boundary, an optional "assert the manifest's `compiler-version` matches what I expect" would let a dApp reject artifacts built by an unexpected toolchain, which is a real supply-chain angle adjacent to what this PR is for. Fine as a follow-up, just flagging the opportunity while the code is open.

**5. Smaller notes:**
- `isHtmlFallback` keys on `content-type: text/html`. A server that returns the real manifest with a wrong `text/html` content-type would be treated as "manifest absent" and (under `require`) throw even though the file was present. Fail-closed so it's safe, just a possible confusing false negative worth a comment.
- The fetch provider URL-encodes `circuitId` (`encodeURIComponent`) but the manifest lookup key uses the raw `circuitId`. `assertSafeName` presumably keeps the charset narrow enough that these can't diverge; worth confirming that assumption holds so a name that's valid but URL-encoded can't miss its manifest entry.
- Agree with your own follow-up on making the Node small-file `byteOffset` negative test deterministic; that's the one spot where a regression could silently slip through.

Net: the logic looks correct and the fail-closed posture is the right call. Nothing here blocks it in my view. Item 1 (surfacing the trust boundary in the docs) is the one I'd really want in before merge, since the whole value of the feature hinges on users understanding what it does and does not guarantee.

sp-io:
Thanks for the thorough read — I agree with most of it. A few responses:

**On point 2 (the `require`-by-default "breaking runway"):** I don't think the migration hazard is as broad as framed. This release is tightly coupled to the Ledger-v9 stack — it pins `ledger-v9 1.0.0-rc.2`, `onchain-runtime-v4 4.0.0-rc.2`, `compact-runtime 0.18.0-rc.0`, and `compactc 0.33.0-rc.0`. Those runtime/ledger bumps are ABI-breaking: prover/verifier keys and ZKIR produced by an older `compactc`/runtime won't load against this runtime regardless of integrity checks. So upgrading to this midnight-js already forces a recompile with `compactc 0.33.0-rc.0`, which emits `compiler/contract-manifest.json` by default. The manifest therefore ships for free with the recompile that's already mandatory — `require` doesn't add a new breaking surface on top of it.

Consequently I'm keeping `require` as the default rather than staging through `warn` for a release: there's no population of "compatible-but-manifest-less" artifacts to protect (the only genuine residual is hand-assembled key directories, which the `verify: 'off'`/`'warn'` hatch already covers). I do agree the *error message* is worth improving to name those escape hatches for anyone who hits the edge case.

**On point 5 (encoded `circuitId` vs raw manifest key):** confirmed safe — the assumption holds. `assertSafeName` enforces `/^[a-zA-Z0-9._-]+$/`, and every character in that set is left untouched by `encodeURIComponent`, so the fetched URL path and the raw manifest lookup key can never diverge.

**On point 1 (surfacing the trust boundary in the TSDoc):** agreed, this is the one I'll fold in before merge — moving the "without `expectedManifestHash`, this detects corruption but not a malicious co-located manifest" caveat up from the design notes into the `verify` / `expectedManifestHash` doc comments.

Points 3 (`size` unused), 4 (unused compiler/language/runtime versions as a supply-chain assertion), and 5c (deterministic Node `byteOffset` negative test) I'll take as follow-ups — all fair.

VanessaPC (approved):

PR #1028, opened 2026-07-02
test(testkit-js): close MIP-0002 event coverage gaps

jtsang586:
## Summary
- **Unit**: add the missing test for the null-`contractEvents` subscription payload guard, bringing the provider package's events code to 100% line/branch coverage; consolidate the five Apollo subscribe mock casts into one typed helper
- **Un-skip** `UnshieldedSpend`/`UnshieldedReceive` e2e (their indexer decode bug midnight-indexer#1279 is fixed; requires indexer >= 4.4.0-pre-alpha.13, satisfied by the pinned devnet stack)
- **Consolidate the events e2e suites for CI cost**: 4 suites → 2, 12 emission proofs → 7 (~49 → ~25 runner-minutes; each test file is its own CI matrix job with its own Docker environment, and proofs dominate the cost)
  - `contracts.events.provider.it.test.ts`: the `PublicDataProvider` read surface — `types` narrowing, `fieldPrefixes`, inclusive block windows, subscription replay / live delivery / `fromId` resume / server-side `toBlock` completion — over a minimal 4-proof corpus; absorbs the former core (lifecycle) and shielded suites; the single live emission (ShieldedBurn) doubles as the `shieldedAmount`-alias decode coverage
  - `contracts.events.unshielded.it.test.ts`: the three address-carrying variants (Spend/Receive/Burn) — #1279 showed these decode paths fail independently; `UnshieldedMint` (no address, fields a subset of Spend's) is unit-covered only
  - Every cut assertion remains covered by the provider package unit suites (all 11 `LogEventType` variants)
- **`Misc` e2e stays skipped**: the proof itself passes in CI (~60s), but proving `emitMisc` makes the verbose proof server emit ~130MB of logs, which the Actions runner's per-line log processing cannot absorb within the 30-minute job timeout — the job is cancelled *after* the tests pass. Root cause is proof-server log volume (Ledger team); the test passes locally
- **Harden the shared events test env**: teardown tolerates `beforeAll` failure, and a deploy failure after container start shuts the containers down instead of leaking them

Remaining scope on #952 is the version-fallback decoder (`Misc`/version 0 for non-conforming payloads), which is not implemented yet — tracked there, not part of this PR. Part of #952.

## Test plan
- [x] Both events e2e suites pass locally against the pinned devnet images (provider 13 passed + 1 skipped, unshielded 3/3), incl. the skipped Misc test when run locally
- [x] Provider unit suite passes (207 tests), events files at 100% line/branch coverage
- [x] Lint clean, typecheck clean, build succeeds


sp-io (approved):
lgtm!

Issue #1025, opened 2026-07-02
CCC (#967): critical correctness bugs remaining on main after merge


sp-io:
## Context

Cross-Contract Call support landed in #967 (merged as `be6012c5`). Review comments on the PR were marked resolved but the underlying issues were **not actually fixed** in the merged code. This issue tracks the **critical correctness bugs** that remain on `main` and need a follow-up fix PR. Line references are against `be6012c5`.

These are ordered by severity — the first two produce runtime failures for users; the third is breaking/lossy behavior.

Companion issues: robustness/API/cleanup #1026, test coverage #1027.

---

### 1. Coherence-guard race in `queryZSwapAndContractState` — intermittent tx-build failures + broken historical reads

`packages/indexer-public-data-provider/src/provider.ts:203`

The reworked query reads `contract.state` "as of" a pinned block, but pulls `zswapState`/`ledgerParameters` from `actions(limit:1)[0]`. Per the indexer schema, `Contract.actions` is **newest-first and not offset-filtered** (the code comment concedes this). A guard throws when `contract.state !== actions[0].state`. It fires in two non-error situations:

- **Race:** `createUnprovenCallTx` pins `blockHash = latestBlock.hash` and then issues this query as a **separate** round-trip. Any transaction indexed against the same contract in between makes the two states differ → build fails with a misleading *"a historical offset was likely supplied"* error. This will hit any actively-used contract under concurrent load.
- **Historical reads:** any caller of the public `queryZSwapAndContractState(addr, { type: 'blockHash', blockHash: <older> })` now always throws once the contract has newer actions. Pre-PR, `contractAction(address, offset)` returned a coherent historical triple.

**Fix direction:** read `zswapState`/`ledgerParameters` from an offset-filtered action (as the old `contractAction` query did), or split the API so the block-pinned CCC read and the latest-state read are distinct, typed operations.

---

### 2. Cached `TransactionContext` states mixed with a freshly-queried latest block

`packages/contracts/src/unproven-call-tx.ts:377`

[...]

PR #1031, opened 2026-07-02
fix: Critical correctness issues for cross-contract calls

JosephDenman:
Addresses issues mentioned in https://github.com/midnightntwrk/midnight-js/issues/1025


JosephDenman:
We're waiting on a fix on the indexer side for the indexer state query consistency issue.

sp-io (approved):

Issue #984, opened 2026-06-19
fix(http-client-proof-provider): per-call timeout is amplified ~8× by retry layer (effective ceiling is not timeout)


HarleysCodes:
## Summary

`httpClientProofProvider().proveTx()` claims to honor a per-call `ProveTxConfig.timeout`, but even with the parameter correctly wired through, the **effective wall-clock ceiling under transient server failures is approximately `timeout × 8`**, not `timeout`. The retry layer in the underlying `httpClientProvingProvider` amplifies any configured timeout exponentially and is not bounded by it.

This means a caller passing `proveTxConfig.timeout: 5000` to bound a 5-second proving operation can actually wait up to ~40 seconds before the call returns an error, when the proof server returns intermittent 5xx responses. That's not what the parameter documents.

## Environment

- `@midnight-ntwrk/midnight-js-http-client-proof-provider`, current `main`
- Affects both `httpClientProofProvider` (high-level) and `httpClientProvingProvider` (low-level) — the latter is reachable directly by callers using `ProvingProvider` directly
- File: `packages/http-client-proof-provider/src/http-client-proving-provider.ts`

## Affected code

`packages/http-client-proof-provider/src/http-client-proving-provider.ts`, lines 28-34:

```typescript
const retryOptions = {
  retries: 3,
  retryDelay: (attempt: number) => 2 ** attempt * 1_000,
  retryOn: [500, 503]
};
```

And the call site, lines 53-58:

```typescript
const response = await fetchRetry(url, {
  method: 'POST',
  body: new Uint8Array(payload),
  headers: { 'Content-Type': 'application/octet-stream', ...headers },
  signal: AbortSignal.timeout(timeout)
});
```

## The math

`fetch-retry` calls `makeHttpRequest` repeatedly. For each attempt it creates a fresh `AbortSignal.timeout(timeout)`. So each attempt has its own per-request timeout — but `retryDelay` adds wall-clock between attempts *that is not bounded by `timeout` at all*.

Worst case for a 5xx-returning server, with `timeout = 5_000`:

| Attempt | Per-attempt wait (capped at `timeout`) | `retryDelay` between attempts | Cumulative wall-clock |
|---------|----------------------------------------|--------------------------------|-----------------------|
| 1       | up to 5s                                | —                              | 0 – 5s                 |
| retry   | up to 5s                                | 1s                             | 6 – 11s                |
| retry   | up to 5s                                | 2s                             | 13 – 18s               |
[...]

Issue #989, opened 2026-06-23
Fork PR CI fails: ghcr.io auth not granted to pull_request events from forks


HarleysCodes:
## Fork PR CI fails at `Log in to GitHub Container registry` step

**Affected:** External contributors submitting PRs from forks against `midnightntwrk/midnight-js` (and likely the other Midnight repos using the same CI pattern).

**Symptom:** All three "post-checkout" CI jobs short-circuit on the first gated step (`Log in to GitHub Container registry`), so `yarn build`, `yarn lint`, `yarn test`, and testkit builds never run. PR shows 3 failing checks (`Validate PR Title`, `Build, Lint and Unit/Integration Tests`, `Build Testkit.js / build / ci`) but the underlying cause is identical: a fork PR cannot authenticate to `ghcr.io` with the workflow's `GITHUB_TOKEN`, because fork PRs run with a read-only token by default and `read:packages` is not in the default token scopes for `pull_request` events from forks.

**Example:** [PR #983](https://github.com/midnightntwrk/midnight-js/pull/983) — all three failing checks show the same root cause; build/test never ran. (The contributor's code is fine; the issue is the CI can't reach the registry to even compile it.)

**Workaround in CONTRIBUTING.md is broken:** `CONTRIBUTING.md` directs contributors to "Fork the Repository", but forking currently produces a permanently-red CI run that no contributor can fix from their side.

**Suggested fix (any of these):**
1. Add a `pull_request_target` workflow that runs CI with trusted secrets, gated by an `allowlist` of approved forks / paths (security trade-off — fork code gets secret access; mitigate with `permissions:` scope down + path filters).
2. Pre-build ghcr.io test images and publish them to a public registry (no auth needed) — best for security, but requires a build pipeline change.
3. Run CI only on `push` to the upstream repo's PR branches; require contributors to open the PR from a branch in the main repo (granted via temporary write access), then revert on close. Annoying for contributors.
4. Document the workaround explicitly in CONTRIBUTING.md: contributors must request maintainer approval to run CI before opening the PR, and the "Fork the Repository" instruction should be removed or qualified.

**Recommendation:** Option 1 (`pull_request_target` + scope-down + path-filter) is the standard pattern for OSS projects that take external PRs and need registry access. Option 4 is a one-line docs change if option 1 is too invasive.

**Blocker:** Real contributions are currently being silently rejected. This is information the maintainers need to know about.

sineadplunkett:
[...]

PR #1054, opened 2026-07-07
fix(midnight-js): honor per-call proveTxConfig.timeout

### LINKED ISSUE #974: @midnight-ntwrk/midnight-js-http-client-proof-provider: proveTx ignores its ProveTxConfig (per-call timeout has no effect)

sp-io:
## Summary

Re-opens #983 from a repo branch so CI (which needs secrets unavailable to fork PRs) can run. Original work by @HarleysCodes; rebased onto current `main` and reconciled with the cross-contract-call changes (#967) that landed since.

`httpClientProofProvider().proveTx(unprovenTx, proveTxConfig?)` accepted a `ProveTxConfig` but **silently discarded it** — the underlying proving provider was built once at construction time with a fixed timeout, so `proveTxConfig.timeout` had no effect.

- `httpClientProvingProvider` now returns a `TimeoutAwareProvingProvider` whose `check`/`prove` take an optional per-request `overrideTimeout`. The extra param is optional and trailing, so the type stays assignable to `ProvingProvider` — **no casts** needed anywhere.
- `proveTx` resolves the timeout (`per-call > construction-time > DEFAULT_TIMEOUT`) and wraps the once-built provider with a closure that pre-binds it. The inner provider is still built **once**, so URL validation (`InvalidProtocolSchemeError`) and the insecure-URL warning stay eager and fire only once (fixes the regression flagged in the original #983 review).
- `ProveTxConfig.timeout` JSDoc clarified: per-request timeout, **not** a wall-clock ceiling for the whole `proveTx` (internal retry/backoff can exceed it — tracked separately by #984/#1023).

### Reconciliation with `main`
The original PR conflicted with #967. Reconciled by keeping `main`'s `ZKConfigProvider<K> | ZKConfigRegistry` signatures, `makeKeyMaterialResolver`, and the new `lookupKey` on the provider — the per-call wrapper forwards `lookupKey` too.

### Note on the original tests
The original PR's `stubTx` used a no-op `prove` mock that never drove the proving provider, so its per-call-timeout assertions couldn't actually observe the threaded value. The stub now invokes `provingProvider.prove(...)` exactly as the real `UnprovenTransaction.prove` does, so the timeout is verified end-to-end. Added `overrideTimeout` unit tests on the proving provider (spying on `AbortSignal.timeout`).

## Test plan
- [x] Tests written for per-call timeout precedence (per-call > construction > default) and `overrideTimeout` threading
- [x] All existing tests pass — 38 passed (no regressions); existing `httpClientProvingProvider` coverage retained
- [x] Lint clean (`eslint`), full `turbo build` succeeds (16/16), `typecheck:tests` passes

Closes #974
Supersedes #983


HarleysCodes:
thanks szymon — really appreciate you taking the time to rebase and land this from a repo branch. the CI secrets workaround isn't a quick thing to fix on the fork side, so opening #1054 was the right call.

quick read on the reconciliation:

- the TimeoutAwareProvingProvider shape (trailing optional `overrideTimeout`) is cleaner than what i had in #983 — no cast, stays assignable to `ProvingProvider`, much better.
- keeping main's `lookupKey` and forwarding it through the per-call wrapper is the bit i would have gotten wrong on a rebase; glad you caught it.
- the test rewrite (stub now calls `provingProvider.prove(...)` end-to-end) is the right observation — my `stubTx` was a no-op mock and could only assert on construction, not on the timeout actually being threaded through. worth keeping as-is.

one thing for the close: i'll close #983 on my side with a pointer here once you merge, so it doesn't sit open as a duplicate. dropping it now would be premature (it still has #989 cross-refs that may matter if fork CI ever gets fixed) — the close should land with the merge.

— harley

VanessaPC (approved):
LGTM

sp-io:
@HarleysCodes Our automation closed the PR you raised. Huge thanks for the contribution, and I am really sorry that it took so much time.

HarleysCodes:
thanks szymon — really appreciate the merge and the kind note. no worries on the timing, you handled it correctly (CI secrets on fork PRs is a maintainer-side fix, not a contributor-side workaround).

closing #983 now with a pointer here. thanks again for picking it up and reconciling with #967 cleanly — the TimeoutAwareProvingProvider shape and the end-to-end test rewrite were both improvements over what i had. 🤝

PR #1063, opened 2026-07-07
fix(types): createProofProvider honors proveTxConfig.timeout (#1061)

### LINKED ISSUE #1061: createProofProvider silently drops proveTxConfig.timeout — same defect class as #974, on the canonical factory used by dappConnectorProofProvider
## The bug

`createProofProvider()` in `packages/types/src/proof-provider.ts` accepts a `proveTxConfig?: ProveTxConfig` parameter per the `ProofProvider` interface contract, but the implementation declares only the first parameter and silently drops `proveTxConfig` at runtime. Every `ProofProvider` built via `createProofProvider()` — including the wallet-backed `dappConnectorProofProvider()` — exhibits the same bug: callers passing `proveTxConfig.timeout` get zero effect, with no warning.

This is the same defect class as #974 (fixed on the HTTP-client side by PR #1054), but on the canonical factory used by `dappConnectorProofProvider` and any consumer building a `ProofProvider` from a low-level `ProvingProvider`. The HTTP-side fix did not propagate here.

## Reproduction (no network, no proof server, ~10 lines)

```ts
import { createProofProvider } from '@midnight-ntwrk/midnight-js-types';
import type { UnprovenTransaction, UnboundTransaction, ProvingProvider } from '...';

let observedOverrideTimeout: number | undefined;
const fakeProvingProvider: ProvingProvider = {
  async check(_p: Uint8Array, _k: string, overrideTimeout?: number) {
    observedOverrideTimeout = overrideTimeout;
    return [0n];
  },
  async prove(_p: Uint8Array, _k: string, _b?: bigint, overrideTimeout?: number) {
    observedOverrideTimeout = overrideTimeout;
    return new Uint8Array();
  }
};

const fakeUnprovenTx = {
  prove: async (pp: ProvingProvider) => {
    await pp.prove(new Uint8Array(), 'key-1');
    return {} as UnboundTransaction;
  }
} as unknown as UnprovenTransaction;

const pp = createProofProvider(fakeProvingProvider);
await pp.proveTx(fakeUnprovenTx, { timeout: 5_000 });

console.log(observedOverrideTimeout); // undefined — per-call timeout silently dropped
```

Expected: `observedOverrideTimeout === 5_000`.
Actual: `observedOverrideTimeout === undefined`.

The same reproduction works against any `dappConnectorProofProvider(api, zkConfig, costModel)` instance — the wallet-backed proving provider receives `undefined` for `overrideTimeout` on every `proveTx` call regardless of what the caller passes.

## Impact

- **Wallet-backed DApp flows silently ignore `proveTxConfig.timeout`.** A dApp developer writing `await pp.proveTx(tx, { timeout: 60_000 })` against a `dappConnectorProofProvider` to bound a slow wallet-side proof gets zero timeout enforcement — the wallet's internal proving-provider timeout (or a default) is what actually bounds the call. The intended bound is silently violated.
- **Inconsistent behavior across `ProofProvider` factories.** `httpClientProofProvider` honors `proveTxConfig.timeout`; `createProofProvider` (and its wrapper `dappConnectorProofProvider`) doesn't. Code that works against one factory breaks silently against the other.
- **API surface misleads consumers.** The `ProveTxConfig` JSDoc at `packages/types/src/proof-provider.ts:33–39` explicitly cites issue #974 as the reference for this parameter's semantics. A reader of the docs would reasonably believe the parameter works everywhere; it doesn't.

## Suggested direction (not a required solution)

1. **Mirror the `httpClientProofProvider` pattern in `createProofProvider`** — accept `proveTxConfig`, thread per-call overrides through to the underlying `ProvingProvider.check` / `ProvingProvider.prove` calls. Minimal API change, restores the contract.
2. **Validate and reject** — if threading overrides is out of scope, at minimum log a warning or throw when a non-`undefined` `proveTxConfig` is passed. The current "silently accept and drop" is the worst possible failure mode.

## File:line references

- `packages/types/src/proof-provider.ts:46–56` — the interface contract
- `packages/types/src/proof-provider.ts:66–73` — the buggy implementation (declares only one parameter)
- `packages/dapp-connector-proof-provider/src/dapp-connector-proof-provider.ts:35–42` — the canonical caller that inherits the bug
- `packages/http-client-proof-provider/src/http-client-proof-provider.ts:74–119` — the asymmetric correct implementation (fixed by PR #1054)

## Related

- #974 — original issue for the per-call timeout; fixed by PR #1054 on the HTTP-client side
- PR #1054 — the HTTP-side fix that didn't propagate to `createProofProvider`

HarleysCodes:
## What this fixes

Closes #1061.

The companion fix to PR #1054. `createProofProvider()` in `packages/types/src/proof-provider.ts` previously declared `proveTx(unprovenTx)` on its returned `ProofProvider`, silently dropping the `proveTxConfig.timeout` argument the caller passes. Every `ProofProvider` built via `createProofProvider()` — including the wallet-backed `dappConnectorProofProvider()` in `packages/dapp-connector-proof-provider` — inherited the drop.

This change makes `createProofProvider` honor the per-call timeout the same way `httpClientProofProvider` does, closing the asymmetry between the two factories on the same `ProofProvider` interface.

## What changed

**`packages/types/src/proof-provider.ts`** (+98 / −5)

1. **Adds `TimeoutAwareProvingProvider` interface.** Extends `ProvingProvider` with an optional trailing `overrideTimeout` parameter on `check` / `prove`. Stays assignable to `ProvingProvider` (callers that don't need per-request control can ignore it). The Midnight `httpClientProvingProvider` conforms; custom providers can opt in by extending this interface.

2. **Adds `isTimeoutAwareProvingProvider` runtime feature-detect.** Uses `prove.length >= 4` (arity check) to narrow `ProvingProvider` to `TimeoutAwareProvingProvider`. The Midnight protocol `ProvingProvider.prove` declares 3 positional parameters; the timeout-aware extension adds a 4th.

3. **`createProofProvider` threads the per-call timeout when supported.** When the caller passes `proveTxConfig.timeout` AND the underlying provider is timeout-aware, the implementation wraps the provider so every circuit-level `check` / `prove` inside `proveTx` is called with the per-call timeout. When the caller omits `proveTxConfig` OR the underlying provider doesn't support the override, passes through unchanged (the underlying provider's own default remains in effect — graceful drop, not a hard error).

4. **Forwards `lookupKey` on the wrapper.** The upstream `ProvingProvider` interface includes `lookupKey` (used by the transaction prover to resolve key material). The wrapper forwards it unchanged so the timeout-threading fix doesn't drop the upstream surface. Matches `httpClientProofProvider`'s pattern.

## Tests

**`packages/types/src/test/proof-provider.test.ts`** (new file, +175 lines, 6 tests)

- `proveTx` exists and is a function on the returned `ProofProvider`
- `proveTx` does not call `proveTxConfig` when omitted — passes through
- `proveTx` threads `proveTxConfig.timeout` into both `check` and `prove` on a timeout-aware provider (the core fix)
- `proveTx` threads a distinct timeout per call (no shared mutable state across calls)
- `proveTx` passes through unchanged when the underlying provider doesn't conform to `TimeoutAwareProvingProvider` (graceful drop, no error)
- `proveTx` does not mutate the caller-supplied `proveTxConfig` (defensive)

## Verification

I could not run the full test suite locally — `yarn install` requires `GITHUB_TOKEN` for the `@midnight-ntwrk` private npm scope (same auth wall as #989), so the dependencies don't install. Both files pass `node --check` syntax validation.

The CI in this repo will exercise it cleanly once it lands: the change is additive (a new interface + a runtime detect), the wrapper pattern is the same shape `httpClientProofProvider` already uses post-PR #1054, and the test mocks the `ProvingProvider` shape directly so no network or proof server is required.

If CI flags any issue with the `prove.length` arity check or the `lookupKey` forwarding, happy to address.

## Related

- #974 — original issue for per-call `ProveTxConfig.timeout` (silent drop)
- PR #1054 — the HTTP-client side fix (`httpClientProofProvider`)
- #1061 — the companion bug filing on the canonical-factory side

## Out of scope for this PR

- A future cleanup could move the local `TimeoutAwareProvingProvider` in `packages/http-client-proof-provider/src/http-client-proving-provider.ts` to extend the canonical one in the types package. Skipped here to keep the PR focused; happy to do as a follow-up.
- The retry-budget cap from issue #984 (the deeper `ProveTxConfig.timeout` semantics: making the timeout a wall-clock ceiling across `fetch-retry` attempts, not just per-request). That's a separate change in `httpClientProvingProvider`'s retry-options configuration and unrelated to this threading fix.


packages/types/src/proof-provider.ts
```diff
@@ -54,19 +54,112 @@ export interface ProofProvider {
...
+  const perCallTimeoutSupported = isTimeoutAwareProvingProvider(provingProvider);
+
+  return {
+    async proveTx(
+      unprovenTx: UnprovenTransaction,
+      proveTxConfig?: ProveTxConfig
+    ): Promise<UnboundTransaction> {
+      const perCallTimeout = proveTxConfig?.timeout;
+
+      if (perCallTimeout === undefined || !perCallTimeoutSupported) {
```
sp-io, line 158: **This branch never fixes the case #1061 names.** `createProofProvider` has exactly one production caller — `dappConnectorProofProvider`, which passes `dappConnectorProvingProvider`. That provider's `prove` is declared with **3 parameters** (`packages/dapp-connector-proof-provider/src/dapp-connector-proving-provider.ts` — `prove: (serializedPreimage, keyLocation, overwriteBindingInput) => …`), so `prove.length === 3`, `isTimeoutAwareProvingProvider` returns `false`, and every `proveTx({timeout})` call falls into this graceful-drop path. Meanwhile the only timeout-aware provider (`httpClientProvingProvider`) routes through `httpClientProofProvider`, which has its **own** wrapper and never calls `createProofProvider`.

Net: the new timeout-threading branch below is dead code for every current in-repo provider, and the wallet-backed `dappConnectorProofProvider()` — the exact case the PR body says inherited the drop — still silently ignores `proveTxConfig.timeout`.

To actually close #1061, `dappConnectorProvingProvider` (and the wallet's `getProvingProvider`) need to be made timeout-aware in this same change; otherwise the PR's claim should be rescoped to "external custom timeout-aware providers only."

packages/types/src/proof-provider.ts
```diff
@@ -54,19 +54,112 @@ export interface ProofProvider {
...
+ * the underlying provider is used as-is.
+ */
+const isTimeoutAwareProvingProvider = (
+  pp: ProvingProvider
+): pp is TimeoutAwareProvingProvider => {
+  // `Function.prototype.length` returns the number of declared parameters before any rest
+  // parameter; an `overrideTimeout?: number` trailing parameter is counted. The Midnight
+  // protocol `ProvingProvider.prove` signature has 3 positional params (preimage, key,
+  // overwriteBindingInput); the timeout-aware extension adds a 4th.
+  return pp.prove.length >= 4;
```
sp-io: **Arity heuristic (`prove.length >= 4`) is fragile in both directions.** `Function.prototype.length` stops counting at the first parameter with a default value or a rest param. Verified empirically:

- A genuinely timeout-aware provider authored as `prove(preimage, key, overwriteBindingInput = undefined, overrideTimeout?)` reports `length === 2` → **false negative** → timeout silently dropped.
- Any provider that declares an unrelated 4th positional parameter reports `length === 4` → **false positive** → `createProofProvider` passes `perCallTimeout` (a millisecond number) into that foreign slot.

An explicit capability marker (a discriminant property / branded interface the provider opts into) would be robust where parameter arity is not.

sp-io (commented):
Two findings on the timeout-threading logic — verified against every in-repo caller of `createProofProvider` and confirmed empirically with `Function.prototype.length`. The plumbing is correct and well-tested, but the branch it adds is currently unreachable by any real provider, so #1061's dapp-connector case is not actually fixed. Details inline.

HarleysCodes:
Pushed a follow-up addressing both findings — see the new commits at the head (`18d70ddc`). The two changes are:

**Finding 1 (branch was unreachable).** Made `dappConnectorProvingProvider` conform to `TimeoutAwareProvingProvider`:
- The returned provider now carries the new `TIMEOUT_AWARE_BRAND` brand.
- The wrapper accepts the optional trailing `overrideTimeout` parameter on `check`/`prove` and forwards it to the wallet's `getProvingProvider` result.
- Return type widened from `ProvingProvider` to `TimeoutAwareProvingProvider` (additive — structural assignability to `ProvingProvider` holds).

Honest caveat: the current upstream wallet's `getProvingProvider` does not accept the override, so the per-call timeout is **silently ignored at the wallet boundary** for now. The framework side is structurally ready, and wallet implementations can opt in to honor the override without further changes on this side. Documented this in the JSDoc.

**Finding 2 (arity heuristic was fragile).** Replaced `prove.length >= 4` with an explicit brand property:

```typescript
export const TIMEOUT_AWARE_BRAND: unique symbol = Symbol.for('@midnight-ntwrk/midnight-js-types/timeout-aware-proving-provider');

export interface TimeoutAwareProvingProvider extends ProvingProvider {
  readonly [TIMEOUT_AWARE_BRAND]: true;
  // ... existing signatures
}

const isTimeoutAwareProvingProvider = (pp: ProvingProvider): pp is TimeoutAwareProvingProvider => {
  return (pp as Partial<TimeoutAwareProvingProvider>)[TIMEOUT_AWARE_BRAND] === true;
};
```

`httpClientProvingProvider` and `dappConnectorProvingProvider` now declare `[TIMEOUT_AWARE_BRAND]: true` on the returned objects. Runtime detection tests the brand, which is immune to:
- default-valued trailing parameters (false negative in the arity heuristic — `Function.prototype.length` stops counting at the first default)
- unrelated 4th positional parameters (false positive)

Two new tests cover both directions:

1. `does not false-positive on a provider whose prove has 4 positional params but no brand` — verifies that a brandless provider is correctly NOT detected as timeout-aware.
2. `detects a provider that carries the brand but has prove.length < 4 (e.g. default-valued trailing param)` — verifies that a brand-aware provider with a default-valued `overrideTimeout` IS detected as timeout-aware.

**One note on the duplicate interface:** `TimeoutAwareProvingProvider` is now declared in both `packages/types/src/proof-provider.ts` (canonical) AND `packages/http-client-proof-provider/src/http-client-proving-provider.ts` (pre-existing local copy). I kept the local copy for now and added the brand to it, with a JSDoc note acknowledging the duplicate. A follow-up could consolidate by re-exporting from `@midnight-ntwrk/midnight-js-types` — that's the cleanup I flagged as "out of scope" in the original PR body. Happy to do it as a separate small PR if you prefer; flagging it as an open thread rather than expanding this one's scope.

Files changed in this push:
- `packages/types/src/proof-provider.ts` — added `TIMEOUT_AWARE_BRAND`, replaced arity heuristic, updated JSDoc
- `packages/types/src/test/proof-provider.test.ts` — updated fixtures to carry/exclude the brand, added two regression tests
- `packages/http-client-proof-provider/src/http-client-proving-provider.ts` — added brand to interface + returned object, imported `TIMEOUT_AWARE_BRAND`
- `packages/dapp-connector-proof-provider/src/dapp-connector-proving-provider.ts` — wrapper now conforms, return type widened

Diff is clean (+26/+8/+49/+71 lines, −6/−2/−27/−6 lines across the four files vs the previous PR head). Ready for re-review whenever you have a moment.

sp-io:
Thanks for the follow-up work on 18d70ddc, and sorry for the long silence -- I've been heads-down on a large feature under a tight deadline and this fell off my plate for far longer than it should have.

The branded `TimeoutAwareProvingProvider` approach is the right direction, and I want to give the re-review the attention it deserves rather than a rushed pass. I'll come back to it as soon as the current feature work lands. Thanks for your patience.

PR #1078, opened 2026-07-08
feat(midnight-js): add provider(options) factories for proof and zk-config providers

sp-io:
# Overview

Moves the provider layer toward a uniform `provider(options)` convention, so wiring a provider means passing one options object instead of remembering positional-argument order. This is additive and backward-compatible — existing positional/class usage keeps working.

- **`httpClientProofProvider`** — new object-options form with a flattened config (`{ url, zkConfigProvider, timeout?, headers? }`). The positional signature is retained as a `@deprecated` overload for one release cycle (same pattern as `indexerPublicDataProvider`).
- **`nodeZkConfigProvider(options)` / `fetchZkConfigProvider(options)`** — thin factory functions alongside the existing `NodeZkConfigProvider` / `FetchZkConfigProvider` classes. The classes stay exported (subclassing/composition untouched); only a consistent call-site is added.

Low-level `httpClientProvingProvider` is intentionally left positional and tracked as a follow-up — converting it re-couples internal test mocks and it is a lower-traffic "advanced scenarios" primitive.

## Submission Checklist

- [x] Self-reviewed the diff
- [ ] Reviewer requested
- [ ] Update README.md file (if relevant)
- [ ] Update documentation (if relevant)
- [x] No new todos introduced

## Test plan

- [x] TDD: tests written first, watched RED (wrong-reason ruled out), then GREEN
- [x] `httpClientProofProvider`: 46/46 tests pass; object-options wiring, flattened `timeout`/`headers`, per-call `proveTxConfig` precedence (#974), positional parity
- [x] `nodeZkConfigProvider`: 43/43 tests pass; instance, directory pass-through, integrity-option propagation
- [x] `fetchZkConfigProvider`: 35/35 tests pass; instance, eager URL-scheme validation, `fetchFunc` propagation
- [x] Lint clean, `tsc --noEmit` clean, no `any`/`unknown` casts introduced


packages/http-client-proof-provider/src/http-client-proof-provider.ts
```diff
@@ -66,22 +66,63 @@ const resolveTimeout = (
  * **Note:** The /prove-tx endpoint is NOT used. All proving is done through
  * individual circuit operations using /check and /prove endpoints.
  */
-export const httpClientProofProvider = <K extends string>(
+/**
+ * Options for {@link httpClientProofProvider}. The connection fields (`url`, `zkConfigProvider`)
+ * are combined with the underlying {@link ProvingProviderConfig} into a single flat object.
+ */
+export interface HttpClientProofProviderOptions<K extends string> extends ProvingProviderConfig {
```
VanessaPC: **Orphaned architecture JSDoc (moderate — doc regression).**

The large `/** Creates a high-level ProofProvider … @param url … @param zkConfigProvider … @returns … @remarks */` block now sits directly above this `interface`, with a second `/** Options for … */` block between it and the function. TSDoc/IDEs attach the *nearest* comment, so:

- the primary (object) overload of `httpClientProofProvider` ends up **undocumented**, and
- the `@param url/zkConfigProvider/config` tags now describe the wrong declaration.

Suggest moving the architecture block down to sit directly above the first `export function httpClientProofProvider(options …)` overload, and dropping/relocating the positional `@param` tags (they only apply to the deprecated overload).

packages/fetch-zk-config-provider/src/fetch-zk-config-provider.ts
```diff
@@ -166,3 +166,21 @@ export class FetchZkConfigProvider<K extends string> extends ZKConfigProvider<K>
...
+ */
+export interface CreateFetchZkConfigProviderOptions extends FetchZkConfigProviderOptions {
+  /** The endpoint to query for ZK artifacts. */
+  readonly baseURL: string;
+}
+
+/**
+ * Factory for {@link FetchZkConfigProvider} following the `provider(options)` convention.
+ */
+export function fetchZkConfigProvider<K extends string>(
```
VanessaPC, line 187: **`K` is not inferable here (minor — API ergonomics).**

`<K extends string>` appears only in the return type, never in a parameter, so there is no inference site. A call like `fetchZkConfigProvider({ baseURL })` silently widens `K` to `string`, and `getProverKey(circuitId: string)` loses circuit-ID narrowing unless the caller writes `fetchZkConfigProvider<'a'|'b'>({…})` explicitly. Same applies to `nodeZkConfigProvider`.

This is parity with the existing classes (same limitation), not a regression — but the factory doesn't improve it either. A one-line doc note telling callers to supply the type argument for narrowing would help. (Note the `httpClientProofProvider` object form does *not* have this issue — `K` is inferred from `zkConfigProvider`.)

packages/fetch-zk-config-provider/src/fetch-zk-config-provider.ts
```diff
@@ -166,3 +166,21 @@ export class FetchZkConfigProvider<K extends string> extends ZKConfigProvider<K>
     return createZKIR(bytes);
   }
 }
+
+/**
+ * Options for {@link fetchZkConfigProvider}. Combines the remote `baseURL` with fetch and integrity options.
+ */
+export interface CreateFetchZkConfigProviderOptions extends FetchZkConfigProviderOptions {
```
VanessaPC, line 176: **Naming inconsistency across the three options types (nit).**

`HttpClientProofProviderOptions`, `CreateFetchZkConfigProviderOptions`, and `NodeZkConfigProviderOptions` use three different conventions. The `Create…` prefix here is justified — a `FetchZkConfigProviderOptions` already exists for the constructor — but a short comment noting *why* would stop a future reader from “fixing” the inconsistency.

VanessaPC (commented):
Reviewed the `provider(options)` factories. Backward-compatible, well-scoped, and the tests are meaningful. Three findings inline — only the first is worth blocking on.

packages/http-client-proof-provider/src/http-client-proof-provider.ts
sp-io: Fixed in 097f5d63. Moved the architecture JSDoc block down so it sits directly above the primary `httpClientProofProvider(options)` overload (now `@param options` + `@returns`/`@remarks`), and relocated the positional `@param url/zkConfigProvider/config` tags onto the deprecated positional overload where they apply. The interface keeps its own `/** Options … */` doc.

packages/fetch-zk-config-provider/src/fetch-zk-config-provider.ts
sp-io: Fixed in 097f5d63. Added a `@typeParam K` note on both `fetchZkConfigProvider` and `nodeZkConfigProvider` telling callers to supply the type argument explicitly (e.g. `fetchZkConfigProvider<'a'|'b'>({…})`) to retain `getProverKey` narrowing, otherwise `K` widens to `string`.

packages/fetch-zk-config-provider/src/fetch-zk-config-provider.ts
sp-io: Fixed in 097f5d63. Added a comment on `CreateFetchZkConfigProviderOptions` noting the `Create` prefix disambiguates from the pre-existing `FetchZkConfigProviderOptions` (the class constructor options), so a future reader will not "fix" it into a name clash.

VanessaPC (approved):

Issue #876, opened 2026-05-07
Error: Insufficient funds — `midnight-js-contracts` pins every shielded coin to `DEFAULT_SEGMENT_NUMBER=0`, ignoring the partitioned transcript


0xisk:
**TL;DR.** When a Compact contract emits a user-bound shielded coin (e.g. an LP token mint on first `addLiquidity`) inside the fallible segment of a circuit, [`midnight-js-contracts`](https://www.npmjs.com/package/@midnight-ntwrk/midnight-js-contracts) still places the resulting `ZswapOutput` in segment `0` (guaranteed), because `zswapStateToOffer` hardcodes `DEFAULT_SEGMENT_NUMBER = 0` and never consults the segment information that `compact-runtime` already produced. The wallet's balancer then sees a negative imbalance for the freshly-minted token color in the guaranteed segment, can't find a matching wallet UTXO (none exists — the token was minted from nothing), and throws `Wallet.InsufficientFunds`. The bug is reproducible in both `4.0.2` and `4.0.4`.

---

### Context & Versions

- **Branch:** `feat/lunarswap-0.28.0-integration`
- **Bundler:** Vite `7.3.1` / esbuild `0.27.3`
- **Package manager:** pnpm `10.4.1`
- **Node:** `v22.22.0`
- **Network:** local (`undeployed`)
- **OS:** Fedora Linux

#### @midnight-ntwrk packages

| Package                                        | Version |
|------------------------------------------------|---------|
| @midnight-ntwrk/midnight-js-contracts          | 4.0.2 (also reproduced against 4.0.4) |
| @midnight-ntwrk/wallet-sdk-facade              | 4.0.0   |
| @midnight-ntwrk/wallet-sdk-capabilities        | 3.2.0   |
| @midnight-ntwrk/wallet-sdk-dust-wallet         | 4.0.0   |
| @midnight-ntwrk/wallet-sdk-shielded            | 3.0.0   |
| @midnight-ntwrk/wallet-sdk-unshielded-wallet   | 3.0.0   |
| @midnight-ntwrk/wallet-sdk-hd                  | 3.0.2   |
| @midnight-ntwrk/wallet-sdk-address-format      | 3.1.1   |
| @midnight-ntwrk/wallet-sdk-abstractions        | 2.1.0   |
| @midnight-ntwrk/compact-runtime                | 0.15.0  |
| @midnight-ntwrk/ledger-v8                      | 8.0.3   |
[...]

PR #1079, opened 2026-07-09
test(testkit-js): add regression test for #731 mintShieldedToken + receiveUnshielded combo

### LINKED ISSUE #731: mintShieldedToken + receiveUnshielded combo fails when wallet has only unshielded NIGHT
## Summary

Circuits that combine `mintShieldedToken` (to mint an NFT) with `receiveUnshielded` (to charge a fee in NIGHT) fail during wallet balancing. Each operation works independently — the bug only manifests when both are in the same circuit.

This is a common NFT pattern: mint a shielded token to the caller and collect a fee in unshielded NIGHT.

## Context

Post-network-reset, wallets funded via the faucet have **only unshielded NIGHT** (0 shielded balance). The wallet SDK's `balanceUnboundTransaction` runs shielded balancing before unshielded balancing. When `mintShieldedToken` is present, the shielded balancer attempts to create a Zswap offer but fails because there are no shielded funds — and the unshielded balancer never gets a chance to handle the `receiveUnshielded` fee.

## Versions

```
@midnight-ntwrk/midnight-js-contracts: 4.0.2
@midnight-ntwrk/midnight-js-*: 4.0.2
@midnight-ntwrk/ledger-v8: 8.0.3
@midnight-ntwrk/compact-js: 2.5.0
@midnight-ntwrk/compact-runtime: 0.15.0
@midnight-ntwrk/wallet-sdk-facade: 3.0.0
@midnight-ntwrk/wallet-sdk-shielded: 2.1.0
@midnight-ntwrk/wallet-sdk-unshielded-wallet: 2.1.0
@midnight-ntwrk/wallet-sdk-dust-wallet: 3.0.0
@midnight-ntwrk/wallet-sdk-hd: 3.0.1
compactc: 0.30.0
proof-server: 8.0.3
Network: preview (post-reset 2026-03-25)
```

### Contract (`fee-mint.compact`)

```compact
pragma language_version >= 0.20;
import CompactStandardLibrary;

witness localSecretKey(): Bytes<32>;

export pure circuit publicKey(sk: Bytes<32>): Bytes<32> {
    return persistentHash<Vector<2, Bytes<32>>>([pad(32, "feemint:pk:"), sk]);
}

export struct Metadata {
    name: Bytes<32>;
    description: Bytes<64>;
    owner: Bytes<32>;
}

export ledger metadata: Map<Bytes<32>, Metadata>;
export ledger nonce_counter: Counter;
export ledger nonce: Bytes<32>;
export ledger mint_fee: Uint<128>;

constructor(_nonce: Bytes<32>, _mint_fee: Uint<128>) {
    nonce = disclose(_nonce);
    mint_fee = disclose(_mint_fee);
}

export circuit mint(name: Bytes<32>, description: Bytes<64>): Bytes<32> {
    assert(!metadata.member(disclose(name)), "Name already taken");
    const caller_pk = disclose(publicKey(localSecretKey()));
    const token_metadata = Metadata { name: name, description: description, owner: caller_pk };

    // Mint a shielded token
    const coin_info = mintShieldedToken(
        pad(32, "fee-mint"), 1 as Uint<64>, nonce,
        left<ZswapCoinPublicKey, ContractAddress>(ownPublicKey())
    );
    metadata.insert(disclose(name), disclose(token_metadata));

    // Collect fee in unshielded NIGHT
    if (mint_fee > (0 as Uint<128>)) {
        receiveUnshielded(nativeToken(), mint_fee);
    }

    nonce = evolveNonce(nonce_counter, nonce);
    nonce_counter.increment(1);
    return coin_info.color;
}
```

### Control contracts that WORK

**Fee-only (no mintShieldedToken):**
```compact
export circuit register(name: Bytes<32>): [] {
    assert(!records.member(disclose(name)), "Name already taken");
    records.insert(disclose(name), disclose(publicKey(localSecretKey())));
    if (mint_fee > (0 as Uint<128>)) {
        receiveUnshielded(nativeToken(), mint_fee);
    }
}
```
Result: **WORKS** with `tokenKindsToBalance: 'all'` and fee > 0.

**Free mint (fee=0):**
Same `fee-mint.compact` contract deployed with `mint_fee = 0`.
Result: **WORKS** — `mintShieldedToken` alone is fine when no `receiveUnshielded` fires.

### With `tokenKindsToBalance: 'all'` (default)

```
Wallet.InsufficientFunds: Insufficient funds
  at Object.catch (wallet-sdk-shielded/dist/v1/Transacting.js:178:32)
```

The shielded wallet balancer runs first in `balanceUnboundTransaction`, sees the `mintShieldedToken` effects, tries to create a Zswap offer from shielded funds (which are 0), and throws. The unshielded balancer never runs.

### With `tokenKindsToBalance: ['unshielded', 'dust']`

```
RpcError: 1010: Invalid Transaction: Custom error: 186
```

Bypasses the shielded balancer, so the transaction is constructed and submitted. But the `mintShieldedToken` creates a shielded spend commitment that isn't properly balanced without shielded balancing — the chain rejects with `EffectsCheckFailure`.

### Wallet state at time of failure

```
Shielded NIGHT:   0 (0 tNIGHT)
Unshielded NIGHT: 1000000000 (1000 tNIGHT)
```

## Expected Behavior

The wallet SDK should handle the case where:
1. A circuit mints a shielded token (requires shielded balancing for the Zswap offer)
2. The same circuit calls `receiveUnshielded` (requires unshielded NIGHT from the wallet)
3. The wallet has only unshielded NIGHT (standard faucet-funded state)

This is the standard NFT-with-fee pattern and should work regardless of whether the wallet's NIGHT is shielded or unshielded.

## Related Issues

- #667 — `receiveUnshielded` FeeCalculation bug (fixed in ledger 8.0.3)
- #695 — Reverted `addCalls` approach that broke shielded ops (merged into 4.0.2)
- #702 — E2E test coverage gap: `mintShieldedToken` has no midnight-js E2E coverage
- #707 — `mintUnshieldedToken` error 186 (fixed in 4.0.2, different path)
- midnightntwrk/midnight-ledger#233 — `EffectsCheckFailure` for mint to different address (open)

sp-io:
## Summary

- Adds e2e regression coverage for #731 — minting a shielded token together with a fee leg in the **same** circuit.
- Root cause was filed against the wallet SDK (midnightntwrk/midnight-wallet#250) but closed as a **midnight-js** bug: *"the shielded output should've been in the same segment as the mint segment"* — the same segment-routing defect fixed by #876/#877 (`zswapStateToSegmentedOffer`). No existing test exercised the mint-shielded + fee combo.
- Adds a dedicated `fee-mint.compact` contract with **two** circuits, each reproducing a fee variant from the issue:
  - `mintWithUnshieldedFee` — `mintShieldedToken` + `receiveUnshielded` (original report).
  - `mintWithShieldedFee` — `mintShieldedToken` + `receiveShielded` (@robbyja's variant, which does **not** sidestep the bug).
- Both circuits carry a heavy pre-checkpoint write section (mirroring the #876 fixture) forcing the whole circuit into the fallible segment.
- Tests inspect the constructed `UnprovenTransaction` directly (like the #876 test) rather than relying on the wallet balancer's error, so they are independent of the wallet's shielded-vs-unshielded funding state. Both cases assert: transcript entirely fallible; `shieldedMints` present in the fallible segment; the minted output routed to the fallible offer (`guaranteedOffer` undefined, `fallibleOffer` defined) matching a fallible-transcript shielded commitment.
- Per-variant: the unshielded case has a fallible `unshieldedInputs` entry; the shielded case has **no** unshielded effects and claims >1 shielded commitment (minted output + received fee coin).

Pre-#877 the minted output was pinned to `DEFAULT_SEGMENT_NUMBER = 0` (guaranteed offer) — these tests fail on that code and pass on the fixed routing.

## Test plan

- [x] Contract compiles with compactc 0.33.0-rc.1 (`yarn compact-fee-mint`, 2 circuits); artifacts committed alongside the existing `shielded-fallible` fixture.
- [x] ESLint clean on new/modified files.
- [x] Project typecheck reports no type errors in the new files (only pre-existing `TS6059` rootDir notes, affecting every `test/**` file).
- [ ] e2e run (`yarn e2e`) — requires Docker (proof-server, indexer, node); to be verified in CI / by a maintainer.

Closes #731 once CI confirms the e2e run is green.


packages/fetch-zk-config-provider/src/fetch-zk-config-provider.ts
```diff
@@ -166,3 +166,21 @@ export class FetchZkConfigProvider<K extends string> extends ZKConfigProvider<K>
...
+ */
+export interface CreateFetchZkConfigProviderOptions extends FetchZkConfigProviderOptions {
+  /** The endpoint to query for ZK artifacts. */
+  readonly baseURL: string;
+}
+
+/**
+ * Factory for {@link FetchZkConfigProvider} following the `provider(options)` convention.
+ */
+export function fetchZkConfigProvider<K extends string>(
```
VanessaPC, line 187: **Scope: undocumented public-API change bundled in a test PR.**

This commit (`feat: add provider(options) factories`) adds new public API across three packages — `fetchZkConfigProvider`, `nodeZkConfigProvider`, the object-form `httpClientProofProvider` overload, plus new type exports and `index.ts` re-exports. None of it is mentioned in the PR description, which is entirely about the #731 regression test.

This is a semver-relevant surface addition, and the two concerns share no code. Recommend splitting the factory feature into its own PR (or at minimum documenting it in the description + changelog) so it gets reviewed as the API change it is, rather than being approved implicitly alongside a regression test.

packages/fetch-zk-config-provider/src/fetch-zk-config-provider.ts
```diff
@@ -166,3 +166,21 @@ export class FetchZkConfigProvider<K extends string> extends ZKConfigProvider<K>
     return createZKIR(bytes);
   }
 }
+
+/**
+ * Options for {@link fetchZkConfigProvider}. Combines the remote `baseURL` with fetch and integrity options.
+ */
+export interface CreateFetchZkConfigProviderOptions extends FetchZkConfigProviderOptions {
```
VanessaPC, line 176: **Naming inconsistency across the three factory option interfaces.**

For the same `provider(options)` convention, this one is `CreateFetchZkConfigProviderOptions` (with a `Create` prefix), while the others are `NodeZkConfigProviderOptions` and `HttpClientProofProviderOptions<K>` (no prefix). Pick one convention across the three packages — most likely dropping the `Create` prefix here.

packages/http-client-proof-provider/src/http-client-proof-provider.ts
```diff
@@ -66,22 +66,63 @@ const resolveTimeout = (
  * **Note:** The /prove-tx endpoint is NOT used. All proving is done through
  * individual circuit operations using /check and /prove endpoints.
  */
-export const httpClientProofProvider = <K extends string>(
+/**
+ * Options for {@link httpClientProofProvider}. The connection fields (`url`, `zkConfigProvider`)
+ * are combined with the underlying {@link ProvingProviderConfig} into a single flat object.
+ */
+export interface HttpClientProofProviderOptions<K extends string> extends ProvingProviderConfig {
```
VanessaPC: **Detached JSDoc.** The refactor from arrow-const to an overloaded `function` left the detailed doc block above (with `@param url`, `@param zkConfigProvider`, `@param config`, `@returns`) stranded directly above this `interface` instead of above the function. Those `@param`/`@returns` tags are now meaningless on an interface, and none of the three function overloads carries the descriptive doc anymore (the impl signature has none; the deprecated overload only has `@deprecated`). Move the descriptive block onto the primary object-form overload.

sp-io:
Accidental branch pull from #1078, removed and cleaned up

VanessaPC (approved):

PR #1105, opened 2026-07-13
docs(guides): add Vite WASM resolution guide for monorepo setups

### LINKED ISSUE #1052: WASM dual-instantiation in Vite monorepo requires resolve.dedupe workaround — no canonical solution documented
### Environment
- @midnight-ntwrk/midnight-js: ^4.1.1
- @midnight-ntwrk/onchain-runtime-v3: 3.0.0 (transitive, via ledger-v8)
- @midnight-ntwrk/compact-runtime: ^0.16.0
- vite: ^8.1.1
- vite-plugin-wasm: ^3.6.0
- Node.js: v24.12.0
- OS: macOS (Apple M3)

### Problem Description

In a monorepo where the compiled Compact contract output (`contract/src/managed/`) sits outside the Vite root (`frontend/`), Vite resolves WASM-containing packages from two separate `node_modules` trees:

- `repo-root/node_modules/` — used by `contract/src/managed/*/index.js`
- `frontend/node_modules/` — used by the Vite build

Each tree instantiates its own copy of the WASM binary. Because `instanceof` checks in wasm-bindgen use object identity tied to the specific WASM instance, cross-instance checks silently return `false`. The result is a `ContractRuntimeError: Error executing circuit` that appears only in the browser — CLI tests pass because they use only one resolution tree.

This was discovered on a production mainnet deployment, not a local testnet. CLI-based e2e tests passed completely, so the bug only surfaces when the full browser → Vite → managed contract path is exercised.

### Minimal Reproduction

Project layout:
repo-root/
├── contract/
│   ├── node_modules/        ← resolves @midnight-ntwrk/onchain-runtime-v3 here
│   └── src/managed/auction/contract/index.js
├── frontend/                ← Vite root
│   └── node_modules/        ← resolves @midnight-ntwrk/onchain-runtime-v3 here too
└── package.json

Steps:
1. Compile Compact contract — `contract/src/managed/` is populated
2. Import managed output from `frontend/` via relative path `../../contract/src/managed/...`
3. Call any circuit from the browser
4. Observe: `ContractRuntimeError: Error executing circuit 'createAuction'`
5. CLI execution of the same circuit succeeds

### Our Workaround

Adding `resolve.dedupe` to `vite.config.ts` forces a single copy:

```ts
resolve: {
  dedupe: [
    '@midnight-ntwrk/compact-runtime',
    '@midnight-ntwrk/onchain-runtime-v3',
    '@midnight-ntwrk/ledger-v8',
  ],
},
```

### This Does Not Look Standardized Upstream

`example-bboard`'s `bboard-ui/vite.config.ts` deals with what appears to be the same underlying class of problem (WASM identity across bundler resolution boundaries) but with a completely different, more involved approach — no `resolve.dedupe` at all:

```ts
// custom resolveId hook targeting compact-runtime → onchain-runtime-v3 specifically
{
  name: 'wasm-module-resolver',
  resolveId(source, importer) {
    if (
      source === '@midnight-ntwrk/onchain-runtime-v3' &&
      importer?.includes('@midnight-ntwrk/compact-runtime')
    ) {
      return { id: source, external: false, moduleSideEffects: true };
    }
    return null;
  },
},
```

...plus `rollupOptions.output.manualChunks` isolating `onchain-runtime-v3` into its own chunk, and `optimizeDeps.exclude` for the same package.

Two different Midnight projects independently arrived at two non-overlapping fixes for what appears to be the same root cause. That suggests there is no single documented, canonical solution.

### Expected Behavior

The SDK or official scaffolding should provide a single documented, canonical fix for WASM resolution in Vite monorepos. Without documentation clarifying which approach is correct — or whether both are needed for different scenarios — every new Midnight developer building a Vite monorepo will have to rediscover this the hard way.

One of:
- A single documented, canonical fix (with an explanation of when `resolve.dedupe` suffices vs. when the heavier custom-resolver/manualChunks approach is necessary), or
- If both are legitimate fixes for different scenarios, documentation clarifying which scenario calls for which

### Additional Context

Discovered while building **Midnight Private Auction** (https://github.com/pplmaverick/midnight-private-auction), a sealed-bid auction dApp on Midnight mainnet (contract `4fd31443997bd04bbf0b94e2ef3d5b0ff05479c4fb80bcac0dc74b2c763282e5`).

pplmaverick:
Closes #1052

Adds `docs/guides/vite-wasm-resolution.md` documenting two approaches
for resolving WASM dual-instantiation in Vite monorepo environments:

- **Option A**: `resolve.dedupe` (simpler, recommended for most projects)
- **Option B**: custom `resolveId` hook + `manualChunks` (used in example-bboard)

Also adds a link from README `Further Reading`.

Discovered while building [Midnight Private Auction](https://github.com/pplmaverick/midnight-private-auction)
on Midnight Mainnet — both workarounds are verified in production.


docs/guides/vite-wasm-resolution.md
```diff
@@ -0,0 +1,68 @@
...
+import wasm from 'vite-plugin-wasm'
+import topLevelAwait from 'vite-plugin-top-level-await'
+
+export default defineConfig({
+  plugins: [wasm(), topLevelAwait()],
+  resolve: {
+    dedupe: [
+      '@midnight-ntwrk/ledger-v8',
+      '@midnight-ntwrk/compact-runtime',
+      '@midnight-ntwrk/midnight-js-circuits',
```
sp-io: `@midnight-ntwrk/midnight-js-circuits` is not a real package. The repo publishes 14 packages (`midnight-js-compact`, `-contracts`, `-types`, etc.) and none is `midnight-js-circuits`; no source or `package.json` references it. A reader copying this `dedupe` array is deduping a package that doesn't exist (silently a no-op), which suggests the config was never verified against actual package names. Please replace with a real package or drop the entry.

docs/guides/vite-wasm-resolution.md
```diff
@@ -0,0 +1,68 @@
...
+```typescript
+import { defineConfig } from 'vite'
+import wasm from 'vite-plugin-wasm'
+import topLevelAwait from 'vite-plugin-top-level-await'
+
+export default defineConfig({
+  plugins: [wasm(), topLevelAwait()],
+  resolve: {
+    dedupe: [
+      '@midnight-ntwrk/ledger-v8',
```
sp-io: Please double-check this scope/version. The repo pins the ledger as `@midnightntwrk/ledger-v9` — note the scope is `@midnightntwrk` **without** the hyphen, and it's `v9`, not `@midnight-ntwrk/ledger-v8`. If the published scope is hyphen-less, this `dedupe` string won't match the installed module and won't dedupe anything. Worth confirming the exact published name/version and noting that the guidance is version-specific.

docs/guides/vite-wasm-resolution.md
```diff
@@ -0,0 +1,68 @@
...
+When building a browser dApp with Midnight JS in a Vite monorepo, the
+`@midnight-ntwrk/ledger-v8` package (which uses `wasm-bindgen`) can be
+instantiated multiple times across package boundaries. This causes subtle
+runtime failures where `instanceof` checks silently return `false` even
+for objects of the correct type.
+
+## Symptoms
+
+- ZK proof generation fails with no clear error message
+- `instanceof` checks on Midnight SDK objects return `false` unexpectedly
```
sp-io: Terminology: this says "Midnight **SDK** objects" while the rest of the doc correctly says "Midnight JS." Midnight JS is a framework, not an SDK — please keep the terminology consistent throughout.

README.md
```diff
@@ -218,6 +218,7 @@ yarn lint:fix
 
 - **[DEVELOPMENT.md](./DEVELOPMENT.md)** - Detailed guide for working on packages, debugging, and Turborepo commands
 - **[TROUBLESHOOTING.md](./TROUBLESHOOTING.md)** - Common issues and solutions
+- [Vite + WASM Resolution](docs/guides/vite-wasm-resolution.md) — Fixing WASM dual-instantiation in Vite monorepo setups
```
sp-io: This entry breaks the section's established format. The two existing bullets are bold with a hyphen separator and `./` paths (`- **[DEVELOPMENT.md](./DEVELOPMENT.md)** - …`); this line is non-bold, uses an em-dash, and omits `./`. Suggest matching: `- **[Vite + WASM Resolution](./docs/guides/vite-wasm-resolution.md)** - Fixing WASM dual-instantiation in Vite monorepo setups`

sp-io (commented):
Docs-only review. The guide is a welcome addition — the main asks are verifying that every package name in the `dedupe` array is real and correctly scoped, since that array is the core actionable content. Four inline notes below.

docs/guides/vite-wasm-resolution.md
pplmaverick: Fixed in the latest commits. Thank you for the thorough review!

docs/guides/vite-wasm-resolution.md
pplmaverick: Fixed in the latest commits. Thank you for the thorough review!

docs/guides/vite-wasm-resolution.md
pplmaverick: Fixed in the latest commits. Thank you for the thorough review!

README.md
pplmaverick: Fixed in the latest commits. Thank you for the thorough review!

pplmaverick:
Following up from the discussion in #1052 — happy to revise the scope of this PR to align with whatever format is most useful for the planned stopgap. Just let me know.

docs/guides/vite-wasm-resolution.md
```diff
@@ -0,0 +1,66 @@
...
+import topLevelAwait from 'vite-plugin-top-level-await'
+
+export default defineConfig({
+  plugins: [wasm(), topLevelAwait()],
+  resolve: {
+    dedupe: [
+      '@midnightntwrk/ledger-v9',
+      '@midnight-ntwrk/compact-runtime',
+    ],
+})
```
sp-io, line 44: This snippet is not valid — the `resolve: {` object opened on line 26 is never closed. As written, `})` closes only the `dedupe` array’s parent context and the `defineConfig(` call, leaving `resolve` unterminated, so a reader who copy-pastes this into `vite.config.ts` gets a `SyntaxError`. Since a copy-pasteable config is the whole point of this guide, this should be fixed before merge:

```typescript
  resolve: {
    dedupe: [
      '@midnightntwrk/ledger-v9',
      '@midnight-ntwrk/compact-runtime',
    ],
  },
})
```

sp-io:
Thanks for this, and for the thorough investigation in #1052 — the root-cause writeup there is what made the upstream tickets actionable.

To answer your question from #1052 (should this PR align with the planned stopgap rather than the broader guide): **yes, let's rescope it.** The root-cause fix belongs upstream (`LFDT-Minokawa/compact#611`, `midnightntwrk/midnight-ledger#644`), so what we want to land here is the cheap, interim half: a short troubleshooting note pointing at `resolve.dedupe`. Concretely:

1. **Drop Option B** (custom `resolveId` hook + `manualChunks`). That's the heavier bundler-plugin approach we deliberately don't want to document as a recommended path — an upstream fix would obsolete it, and it invites people to over-engineer. Keeping just `resolve.dedupe` (Option A) is enough.

2. **Frame `resolve.dedupe` as an interim workaround, not the canonical fix.** Until upstream tells us whether instance-scoped WASM identity is by-design or fixable, we don't want the doc to read as "this is the official solution." A one-line note linking #1052 + the two upstream issues is ideal.

3. **Align the `dedupe` list with the verified-working set.** Per your own answers in #1052, the confirmed failing types were in `@midnight-ntwrk/compact-runtime`, with `@midnight-ntwrk/onchain-runtime-v3` added as a cross-check. The current snippet lists `ledger-v9` + `compact-runtime` and drops `onchain-runtime-v3` — so a reader copying it wouldn't get the config you actually validated in production. Let's list the set you verified.

4. **Fix the snippet syntax** — see my inline comment; the `resolve: {` object is missing its closing brace, so the config as written won't parse.

Separately, we'll track the **fail-fast guard** (detect two runtime copies at provider init and throw an actionable error instead of the silent browser-only failure) as its own code change — no need to bundle it into this docs PR.

Once the scope is trimmed to Option A + the corrected list + the interim framing, this is good to merge. Thanks again!

pplmaverick:
Updated per your review — dropped Option B, reframed as interim workaround, corrected the dedupe list and fixed the snippet syntax.

pplmaverick:
All comments resolved. Updated per your review — dropped Option B, reframed as interim workaround, corrected the dedupe list and fixed the snippet syntax.

sp-io (approved):
Docs-only review — LGTM. ✅

Rescope is exactly what we discussed: dropped Option B, reframed as an interim workaround, and links the upstream trackers ([LFDT-Minokawa/compact#611](https://github.com/LFDT-Minokawa/compact/issues/611), [midnightntwrk/midnight-ledger#644](https://github.com/midnightntwrk/midnight-ledger/issues/644)) plus the #1052 write-up. Package names in the `dedupe` array are real and correctly scoped for the released line, the symptom description matches the confirmed `ContractRuntimeError` / `ChargedState`/`QueryContext` `instanceof` chain, and all prior review comments are resolved.

Thanks again for the thorough investigation — the root-cause analysis here is what made the upstream tickets actionable. 🙏

sp-io:
Thanks for this contribution, @pplmaverick! 🙏

We've brought your changes in on an internal branch as #1129 so CI can run with full token permissions (the PR-title check can't write its commit status from a fork PR — a repo-side CI limitation, not anything wrong with your PR). Your authorship is preserved via `Co-authored-by` on the commit, and you're credited in the PR description.

Closing this in favor of #1129. The docs guide and README link land exactly as you wrote them. Really appreciate you documenting the Vite WASM dual-instantiation workarounds!

PR #1129, opened 2026-08-03
docs: add Vite WASM resolution guide for monorepo setups

### LINKED ISSUE #1052: WASM dual-instantiation in Vite monorepo requires resolve.dedupe workaround — no canonical solution documented

sp-io:
## Summary
- Adds `docs/guides/vite-wasm-resolution.md` documenting two approaches for resolving WASM dual-instantiation in Vite monorepo environments:
  - **Option A**: `resolve.dedupe` (simpler, recommended for most projects)
  - **Option B**: custom `resolveId` hook + `manualChunks` (used in example-bboard)
- Adds a link from the README `Further Reading` section.

Originally contributed by @pplmaverick in #1105 (credited via `Co-authored-by`). Re-opened on an internal branch so CI runs with full token permissions — the fork PR's title check could not write its commit status. Closes #1052.

## Test plan
- [x] Docs-only change; no code paths affected
- [x] README link resolves to the new guide
- [x] Lint clean, no license header required (Markdown)

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,7 @@ yarn lint:fix
 
 - **[DEVELOPMENT.md](./DEVELOPMENT.md)** - Detailed guide for working on packages, debugging, and Turborepo commands
 - **[TROUBLESHOOTING.md](./TROUBLESHOOTING.md)** - Common issues and solutions
+- **[Vite + WASM Resolution](./docs/guides/vite-wasm-resolution.md)** — Fixing WASM dual-instantiation in Vite monorepo setups
 
 ## Contributing
 
--- a/docs/guides/vite-wasm-resolution.md
+++ b/docs/guides/vite-wasm-resolution.md
@@ -1,0 +1,66 @@
+# Vite + WASM Resolution in Midnight dApps
+
+When building a browser dApp with Midnight JS in a Vite monorepo,
+`@midnight-ntwrk/compact-runtime` and its underlying WASM layer
+(`@midnight-ntwrk/onchain-runtime-v3`) can be instantiated multiple times
+across package boundaries. Because wasm-bindgen ties class identity to the
+specific WASM instance that created an object, cross-instance `instanceof`
+checks silently return `false`, causing opaque circuit-execution failures
+that only surface in the browser — CLI and test paths pass because they use
+a single resolution tree.
+
+> **Status — interim workaround.** The root cause (instance-scoped WASM
+> class identity in wasm-bindgen) is being tracked upstream:
+> [LFDT-Minokawa/compact#611](https://github.com/LFDT-Minokawa/compact/issues/611)
+> and [midnightntwrk/midnight-ledger#644](https://github.com/midnightntwrk/midnight-ledger/issues/644).
+> The guidance below reflects the best known consumer-side mitigation until
+> upstream reaches a conclusion. See
+> [#1052](https://github.com/midnightntwrk/midnight-js/issues/1052) for
+> the full discovery write-up.
+
+## Symptoms
+
+- Circuit calls fail with `ContractRuntimeError: Error executing circuit '<id>'`
+- `instanceof` checks on Midnight JS objects (`ChargedState`, `QueryContext`)
+  return `false` unexpectedly
+- Errors appear only in browser / Vite builds, not in CLI or test runs
+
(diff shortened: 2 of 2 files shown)


iangregsondev (approved):

sp-io:
@pplmaverick Thanks again!

pplmaverick:
Thanks for bringing it in — glad it landed! 🙏

PR #1131, opened 2026-08-04
feat(indexer): add bounded timeout limits to watchFor transactions

efekrbas:
This PR resolves the documented TODO in PublicDataProvider by implementing bounded timeouts for all watchFor* indexer queries (e.g., watchForTxData). Previously, these observables waited indefinitely, causing silent dApp lock-ups on network drops or unmined transactions. I introduced a configurable maxWaitMs parameter backed by the RxJS timeout operator and implemented a typed IndexerTimeoutError class, significantly improving error handling and overall Developer Experience (DX). Lifecycle tests have also been added to ensure the timeout constraint properly rejects stalled queries.

### DIFF
--- a/packages/indexer-public-data-provider/src/provider.ts
+++ b/packages/indexer-public-data-provider/src/provider.ts
@@ -35,7 +35,9 @@ import type {
   UnshieldedBalances
 } from '@midnight-ntwrk/midnight-js-types';
 import { assertIsContractAddress } from '@midnight-ntwrk/midnight-js-utils';
+import { IndexerTimeoutError } from '@midnight-ntwrk/midnight-js-types';
 import * as Rx from 'rxjs';
+import { timeout } from 'rxjs';
 
 import {
   parseHexContractState,
@@ -270,21 +272,33 @@ export class IndexerPublicDataProvider implements PublicDataProvider {
       .then((maybeContractState) => (maybeContractState ? parseHexContractState(maybeContractState) : null));
   }
 
-  watchForContractState(contractAddress: ContractAddress): Promise<ContractState> {
+  watchForContractState(contractAddress: ContractAddress, maxWaitMs = 300_000): Promise<ContractState> {
     assertIsContractAddress(contractAddress);
     return Rx.firstValueFrom(
-      waitForContractToAppear(this.client, this.pollInterval)(contractAddress)(null).pipe(Rx.map(parseHexContractState))
+      waitForContractToAppear(this.client, this.pollInterval)(contractAddress)(null).pipe(
+        Rx.map(parseHexContractState),
+        timeout({
+          each: maxWaitMs,
+          with: () => Rx.throwError(() => new IndexerTimeoutError(`watchForContractState: Timed out after ${maxWaitMs}ms waiting for contract state of ${contractAddress}`))
+        })
+      )
     );
   }
 
-  watchForUnshieldedBalances(contractAddress: ContractAddress): Promise<UnshieldedBalances> {
+  watchForUnshieldedBalances(contractAddress: ContractAddress, maxWaitMs = 300_000): Promise<UnshieldedBalances> {
     assertIsContractAddress(contractAddress);
     return Rx.firstValueFrom(
-      waitForUnshieldedBalancesToAppear(this.client, this.pollInterval)(contractAddress).pipe(Rx.map(toUnshieldedBalances))
+      waitForUnshieldedBalancesToAppear(this.client, this.pollInterval)(contractAddress).pipe(
+        Rx.map(toUnshieldedBalances),
+        timeout({
+          each: maxWaitMs,
+          with: () => Rx.throwError(() => new IndexerTimeoutError(`watchForUnshieldedBalances: Timed out after ${maxWaitMs}ms waiting for unshielded balances of ${contractAddress}`))
+        })
+      )
     );
   }
 
-  watchForDeployTxData(contractAddress: ContractAddress): Promise<FinalizedTxData> {
+  watchForDeployTxData(contractAddress: ContractAddress, maxWaitMs = 300_000): Promise<FinalizedTxData> {
     assertIsContractAddress(contractAddress);
     return Rx.firstValueFrom(
       pollUntilPresent(
@@ -302,11 +316,16 @@ export class IndexerPublicDataProvider implements PublicDataProvider {
           return toFinalizedDeployTxData(contractAddress, transaction);
         },
         this.pollInterval
+      ).pipe(
+        timeout({
+          each: maxWaitMs,
+          with: () => Rx.throwError(() => new IndexerTimeoutError(`watchForDeployTxData: Timed out after ${maxWaitMs}ms waiting for deploy tx for ${contractAddress}`))
+        })
       )
     );
   }
 
-  watchForTxData(txId: TransactionId): Promise<FinalizedTxData> {
+  watchForTxData(txId: TransactionId, maxWaitMs = 300_000): Promise<FinalizedTxData> {
     return Rx.firstValueFrom(
       pollUntilPresent(
         this.client,
@@ -345,6 +364,11 @@ export class IndexerPublicDataProvider implements PublicDataProvider {
           };
         },
         this.pollInterval
+      ).pipe(
+        timeout({
+          each: maxWaitMs,
+          with: () => Rx.throwError(() => new IndexerTimeoutError(`watchForTxData: Timed out after ${maxWaitMs}ms waiting for tx ${txId}`))
+        })
       )
     );
   }
--- a/packages/indexer-public-data-provider/src/provider.ts
+++ b/packages/indexer-public-data-provider/src/provider.ts
@@ -34,10 +34,9 @@ import type {
   PublicDataProvider,
   UnshieldedBalances
 } from '@midnight-ntwrk/midnight-js-types';
+import { WatchTimeoutError } from '@midnight-ntwrk/midnight-js-types';
 import { assertIsContractAddress } from '@midnight-ntwrk/midnight-js-utils';
-import { IndexerTimeoutError } from '@midnight-ntwrk/midnight-js-types';
 import * as Rx from 'rxjs';
-import { timeout } from 'rxjs';
 
 import {
   parseHexContractState,
@@ -104,6 +103,25 @@ import type { ApolloHandle } from './transport';
  */
 const toBlockOffset = (config?: BlockHeightConfig | BlockHashConfig): InputMaybe<BlockOffset> =>
   config ? (config.type === 'blockHeight' ? { height: config.blockHeight } : { hash: config.blockHash }) : null;
+
+function validateMaxWaitMs(maxWaitMs?: number): void {
+  if (
+    maxWaitMs !== undefined &&
+    (typeof maxWaitMs !== 'number' || !Number.isFinite(maxWaitMs) || maxWaitMs <= 0 || maxWaitMs > 2147483647)
+  ) {
+    throw new RangeError(`maxWaitMs must be a positive integer <= 2147483647, got ${maxWaitMs}`);
+  }
+}
+
+function applyTimeout<T>(maxWaitMs: number | undefined, operation: string, subject: string): Rx.OperatorFunction<T, T> {
+  if (maxWaitMs === undefined) {
+    return Rx.identity;
+  }
+  return Rx.timeout({
+    first: maxWaitMs,
+    with: () => Rx.throwError(() => new WatchTimeoutError(operation, subject, maxWaitMs))
+  });
+}
 
 export class IndexerPublicDataProvider implements PublicDataProvider {
   private readonly handle: ApolloHandle;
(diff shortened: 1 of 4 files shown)


packages/indexer-public-data-provider/src/provider.ts
```diff
@@ -35,7 +35,9 @@ import type {
   UnshieldedBalances
 } from '@midnight-ntwrk/midnight-js-types';
 import { assertIsContractAddress } from '@midnight-ntwrk/midnight-js-utils';
+import { IndexerTimeoutError } from '@midnight-ntwrk/midnight-js-types';
 import * as Rx from 'rxjs';
+import { timeout } from 'rxjs';
```
sp-io: **Blocker (CI): `yarn lint` fails on this import block.** `simple-import-sort/imports` is error-level in this repo, and I verified this exact block errors: `@midnight-ntwrk/midnight-js-types` must sort before `...-utils`, and the second `rxjs` import duplicates the existing namespace import.

Suggested fix: move `IndexerTimeoutError` into the existing `@midnight-ntwrk/midnight-js-types` import area (above the `-utils` line), drop `import { timeout } from 'rxjs'`, and use `Rx.timeout` in the pipes — the `with:` callbacks already use the `Rx.` namespace, so this also removes the mixed style within one expression.

packages/indexer-public-data-provider/src/provider.ts
```diff
@@ -270,21 +272,33 @@ export class IndexerPublicDataProvider implements PublicDataProvider {
       .then((maybeContractState) => (maybeContractState ? parseHexContractState(maybeContractState) : null));
   }
 
-  watchForContractState(contractAddress: ContractAddress): Promise<ContractState> {
+  watchForContractState(contractAddress: ContractAddress, maxWaitMs = 300_000): Promise<ContractState> {
```
sp-io: **Critical: `maxWaitMs` is unvalidated, and every "disable the timeout" value backfires.** I tested this empirically with the repo's rxjs 7.8.2 (applies to all four methods):

| value | result |
|---|---|
| `Infinity` / `Number.MAX_SAFE_INTEGER` / `>= 2^31` | rejects after **~1ms** (Node clamps the timer with `TimeoutOverflowWarning`) — the error message then claims it waited `Infinityms` |
| `0` / `-1` / `NaN` | rejects immediately |

So a caller trying to restore the old wait-forever behavior gets an instant `IndexerTimeoutError` instead. Two fixes needed:

1. Fail fast at method entry: throw a `RangeError` for non-finite, non-positive, or `> 2^31 - 1` values.
2. Provide an explicit documented opt-out (e.g. treat `Infinity` as "skip the timeout operator entirely"), since there is currently no supported way to get the pre-PR behavior.

packages/indexer-public-data-provider/src/provider.ts
```diff
@@ -270,21 +272,33 @@ export class IndexerPublicDataProvider implements PublicDataProvider {
...
     assertIsContractAddress(contractAddress);
     return Rx.firstValueFrom(
-      waitForContractToAppear(this.client, this.pollInterval)(contractAddress)(null).pipe(Rx.map(parseHexContractState))
+      waitForContractToAppear(this.client, this.pollInterval)(contractAddress)(null).pipe(
+        Rx.map(parseHexContractState),
+        timeout({
+          each: maxWaitMs,
+          with: () => Rx.throwError(() => new IndexerTimeoutError(`watchForContractState: Timed out after ${maxWaitMs}ms waiting for contract state of ${contractAddress}`))
+        })
```
sp-io: **Use `first: maxWaitMs` instead of `each: maxWaitMs`** (all four sites). `each` bounds the gap *between* emissions and resets on every value; `first` bounds time to the first emission, which is what `firstValueFrom` needs.

They behave identically today only because `pollUntilPresent` / `waitForContractToAppear` pipe `filter → take(1)` upstream, so exactly one value ever reaches the timeout. If those helpers ever emit intermediate values, `each` silently resets and the "maximum wait" becomes unbounded again — the exact bug this PR fixes, reintroduced invisibly. `first` states the intent and is refactor-proof.

Also worth extracting the four near-identical `timeout({...})` blocks into one small helper — the four `with:` lines are 139–182 chars against the repo's 120-char print width, and a helper removes the duplication and makes the timeout logic testable once.

packages/types/src/public-data-provider.ts
```diff
@@ -346,55 +345,55 @@ export interface PublicDataProvider {
...
+   * - If a transaction is invalid and never appears, this will reject after `maxWaitMs`
    *
    * @param txId The identifier of the call or deployment of interest.
+   * @param maxWaitMs The maximum time to wait for the data in milliseconds. Defaults to 300,000 (5 minutes).
    *
    * @returns A promise that resolves with finalized transaction data when the transaction appears on-chain.
-   *          The promise never rejects due to timeout.
    */
-  watchForTxData(txId: TransactionId): Promise<FinalizedTxData>;
+  watchForTxData(txId: TransactionId, maxWaitMs?: number): Promise<FinalizedTxData>;
```
sp-io: **Critical: the default-on timeout can permanently lose a deployed contract's signing key.**

`submitTx` (`packages/contracts/src/submit-tx.ts:136`) calls `watchForTxData(txId)` with no `maxWaitMs`, so every deploy/call flow inherits the 5-minute default — and `maxWaitMs` is not plumbed through `submitTx` / `submitDeployTx` / `deployContract` / `findDeployedContract` options, so callers cannot opt out.

The dangerous sequence: `submitDeployTx` (`submit-deploy-tx.ts:94-107`) submits the transaction, then awaits finalization, and **only after that** persists the initial private state and the contract signing key. If the indexer lags past 5 minutes (congestion, resync, slow finalization), this now rejects — but the transaction can still finalize on-chain later. The contract exists, and the in-memory signing key is gone forever. The docs in `submit-tx.ts:113-120` describe exactly this as the abort-hazard requiring "manual recovery"; this PR makes it the default outcome.

Note the same file already recommends `submitTxAsync` for "manual finalization handling and timeout control" — the existing design intends timeouts to live at the application layer.

This is also an undeclared breaking change: the deleted JSDoc explicitly guaranteed "never rejects due to timeout" and forbade implementations from adding timeouts. The PR is typed `feat` with no `BREAKING CHANGE:` marker, so release tooling would cut a minor.

Suggested paths (either works):
1. Make the timeout **opt-in**: no default, `undefined` = wait forever (old behavior preserved, purely additive), or
2. Keep the default but plumb `maxWaitMs` through the contracts-layer options **and** persist the signing key before watching for finalization — and mark the release as breaking.

packages/types/src/public-data-provider.ts
```diff
@@ -346,55 +345,55 @@ export interface PublicDataProvider {
...
   /**
    * Monitors for any unshielded balances associated with a specific contract address.
+   * If unshielded balances are not found within `maxWaitMs`, throws `IndexerTimeoutError`.
    *
    * @param {ContractAddress} contractAddress - The address of the contract to monitor for unshielded balances.
+   * @param {number} maxWaitMs - The maximum time to wait for the data in milliseconds. Defaults to 300,000 (5 minutes).
    * @return {Promise<UnshieldedBalances>} A promise that resolves to the detected unshielded balances.
    */
-  watchForUnshieldedBalances(contractAddress: ContractAddress): Promise<UnshieldedBalances>;
+  watchForUnshieldedBalances(contractAddress: ContractAddress, maxWaitMs?: number): Promise<UnshieldedBalances>;
```
sp-io: **Doc consistency across the four methods:**

- These two methods say "throws `IndexerTimeoutError`" while `watchForDeployTxData`/`watchForTxData` say "the promise rejects" — nothing throws synchronously here, so "the returned promise rejects with an `IndexerTimeoutError`" is the accurate phrasing for all four (or use a uniform `@throws` tag, as the repo does elsewhere).
- The "Custom implementations MUST maintain this timeout behavior" clause appears only on `watchForDeployTxData`/`watchForTxData`. If honoring `maxWaitMs` is a contract requirement, it applies to all four; stating it on two implies the other two are exempt.
- Mixed JSDoc styles within the same PR: this block uses `@param {number} maxWaitMs - ...` and `@return {...}`, the other three use brace-free `@param maxWaitMs` — braces are redundant in TypeScript; please normalize.

**Also: "Defaults to 300,000" is promised on the interface, but an interface can't carry a default** — it lives only in `IndexerPublicDataProvider`, hardcoded four times. A custom implementation defaulting differently would silently violate this doc. Suggest exporting a single `DEFAULT_WATCH_TIMEOUT_MS = 300_000` constant from this package, referencing it via `{@link}` here and using it for all four implementation defaults.

**Design suggestion** while the API is still fresh: `opts?: { maxWaitMs?: number }` instead of a bare trailing number. The TODO this PR removes also mentioned retry limits (a future third parameter), an options bag leaves room for an `AbortSignal`, and `contractEventsObservable` below already uses this style (`opts?: { startAt?: ... }`).

packages/types/src/errors.ts
```diff
@@ -111,3 +111,13 @@ export class ImportConflictError extends PrivateStateImportError {
...
+
+/**
+ * An error thrown when an indexer polling query times out.
+ */
+export class IndexerTimeoutError extends Error {
+  constructor(message: string, options?: ErrorOptions) {
+    super(message, options);
+    this.name = 'IndexerTimeoutError';
+  }
+}
```
sp-io, line 128: **Naming and structure of the new error:**

1. **The name bakes one implementation into the provider-agnostic package.** `packages/types` defines the generic `PublicDataProvider` contract; "indexer" is one implementation (which already owns an `Indexer*Error` family in `packages/indexer-public-data-provider/src/errors.ts`). The interface JSDoc now obliges every future non-indexer implementation to throw an indexer-named error. Suggest `PublicDataProviderTimeoutError` or `WatchTimeoutError` (avoid bare `TimeoutError` — it collides with rxjs's and the DOM's). Worth deciding before this ships in the public API, since renaming later is breaking.

2. **Prefer structured fields over message-encoding**, matching the stronger patterns in this same file (`InvalidProtocolSchemeError`, `ImportConflictError` expose `readonly` fields and build their message internally):

```ts
constructor(
  public readonly operation: string,
  public readonly subject: string, // contractAddress or txId
  public readonly maxWaitMs: number,
  options?: ErrorOptions
) {
  super(`${operation}: timed out after ${maxWaitMs}ms waiting for ${subject}`, options);
  ...
}
```

This deduplicates the four near-identical template strings in `provider.ts` and lets callers read `maxWaitMs` programmatically (e.g. retry with backoff) instead of parsing the message. As written, the `options?: ErrorOptions` parameter is dead code — no call site passes it.

packages/indexer-public-data-provider/src/test/indexer-public-data-provider-lifecycle.test.ts
```diff
@@ -113,6 +113,30 @@ describe('indexerPublicDataProvider — config-object overload', () => {
...
+    const handle = createApolloClient(validated);
+    const provider = new IndexerPublicDataProvider(handle, validated.pollInterval);
+
+    // Mock watchQuery to return an observable that never emits any valid transaction data
+    vi.spyOn(handle.client, 'watchQuery').mockImplementation(() => {
+      return new Observable(() => {}) as any;
+    });
+
+    const fakeTxId = '00'.repeat(32) as unknown as Parameters<typeof provider.watchForTxData>[0];
+    const maxWaitMs = 50; // short timeout for test
```
sp-io: A few issues in this block:

- **`as any` is blocked by review in this repo** (CLAUDE.md mandatory rule 14 / Common Mistakes #1). Use `Rx.NEVER` instead of `new Observable(() => {})` and type the mock return against Apollo's `ObservableQuery` (a typed partial works) rather than erasing the type. The `as unknown as` cast on line 132 copies a pre-existing pattern, but rule 13 forbids it too — worth not propagating.
- The sibling test above uses `mockImplementationOnce`; this one uses persistent `mockImplementation` with no restore. The file's `afterEach(vi.clearAllMocks)` clears call history but does **not** restore implementations — it doesn't leak today only because `handle` is test-local. `mockImplementationOnce` (or `vi.restoreAllMocks`) makes it robust rather than accidental.
- Both comments restate the code (and "never emits any valid transaction data" is slightly off — it never emits *anything*); repo convention is self-describing code without comments.

packages/indexer-public-data-provider/src/test/indexer-public-data-provider-lifecycle.test.ts
```diff
@@ -113,6 +113,30 @@ describe('indexerPublicDataProvider — config-object overload', () => {
     await provider.dispose();
   });
 
+  test('watchForTxData throws IndexerTimeoutError when query times out', async () => {
```
sp-io: **Coverage: this tests 1 of the 4 changed methods, and only the failure path.** Good news first: I verified the test is genuine — the spy sits on the real `pollUntilPresent → watchQuery` path, so it exercises the actual timeout operator.

Gaps worth closing before merge:

1. **No happy-path test**: nothing proves the timeout *doesn't* fire when data arrives in time. A misconfigured operator (e.g. `first: 0`, or a future refactor breaking timer cancellation) would make every call reject even on success — and this test would still pass. One test where the mocked `watchQuery` emits a valid payload after ~10ms with a generous `maxWaitMs`, asserting the resolved value, closes the most likely production regression.
2. **The other three methods** (`watchForContractState`, `watchForUnshieldedBalances`, `watchForDeployTxData`) each have their own hand-written timeout block and different upstream observables — each is an independent regression surface with zero coverage. A `test.each` over all four (or extracting a shared timeout helper and testing it once) covers this cheaply.
3. Minor: asserting only the error class — adding `.rejects.toThrow(/watchForTxData.*50ms/)` would also pin the message content, which is the operational value of this feature. And this test sits inside the `describe('...config-object overload')` block it's unrelated to.

sp-io (commented):
Thanks for tackling this — ending the indefinite waits is the right feature, it resolves a standing TODO, and the one test included is a genuine behavioral test of the real code path. The inline comments carry the details (several verified empirically: the rxjs overflow behavior, the lint failure, `Promise.race` semantics); summary of the blocking items:

1. **Behavioral break with a data-loss consequence** — the 5-minute default reaches `submitDeployTx` unplumbed, which persists the contract signing key only after finalization; a slow indexer now means a contract deployed on-chain whose key was never stored. See the comment on `watchForTxData` for two possible fixes (opt-in default, or plumbing + persist-before-watch + breaking-change marker).
2. **`maxWaitMs` validation/opt-out** — `Infinity`, `MAX_SAFE_INTEGER`, `0`, negatives, and `NaN` all reject in ~1ms (verified against rxjs 7.8.2).
3. **Lint gate** — the import block fails `simple-import-sort` (verified with the repo's ESLint config).
4. `timeout({ each })` → `timeout({ first })`, error naming/structure, doc consistency, `as any` in the test, and coverage gaps — all inline.

Two stale spots outside this diff that now contradict the new contract and need updating in the same PR if the default stays:

- `packages/contracts/src/submit-tx.ts:97-105` — the "Blocking Behavior" block still promises "**waits indefinitely** ... will not return until" (and notably recommends `submitTxAsync` for timeout control, which suggests timeouts were designed to live at the app layer).
- `testkit-js/testkit-js-e2e/test/contracts.it.test.ts:325-333` and `:644-652` — tests titled/documented "should wait indefinitely without timing out" describe the removed contract (they'll keep passing, which makes the rot invisible). Those e2e runs also now inherit the 5-minute cap on this repo's known-slow CI — worth a thought.

Process nits: PR body doesn't follow the repo's `## Summary` / `## Test plan` template, and the `indexer` scope isn't in the commitlint `scope-enum` (warning-level only).

efekrbas:
@sp-io Thanks for the detailed review! I've pushed a commit that addresses all your points:

Timeouts are now strictly opt-in: Removed the 5-minute default entirely across all four watch methods, preserving the original "wait indefinitely" behavior unless maxWaitMs is explicitly provided.
Refactored Timeout Operator: Extracted the timeout logic into shared helpers (validateMaxWaitMs and applyTimeout) to eliminate duplication and ensure first: is used consistently instead of each:.
Validation & Fast Fail: We now validate maxWaitMs at method entry, throwing a RangeError synchronously for non-finite, non-positive, or overflowing values.
Error Naming & Structure: Renamed IndexerTimeoutError to WatchTimeoutError and updated its structure to expose operation, subject, and maxWaitMs as readonly fields.
Tests: Replaced all as any type erasures with properly typed partials. I also added a describe block testing both the failure paths across all four methods and a happy path confirming that the timer does not reject if data arrives in time.
Linting: Fixed the import sorting issue in provider.ts and dropped the unused timeout import.
Everything should be ready for another look!

sp-io:
Thanks for turning this around so quickly -- making the timeouts strictly opt-in and pulling the logic into shared helpers is exactly what I was asking for.

Apologies for the delay in getting back to you: I'm working against a tight deadline on a large feature and PR reviews have been backing up. I'll re-review this properly as soon as that clears. Thanks for your patience.

PR #1132, opened 2026-08-04
feat(proof-provider): throw typed ProofServerCommunicationError instead of generic Error

efekrbas:
## Title
feat(proof-provider): throw typed ProofServerCommunicationError instead of generic Error

## Background & Problem Statement
Currently, in `http-client-proof-provider`, if a network request to the proof server fails (e.g., HTTP 503 Service Unavailable or 400 Bad Request due to invalid circuit inputs), the `makeHttpRequest` utility throws a generic Javascript `Error` containing a stringified message with the response context. 

Because a generic error is thrown, dApp developers cannot programmatically or reliably inspect the `statusCode` to implement necessary fallback mechanisms (like automated retries for 5xx errors vs throwing explicit UI alerts for 4xx bad inputs). This forces developers to rely on fragile string manipulation against the error message, posing a severe DX flaw and limiting error-handling resiliency.

## Changes Made
- Introduced a strictly typed `ProofServerCommunicationError` class inside `@midnight-ntwrk/midnight-js-types` that exposes `url`, `statusCode`, and `statusText` as public readonly properties.
- Refactored `makeHttpRequest` in `packages/http-client-proof-provider/src/http-client-proving-provider.ts` to throw this structured error instead of a generic `Error`.

## Test Coverage
- [x] Updated existing assertion checks in `http-client-proving-provider.test.ts` to ensure that `check()` and `prove()` properly reject with `ProofServerCommunicationError` during simulated network failures.
- [x] Typechecked against `packages/types`.

## Developer Experience (DX) Benefits
SDK consumers can now safely catch and type-check failures against the proof server:
```typescript
try {
  await provider.prove(preimage, location);
} catch (error) {
  if (error instanceof ProofServerCommunicationError && error.statusCode >= 500) {
    // Safely retry or report backend issue
  }
}

### DIFF
--- a/packages/http-client-proof-provider/src/http-client-proving-provider.ts
+++ b/packages/http-client-proof-provider/src/http-client-proving-provider.ts
@@ -19,7 +19,13 @@ import {
   parseCheckResult,
   type ProvingKeyMaterial,
   type ProvingProvider} from '@midnight-ntwrk/midnight-js-protocol/ledger';
-import { InvalidProtocolSchemeError, type ZKConfigProvider, ZKConfigRegistry, zkConfigToProvingKeyMaterial } from '@midnight-ntwrk/midnight-js-types';
+import {
+  InvalidProtocolSchemeError,
+  ProofServerResponseError,
+  type ZKConfigProvider,
+  ZKConfigRegistry,
+  zkConfigToProvingKeyMaterial,
+} from '@midnight-ntwrk/midnight-js-types';
 import { warnIfInsecureRemoteUrl, ZkArtifactIntegrityError } from '@midnight-ntwrk/midnight-js-utils';
 import fetch from 'cross-fetch';
 import fetchBuilder from 'fetch-retry';
@@ -89,8 +95,10 @@ const makeHttpRequest = async (url: URL, payload: Uint8Array, timeout: number, h
   });
 
   if (!response.ok) {
-    throw new Error(
-      `Failed Proof Server response: url="${response.url}", code="${response.status}", status="${response.statusText}"`
+    throw new ProofServerResponseError(
+      response.url,
+      response.status,
+      response.statusText
     );
   }
 
--- a/packages/http-client-proof-provider/src/test/http-client-proving-provider.test.ts
+++ b/packages/http-client-proof-provider/src/test/http-client-proving-provider.test.ts
@@ -18,13 +18,14 @@ import {
   encodeContractKeyLocation,
   hashVerifierKey,
   InvalidProtocolSchemeError,
+  ProofServerResponseError,
   type ProverKey,
   type VerifierKey,
   ZKArtifactNotFoundError,
   type ZKConfig,
   type ZKConfigProvider,
   ZKConfigRegistry,
-  type ZKIR
+  type ZKIR,
 } from '@midnight-ntwrk/midnight-js-types';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
@@ -224,8 +225,13 @@ describe('httpClientProvingProvider', () => {
       vi.mocked(ledger.createCheckPayload).mockReturnValue(new Uint8Array([20, 21, 22]));
 
       await expect(provider.check(serializedPreimage, 'test-circuit')).rejects.toThrow(
-        /Failed Proof Server response/
+        ProofServerResponseError
       );
+      await expect(provider.check(serializedPreimage, 'test-circuit')).rejects.toMatchObject({
+        url: mockUrl,
+        statusCode: 500,
+        statusText: 'Internal Server Error'
+      });
     });
   });
 
@@ -345,8 +351,13 @@ describe('httpClientProvingProvider', () => {
       vi.mocked(ledger.createProvingPayload).mockReturnValue(new Uint8Array([30, 31, 32]));
 
       await expect(provider.prove(serializedPreimage, 'test-circuit')).rejects.toThrow(
-        /Failed Proof Server response/
+        ProofServerResponseError
       );
+      await expect(provider.prove(serializedPreimage, 'test-circuit')).rejects.toMatchObject({
+        url: mockUrl,
+        statusCode: 503,
+        statusText: 'Service Unavailable'
+      });
     });
   });
 
--- a/packages/types/src/errors.ts
+++ b/packages/types/src/errors.ts
@@ -111,3 +111,23 @@ export class ImportConflictError extends PrivateStateImportError {
     this.name = 'ImportConflictError';
   }
 }
+
+/**
+ * Thrown when the Proof Server returns a non-OK HTTP response. Network-level failures (connection refused, DNS, timeout) propagate as native fetch errors.
+ * @param url The URL of the Proof Server that failed.
+ * @param statusCode The HTTP status code returned.
+ * @param statusText The HTTP status text returned.
+ */
+export class ProofServerResponseError extends Error {
+  constructor(
+    public readonly url: string,
+    public readonly statusCode: number,
+    public readonly statusText: string
+  ) {
+    super(`Failed Proof Server response: url="${url}", code="${statusCode}", status="${statusText}"`);
+    this.name = 'ProofServerResponseError';
+  }
+}
+
+export const isProofServerResponseError = (e: unknown): e is ProofServerResponseError =>
+  e instanceof Error && e.name === 'ProofServerResponseError';
--- a/packages/http-client-proof-provider/README.md
+++ b/packages/http-client-proof-provider/README.md
@@ -173,7 +173,7 @@ httpClientProofProvider('ws://localhost:6300', zkConfigProvider);
 
 ### Server Errors
 
-HTTP errors throw with details about the failed request:
+HTTP errors throw a `ProofServerResponseError` (imported from `@midnight-ntwrk/midnight-js-types`) with details about the failed request:
 
 ```
 Failed Proof Server response: url="http://localhost:6300/prove", code="500", status="Internal Server Error"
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: packages/http-client-proof-provider/src/http-client-proving-provider.ts [hunk 1]
```diff
@@ -19,7 +19,7 @@ import {
   parseCheckResult,
   type ProvingKeyMaterial,
   type ProvingProvider} from '@midnight-ntwrk/midnight-js-protocol/ledger';
-import { InvalidProtocolSchemeError, type ZKConfigProvider, ZKConfigRegistry, zkConfigToProvingKeyMaterial } from '@midnight-ntwrk/midnight-js-types';
+import { InvalidProtocolSchemeError, type ZKConfigProvider, ZKConfigRegistry, zkConfigToProvingKeyMaterial, ProofServerCommunicationError } from '@midnight-ntwrk/midnight-js-types';
```
sp-io: **Blocker (lint):** `simple-import-sort/imports` fails here because `ProofServerCommunicationError` is appended at the end of the specifier list instead of sorted position. I verified by running the repo linter against this branch:

```
packages/http-client-proof-provider/src/http-client-proving-provider.ts
  16:1  error  Run autofix to sort these imports!  simple-import-sort/imports
packages/http-client-proof-provider/src/test/http-client-proving-provider.test.ts
  16:1  error  Run autofix to sort these imports!  simple-import-sort/imports
```

The same issue exists in the test file's import block. This will fail the Build + Lint + Unit Tests CI gate. Fix is one command:

```
yarn lint:fix
```

packages/types/src/errors.ts
```diff
@@ -111,3 +111,18 @@ export class ImportConflictError extends PrivateStateImportError {
     this.name = 'ImportConflictError';
   }
 }
+
+/**
+ * An error thrown when communication with the Proof Server fails.
+ * Captures the HTTP status code and URL for programmatic error handling.
+ */
+export class ProofServerCommunicationError extends Error {
```
sp-io: **Naming/contract:** the name and JSDoc promise more than the class covers. This error is thrown only for the `!response.ok` branch — an HTTP response was received. The failures a developer would most expect a "communication" error to cover (connection refused when the proof server isn't running — the most common local-dev failure — plus DNS, TLS, and `AbortSignal` timeout) propagate as raw `TypeError`/`DOMException` from fetch and will *not* match `instanceof ProofServerCommunicationError`. So `catch (e) { if (e instanceof ProofServerCommunicationError) showProofServerHint() }` silently misses exactly the case the hint exists for.

Since the symbol is brand new in this PR, renaming is free. Two options, in order of preference:

1. Rename to `ProofServerResponseError` — the shape already says this (required `statusCode`/`statusText` structurally cannot represent a no-response failure).
2. Keep the name but state the boundary explicitly in the JSDoc: *"Thrown when the Proof Server returns a non-OK HTTP response. Network-level failures (connection refused, DNS, timeout) propagate as native fetch errors."*

While touching the JSDoc, please also add `@param` tags for the three constructor parameters — the sibling `InvalidProtocolSchemeError` above documents each param, and this file is the TypeDoc'd public API surface.

packages/types/src/errors.ts
```diff
@@ -111,3 +111,18 @@ export class ImportConflictError extends PrivateStateImportError {
...
+  constructor(
+    public readonly url: string,
+    public readonly statusCode: number,
+    public readonly statusText: string
+  ) {
+    super(`Failed Proof Server response: url="${url}", code="${statusCode}", status="${statusText}"`);
+    this.name = 'ProofServerCommunicationError';
+  }
+}
```
sp-io, line 130: Two optional-but-cheap hardening suggestions for the new class:

1. **Dual-package-safe type guard.** `@midnight-ntwrk/midnight-js-types` dual-publishes CJS + ESM, so a consumer whose dependency graph resolves two copies of this module gets two distinct class objects and `instanceof` silently returns `false`. The class already sets `this.name`, which is the robust discriminant — exporting a guard next to it makes that discoverable:

   ```ts
   export const isProofServerCommunicationError = (e: unknown): e is ProofServerCommunicationError =>
     e instanceof Error && e.name === 'ProofServerCommunicationError';
   ```

2. **Unit test.** A small `packages/types/src/test/errors.test.ts` asserting `name`, the three properties, and the exact message string would (a) pin the message format as a backward-compat contract for anyone still string-matching, and (b) move `errors.ts` off its current 0% coverage.

If you extend the class at all, the highest-value addition would be an optional `readonly body?: string` — proof servers typically return a diagnostic text body on 4xx/5xx that is currently discarded. But that can be a follow-up.

packages/http-client-proof-provider/src/test/http-client-proving-provider.test.ts
```diff
@@ -224,7 +225,7 @@ describe('httpClientProvingProvider', () => {
       vi.mocked(ledger.createCheckPayload).mockReturnValue(new Uint8Array([20, 21, 22]));
 
       await expect(provider.check(serializedPreimage, 'test-circuit')).rejects.toThrow(
-        /Failed Proof Server response/
+        ProofServerCommunicationError
       );
```
sp-io, line 229: **Important:** this assertion swap actually *weakens* the test, and the PR's core value — `url`/`statusCode`/`statusText` being populated correctly — is never verified anywhere. `.rejects.toThrow(SomeClass)` only checks `instanceof`; it says nothing about the properties. Swapping the constructor's `statusCode`/`statusText` arguments at the throw site would still pass every test in this PR.

The mock already provides everything needed (`status: 500`, `statusText`, `url`), so asserting the payload is nearly free:

```ts
await expect(provider.check(serializedPreimage, 'test-circuit')).rejects.toThrow(
  ProofServerCommunicationError
);
await expect(provider.check(serializedPreimage, 'test-circuit')).rejects.toMatchObject({
  url: mockUrl,
  statusCode: 500,
  statusText: 'Internal Server Error'
});
```

(Keep both — `toMatchObject` alone doesn't check the class.) Same applies to the `prove` test below.

packages/http-client-proof-provider/src/test/http-client-proving-provider.test.ts
```diff
@@ -345,7 +346,7 @@ describe('httpClientProvingProvider', () => {
       vi.mocked(ledger.createProvingPayload).mockReturnValue(new Uint8Array([30, 31, 32]));
 
       await expect(provider.prove(serializedPreimage, 'test-circuit')).rejects.toThrow(
-        /Failed Proof Server response/
+        ProofServerCommunicationError
       );
```
sp-io, line 355: Same as the `check` test above: please also assert the error's properties here (this mock uses `status: 503`), e.g. `rejects.toMatchObject({ statusCode: 503, url: mockUrl })` alongside the class check — otherwise the wiring of the constructor arguments is untested.

sp-io (commented):
Thank you for this contribution, @efekrbas — this is a genuinely valuable DX improvement, and it's exactly the right direction for error handling in the framework. 🙏 A few things stood out as particularly well done:

- The error message is preserved byte-for-byte, so anyone currently string-matching keeps working while gaining a structured alternative. Thoughtful backward-compat call.
- The class follows the repo's error conventions precisely (`public readonly` constructor params + `this.name`) — it's actually more correct than the neighboring `InvalidProtocolSchemeError`, which forgets to set `name`.
- Placing it in `types/` makes it reachable without importing the provider package, and it flows through both barrels (`types` and the `midnight-js` meta-package) with no extra work needed.
- Narrow, focused diff: no dead code, no `any`, no swallowed errors.

The inline comments cover the code-level findings. A few items that don't anchor to diff lines:

**Before merge**
1. **Lint failure** — see the inline comment on the imports; `yarn lint:fix` resolves it (this currently blocks the CI lint gate).
2. **Unsigned commit** — commit `aa5fce0e` is unsigned; this repo requires GPG-signed commits, so branch protection will reject the merge. Please re-sign (e.g. `git commit --amend -S` + force-push).
3. **`packages/types/README.md`** — both places that enumerate the package's error exports (the `## Errors` import block and the `// Errors` group under `## Exports`) list all seven existing errors; please add the new one so the docs don't drift.

**Nice to have**
- `packages/http-client-proof-provider/README.md` ("Server Errors" section) still documents only the raw message string. Since this PR is explicitly a DX change, naming the error class there and noting it's imported from `@midnight-ntwrk/midnight-js-types` would help consumers discover it (the "Invalid Protocol" section above it already does this for `InvalidProtocolSchemeError`).
- One heads-up on scope of the benefit: the scoped-transaction API (`packages/contracts/src/internal/transaction.ts`) wraps provider errors in a generic `Error` with `{ cause }`, so consumers on that path will find this error at `error.cause` rather than catching it directly. Pre-existing behavior, not something this PR needs to fix, but worth knowing.
- Small terminology nit: the PR description refers to "SDK consumers" — midnight-js is referred to as a framework in this repo.

None of this diminishes the contribution — the blockers are mechanical fixes, and with the property assertions added to the two tests this is in great shape. Looking forward to the update!

efekrbas:
@sp-io Thanks again for the detailed feedback! I've just force-pushed an update that addresses all of your points:

GPG Signed: The commit is now properly GPG-signed.
Linting: Sorted the imports to pass the simple-import-sort/imports CI gate.
Renaming & JSDoc: Renamed the error to ProofServerResponseError, stated the network boundary explicitly in the JSDoc, and added @param tags.
Type Guard: Added and exported the isProofServerResponseError type guard.
Testing: Added errors.test.ts for property/guard assertions, and strengthened the provider tests to verify payloads using toMatchObject.
Docs: Updated the packages/types/README.md exports block and explicitly mentioned the class in packages/http-client-proof-provider/README.md.
Let me know if there's anything else needed before the merge!

sp-io:
Thanks for the quick turnaround, and for handling all of it -- signing, the import sort, the rename to `ProofServerResponseError`, and the type guard.

Sorry for leaving this sitting: a large feature under a tight deadline has been eating my review time. I'll get back to it as soon as I can. Thanks for your patience.

PR #1135, opened 2026-08-05
chore(config): pin npm to 12.0.2 and add supply-chain cooldown

gilescope:
Supply-chain hardening following the 2026-08-04 keyv/cacheable npm worm (868+ packages compromised via `preinstall` payloads published through legitimate maintainer accounts).

- Pin the npm CLI to 12.0.2 in CI (npm 12 blocks dependency lifecycle scripts by default behind `allowScripts`) instead of floating `npm@latest`/setup-node default
- Bump `node-version` where npm 12 requires it (engines: `^22.22.2 || ^24.15.0 || >=26`)
- Add `.npmrc` cooldown: `min-release-age=7` so freshly-published (potentially wormed) versions can't resolve for 7 days; `@midnight-ntwrk/*` and `@midnightntwrk/*` exempt so same-day internal releases still install
- Renovate: npm CLI pins in workflows/Earthfiles are matched by new custom managers in `midnightntwrk/renovate-config`, so this pin stays bumpable automatically


sp-io:
Pushed `6dffc7fb` onto this branch — the fix for the red check plus a bit of trimming. Kept the approach as it is: tooling stays on npm, npm gets hardened. (I explored routing the tooling through yarn.lock instead in #1219 and closed it — it meant adding `nyc` and its 27 direct dependencies to the root `devDependencies` to serve one CI step, which is not a good trade.)

**The red check.** `npm@12.0.2` declares `engines.node` as `^22.22.2 || ^24.15.0 || >=26.0.0`, and `.nvmrc` was `24.11.1`:

```
npm error code EBADENGINE
npm error notsup Required: {"node":"^22.22.2 || ^24.15.0 || >=26.0.0"}
npm error notsup Actual:   {"npm":"11.6.2","node":"v24.11.1"}
```

Bumped `.nvmrc` to `24.20.0` (current node 24 LTS "Krypton", bundles npm 11.19.0). This is the `node-version` bump the description already promised. Also refreshed the bundled-npm figure in `ci.yml`'s OIDC floor-check comment, which named `24.11.1` / `11.6.2` explicitly and would otherwise have gone stale on the same commit.

**A second failure that had not surfaced yet.** `e2e_tests_reports` in `ci-testkit-js-e2e.yml` has no `setup-build-env` and no `setup-node` of its own — it runs on whatever node the runner image ships. Its `Update NPM` step would have hit the same engine wall once the e2e matrix finished and the job actually ran. It now gets `setup-node` from `.nvmrc`.

**An unguarded pin is worse than no pin.** Every report step in that job is `if: always()`. `Update NPM` was not, so on any earlier failure — a `download-artifact` digest mismatch, a transient artifact-API error — the pin step gets skipped while `npx allure-commandline`, `npx ctrf` and `npm install -g junit-report-merger` below it still run, on the runner's own npm, with nothing in the log saying the pin was lost. That is the guarantee silently evaporating on exactly the path where you would want it.

So the pin now sits above the artifact download, both steps carry `if: always()`, and the step asserts `npm --version` afterwards so a pin that fails to take effect is loud instead of silent.

**Trimmed `.npmrc`.** Dropped the two `min-release-age-exclude[]` entries. They never applied to anything: Yarn 4 does not read `.npmrc` at all, so every `@midnight-ntwrk/*` dependency is installed outside this file's reach and the exemption was exempting nothing. Reworded the comment to state the real scope — the cooldown covers the `npx` and `npm install -g` calls in CI, which happen to be the only registry fetches in this repo that no lockfile governs. Worth being precise about, because the current wording reads as though it protects dependency installs, and it does not.

Happy to drop the `.npmrc` trim if you would rather keep the exemptions as future-proofing for an npm-installed internal package — it just is not doing anything today.

---

Separately, four pre-existing problems I ran into while checking this. None belong in this PR; flagging them so they are not lost:

1. `.github/actions/validate-packages/action.yml` ends both loops with `|| true`, so the step named "Validate Packages Before Publishing" cannot block a publish. Its second loop also never uses `$dir`, so `yarn pack --dry-run` packs the monorepo root once per package directory.
2. `commitlint.config.js` downgrades `type-enum`, `scope-enum`, `subject-empty`, `type-empty` and `scope-empty` to severity 1. Verified: `wibble(nonsense): whatever` passes the Commit Lint gate with exit 0.
3. `commit-lint.yml` installs `@commitlint/cli@18.6.1` globally while the husky hook uses the root devDependency, currently resolving to `20.5.0` — CI and local developers lint commits with different majors.
4. `ci-base.yml`'s coverage chain passes green when `find` matches zero `coverage-final.json` files: `nyc merge` writes `{}`, `nyc report` renders an empty HTML tree, and the artifact uploads look real.

PR #1151, opened 2026-08-14
fix(deps): align Effect dependencies on 3.22.1

mhounslow:
## Summary

- Raise the direct Effect dependency floor in `midnight-js-contracts` and `midnight-js-types` to 3.22.1.
- Resolve Effect to 3.22.1 across the monorepo.
- Regenerate the Yarn lockfile with the aligned version.

## Why

The dynamic cross-contract E2E stack loaded Effect 3.21.4 through Midnight JS and Effect 3.22.1 through Compact JS. When values crossed that package boundary, Effect repeatedly reported:

> Executing an Effect versioned 3.21.4 with a Runtime of version 3.22.1, you may want to dedupe the effect dependencies.

The Midnight JS package ranges allowed consumers to resolve 3.21.4 even though Compact JS uses 3.22.1. Raising the direct minimum and aligning the workspace resolution prevents that runtime version skew.

## Validation

- `corepack yarn why effect` — All workspace and transitive references resolve to 3.22.1.
- `corepack yarn build:core`
- `corepack yarn turbo run build --filter=@midnight-ntwrk/testkit-js`
- `corepack yarn lint`
- `corepack yarn test` — 27/27 tasks passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "@midnightntwrk/ledger-v9": "1.0.0-rc.3",
     "@midnightntwrk/onchain-runtime-v4": "4.0.0-rc.3",
     "@midnight-ntwrk/compact-runtime": "0.18.0-rc.1",
+    "effect": "3.22.1",
     "undici": ">=7.24.0",
     "tar": ">=7.5.19",
     "fast-xml-parser": ">=5.7.0",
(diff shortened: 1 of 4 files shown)


mhounslow:
Closing in favor of #1136, which already updates the same direct Effect dependencies. I added the dynamic-call version-skew context there, including the need to align with Compact JS on Effect 3.22.1.

PR #1155, opened 2026-08-17
feat(midnight-js): add protocol version identity module and HF gates

### LINKED ISSUE #1004: [HF v8→v9] MJS-01 — protocol package: unified v8/v9 dispatch APIs
**Workstream:** Midnight.JS — protocol package (midnight-js)
**Part of:** Ledger v8→v9 Hard Fork Migration — [SOW-Q3-10 / product#119](https://github.com/shieldedtech/product/issues/119) ([scope comment](https://github.com/shieldedtech/product/issues/119#issuecomment-4833224770))

## Summary
Add unifying APIs to the Midnight.JS `protocol` package that dispatch internally between Ledger v8 and v9.

## Scope
- Unified API surface in the `protocol` package abstracting v8 vs v9
- Internal dispatch to the correct version implementation
- Versioned types/serialization as needed to support both

## Acceptance criteria
- `protocol` package exposes unified APIs that work for both v8 and v9
- Internal dispatch selects the correct version
- Covered by tests against both versions

## Dependencies
- Foundational for MJS-02 (contracts orchestration) and MJS-03 (provider API updates)

---
_First-pass breakdown derived from the SOW-Q3-10 scope comment. Sizing and implementation plan to be reviewed and confirmed by the team._

sp-io:
## Summary

- **`protocol` — version identity.** `LedgerVersion` (`'v8' | 'v9'`) and `protocolVersionToLedger`, mapping a raw indexer/node `protocolVersion` integer onto a ledger runtime by node major: `1_000_000–1_999_999 → v8`, `2_000_000–2_999_999 → v9`. Anything else throws rather than guessing. Sourcing helpers `versionOfRecord` / `networkHeadVersion` tag the error with the call path automatically. A compile-time guard makes adding a `LedgerVersion` without a mapping entry a type error. Reachable from the root barrel and from a lightweight `./version` subpath that does not load the ledger/compact-js/onchain-runtime/platform namespaces the barrel re-exports.
- **`protocol` — typed errors.** `UnknownProtocolVersionError` with a stable `code` and a `reason` separating `'malformed'` (not a non-negative integer — your data source is wrong) from `'unknown'` (a well-formed version this build does not map — upgrade midnight-js). Reachable from a new lightweight `./errors` subpath, so `utils` can read error codes without pulling in the ledger/compact-js/onchain-runtime/platform namespaces.
- **Build — one rollup pass per package.** The shared factory `main` gained in #1180 runs one independent rollup pass per entry, which inlines any module reached by two entries into both. Two copies of a class are two classes, so `instanceof` silently answers `false` across two subpaths of the same package — exactly where a caller tells failure modes apart. The factory now emits the JavaScript for every entry in a **single** pass, so the error module is emitted once and imported by relative path. Entries still come from the `exports` map, so no package config changed. Declarations stay one pass per entry (see below). `dist-error-identity.test.ts` gates this against the built artifact, taking its subjects from the `exports` map so a subpath added later is covered without editing the test.
- **`utils` — shared primitives.** `assertNever` (no `JSON.stringify`, so BigInt/circular/sensitive values cannot leak into a message) and a central error-code registry `MIDNIGHT_JS_ERROR_CODES` with `hasErrorCode` guards.
- **Gates.** ESLint bans runtime imports of `protocol/v8` outside `packages/protocol/src/` — type-only imports allowed, dynamic `import()` covered in both its statically matchable forms. The `packages/protocol/src/` exemption is expressed with `ignores` on the block that adds those selectors, never by disabling the rule, so an unrelated selector another block adds to `no-restricted-syntax` cannot be dropped along with it. Canaries cover one case per config artifact, plus the shape invariant itself.

Part of the Ledger v8→v9 hard-fork migration (#1004, MJS-01). Architecture doc: `docs/superpowers/architecture/2026-08-05-ledger-v8-v9-dual-support-architecture.md` on branch `docs/superpowers-specs` (not yet on `main`).

## Scope decisions taken during review

- **Node 0.x is not mapped.** midnight-js will meet only node 1.x or 2.x by the time it reaches the environment, so the major-0 exemption and its table were removed. `22_000–22_999` now throws `'unknown'` instead of returning `v8`. The indexer's own table still maps node 0.22; that mapping stays authoritative for the indexer and is simply out of scope here.
- **`Symbol.hasInstance` was tried and dropped.** It made `instanceof` code-based to survive duplicated module copies, but it narrowed to a type promising `protocolVersion`/`path`/`reason` while checking only `code` — so an `Error` merely carrying the code compiled clean and threw `TypeError` on field access, and a two-branch `reason` test then defaulted to the "upgrade midnight-js" advice for what was actually malformed input. It also inherited into any subclass, threw out of the guard itself on a throwing `code` getter, and answered `false` cross-realm and after `structuredClone`. With one copy guaranteed by the build, plain prototype identity is both correct and honest.
- **`parseSerializedTag` was removed.** Its `namespace:version:` grammar does not match real tags: `midnight:verifier-key[v6]:` fails the charset, and `midnight:transaction[v4](signature[v1],proof-preimage,embedded-fr[v1]):` fails the 64-byte scan bound before the charset is reached. Real tags are `namespace:type-descriptor:`, and the descriptor's `vN` is the version of that object's serialized shape, **not** the ledger era — serializing an unproven transaction against the pinned `ledger-v8@8.1.1` emits `transaction[v9]`. A field named `version` next to `LedgerVersion` in a PR about telling v8 from v9 apart invites exactly the wrong inference. Nothing in the MJS-01 stack consumed it. It returns with its first real caller, against the authoritative grammar and real fixtures.

## Test plan

- [x] Tests written first (TDD), verified failing before implementation
- [x] All existing tests pass (no regressions) — repo `test:unit` 20/20 packages
- [x] Lint clean, `typecheck:tests` clean, forced build 16/16
- [x] `protocol` 105 tests at 100% statements/branches; `utils` above thresholds
- [x] Gate tests verified by sabotage: restoring `main`'s per-entry factory turns four `dist-error-identity` tests red; reintroducing `'no-restricted-syntax': 'off'` turns the config canary red
- [x] Both loaders exercised against the built package on Node 22.13: `import` and `require(esm)` each resolve `.`, `./errors` and `./version`, and the error thrown from `./version` passes `instanceof` against the class from `./errors`
- [x] README claims checked against the built artifacts — every documented export resolves, and the documented mapping table matches runtime behaviour at every boundary

## Merge flow (MJS-01 integration branch)

This PR targets the long-living branch `feat/1004-protocol-dual-ledger`; every MJS-01 PR merges there first, and a single final PR merges that branch to `main` and closes #1004. Do not squash-merge this into `main` directly.

This branch has been merged up to current `main` (through the integration branch, which was a pure fast-forward — it carried no commits of its own).

## Synced onto ESM-only `main` (#1180)

#1180 dropped CJS publishing and rewrote the build. Four things had to change here, three of which would otherwise have merged cleanly and wrongly.

### The single-pass build moved into the shared factory

#1180's `createRollupConfig` derives entries from the `exports` map — which is strictly better than the hand-written entry lists this PR used to carry — but it still runs a pass per entry, so it reintroduces the inlining this PR exists to remove. The single pass now lives in that factory instead of beside it, and `build-tools/rollup.config.multi-entry.mjs` is gone along with both packages' bespoke configs. `packages/{protocol,midnight-js}/rollup.config.mjs` are back to `main`'s four-line factory call.

**Declarations deliberately stay one pass per entry.** Bundling them together too makes `rollup-plugin-dts` append the *other* entries' external imports to an unrelated entry's declaration — `ledger.d.ts` picked up eight side-effect imports of `compact-js`, `platform-js` and friends. Types are structural, so a declaration repeated in two entries costs nothing at the type level; only runtime classes have identity.

Measured rather than argued: a whole-repo forced build compared file by file against the same build with `main`'s per-entry factory is **96 of 98 artifacts byte-identical**. The two that differ are `protocol/dist/index.js` and its sourcemap, which now import the error classes from `./errors.js` instead of declaring their own copy. No package grew a `dist/shared/` chunk.

### `./version` became a subpath

#1180 added an export-surface gate requiring one `exports` subpath per top-level module under `src/`, and `src/version.ts` had none. The subpath earns its keep rather than merely satisfying the gate: the root barrel re-exports the ledger, compact-js, platform-js and onchain-runtime namespaces, so reading a protocol version through it loads all four. `./version` is the same lightweight door `./errors` already was.

### Two silent merges

- `packages/protocol/package.json` merged without a conflict but left `./errors` in the old `{ import, require }` shape among ten `{ types, default }` siblings. Fixed.
- `dist-error-identity.test.ts` reads the `exports` map to find its subjects. Its reader only recognised `{ import, require }`, so post-merge it would have found **zero** subpaths and passed while asserting nothing. It now reads the `default` condition.

### The combined `no-restricted-syntax` selectors

The merge-time action this description used to carry, executed. #1161's unsafe-cast gate and this PR's v8 gate both configure `no-restricted-syntax`, with *different* exemptions — casts are off in test files, the v8 ban is off in `packages/protocol/src/`. Flat config replaces a rule's options wholesale, so a file matched by both blocks keeps only the later one's selectors.

Both selector lists are now named constants applied through four **mutually exclusive** blocks, each spelling out every list that applies to its scope, so block order cannot decide the outcome:

| Scope | Casts | v8 |
|---|---|---|
| package sources | yes | yes |
| `packages/protocol/src/` | yes | exempt |
| `packages/*/src/test/` | exempt | yes |
| `testkit-js/` | exempt | yes |

Four new canaries assert exactly that table, including the overlap case (one file, one cast and one dynamic v8 import, two errors expected). A shadowing mistake fails the suite instead of silently disabling a gate.

## Follow-up for the rest of the MJS-01 stack

#1156, #1159 and #1164 sit on top of this branch and each need the same parent-to-child merge before they are readable against current `main`.


VanessaPC:
One thing on the caching gates, outside the diff but directly related to it.

`turbo.json` now lists `build-tools/**` and `eslint.config.mjs` under `globalDependencies` so a change there can't replay a cached PASS. The CI paths-filter one layer up still has that same hole.

In `.github/workflows/ci.yml`, the `shared` filter covers only `.github/workflows/**`, `yarn.lock` and root `package.json`, while `midnight-js` and `testkit-js` fire on `packages/**` / `testkit-js/**`. `build-tools/**`, `eslint.config.mjs` and `turbo.json` are in none of them.

So a later PR that only reverts `createRollupConfig` to a pass-per-entry (bringing back the duplicated error class), or only sets `'no-restricted-syntax': 'off'`, touches zero files under `packages/**`. All three filter outputs come back false, `ci-midnightjs` and `ci-testkit` get skipped by `if: inputs.should_run == 'true'`, and it merges green with `dist-error-identity.test.ts` and the lint canaries never having run — which are the two gates this PR adds.

This PR isn't affected itself, since it also touches `packages/**`. Adding `build-tools/**`, `eslint.config.mjs` and `turbo.json` to the `shared` filter would close it. Fine as a follow-up if you'd rather not widen the scope here.

sp-io:
Confirmed, on `main` and on this branch's head. The escape route is real: `ci-midnight-js.yml:33` gates its only real job on `if: inputs.should_run == 'true'`, so with all three filter outputs false the whole build/lint/test job is skipped, and a job skipped by `if` satisfies branch protection. A PR touching only `build-tools/**` or `eslint.config.mjs` merges green with `dist-error-identity.test.ts` and the lint canaries never having run.

Not widening the filters in this PR. Two reasons: the hole predates it -- this PR closes the turbo-hash layer, paths-filter is the layer above -- and #1155 sits mid-stack under `feat/1004-protocol-dual-ledger`, where the stack syncs parent to child by merge, so a shared `ci.yml` edit here propagates through the whole chain. Tracked as #1209.

Two paths to add to your list, same failure mode:

- `.github/actions/**` and `.github/scripts/**`. `shared` covers only `.github/workflows/**`, while `ci-base.yml:36` and `ci-midnight-js.yml:50` both run `./.github/actions/setup-build-env`. A PR editing only the composite action that sets up the toolchain skips all of CI.
- `tsconfig.base.json`. Missing from the filters *and* from the `globalDependencies` this PR adds. Compiler-option changes move no task hash (`build.inputs` is `src/**/*.ts`) and trigger no job, so both layers are open. #1185 section 3 lists it on the turbo side; #1209 carries it on the CI side.

One trade-off named in #1209 rather than left implicit: putting `turbo.json` and `build-tools/**` into `shared` means every change to them triggers a full run including testkit e2e. That is the correct cost for inputs that decide every artifact, but it is a real cost per PR.

@VanessaPC are you ok to approve ?

VanessaPC (approved):

packages/protocol/src/test/dist-error-identity.test.ts
```diff
@@ -0,0 +1,135 @@
...
+};
+
+const isErrorClass = (value: unknown): boolean =>
+  typeof value === 'function' && Object.prototype.isPrototypeOf.call(Error, value);
+
+/** Every exported error class of a module namespace, keyed by export name. */
+const errorClassesOf = (namespace: object): Map<string, unknown> =>
+  new Map(Object.entries(namespace).filter(([, value]) => isErrorClass(value)));
+
+describe('dist error-class identity gate', () => {
```
sp-io, line 93: Same fix as the one asked for on #1156 (`dist-laziness.test.ts`), applied here too: this gate no longer skips when a bundle is missing.

A vitest skip is reported as a pass, so `skipIf(!bundlesExist && !process.env.CI)` meant the gate silently stopped guarding outside CI. Requiring a build is fine — the orchestration already guarantees one: the turbo `test` task has `dependsOn: ["build", "^build"]`, and CI runs `yarn build` before `yarn test`. So a missing bundle is a real failure everywhere, and `contentOf()` now throws a message naming the build command.

Verified on this branch: without `dist/` → 15 failed, 0 skipped; after `yarn build` → 105 passed, 0 skipped, coverage 100%. Lint and `tsc --noEmit` clean.

PR #1156, opened 2026-08-17
feat(midnight-js): add protocol/v8 subpath and lazy load V8 accessor

### LINKED ISSUE #1004: [HF v8→v9] MJS-01 — protocol package: unified v8/v9 dispatch APIs

sp-io:
## Summary
- `protocol` gains the `./v8` subpath re-exporting `@midnightntwrk/ledger-v8@8.1.1` (exact pin; the only place in the tree resolving it — AC6 slice)
- Lazy, memoised `loadLedger8()` accessor in `src/lib/load-v8.ts`, reaching the subpath through a relative `import('../v8.js')` (ADR 0004) — the v8 WASM never loads in a v9-only session (NFR6); dist-level CI-testable gate asserts `dist/index.js` carries zero static `ledger-v8` linkage and links the v8 chunk only through that dynamic import (verified on a cold build)
- `./v8` in the exports map is what makes `src/v8.ts` a build entry under the ESM-only single-pass build; protocol export surface strictly additive (D3), asserted by the sorted-key ACL test
- dist-laziness test skips visibly when `dist/` is absent instead of failing with ENOENT; the v8 surface and failure-path suites resolve the module from source and need no prior build

**Merge flow (MJS-01 integration branch):** stacked on #1155 (base: `feat/1004-version-module-gates`). After #1155 merges into the long-living `feat/1004-protocol-dual-ledger`, retarget this PR to `feat/1004-protocol-dual-ledger` and merge it there. Every MJS-01 PR merges to that branch first; a single final PR merges it to `main` and closes #1004. Part of the Ledger v8→v9 HF migration (#1004, MJS-01 plan PR 1004-C).

Note for reviewers (scope check updated 2026-08-17): **both** scopes publish `ledger-v8` on the public registry — `@midnight-ntwrk/ledger-v8` (8.0.2…8.1.1, the historical scope the HF spike used) and `@midnightntwrk/ledger-v8` (8.1.1, 8.2.0-rc.1, matching the `@midnightntwrk/ledger-v9` scope). This PR pins the latter at 8.1.1. The OQ2 credentialed org-ownership check for both scopes remains a pre-merge action (see spec OQ2) — with dual-published scopes it matters more, not less.

Synced with #1155 on 2026-08-20 (merge, not rebase). The parent brought the ESM-only build (#1180), so this PR now follows it: `./v8` maps to `dist/v8.js` / `dist/v8.d.ts`, the loader uses the relative import instead of the package self-reference, and `load-v8.ts` sits in `src/lib/` — the parent's new `export-surface` gate pairs every module directly under `src/` with an exports subpath, and the accessor ships through the root barrel rather than a subpath of its own (`lib/` = not a build entry, distinct from `contracts/src/internal`, which marks implementation hidden from consumers).

## Test plan
- [x] Tests written first (TDD), verified failing before implementation
- [x] All existing tests pass (protocol 118/118, coverage 100/100/100/100; full-repo `yarn test:unit` 28/28 tasks green on this head)
- [x] Lint clean, `yarn typecheck:tests` clean, cold `yarn build` exit 0 with the dist gate running (not skipping) and green


sp-io:
Review fixes pushed (f00027f7, 4c48a992):

- **loadV8 failure contract**: failed import now rejects with `Ledger8RuntimeMissingError` (`MIDNIGHT_JS_P_LEDGER8_RUNTIME_MISSING`, original error as `cause`) and the memo is cleared so the next call retries; negative-path test added.
- **Laziness gate hardened**: dist test now also forbids static linkage of the `protocol/v8` subpath (the realistic regression the old `ledger-v8` regex missed), positively asserts the dynamic import survives the build, and checks the `./v8` exports-map artifacts exist.
- **Public publish fix**: `publish-public-npm.mjs` now rewrites `@midnight-ntwrk` → `@midnightntwrk` inside built JS/DTS files too — the compiled self-reference specifier baked into `dist/` otherwise fails to resolve in the `@midnightntwrk` publish (also fixes cross-package specifiers for every published package).
- `typeof import()` alias replaced with `import type` (drops the eslint-disable and its incorrect justification); comments no longer reference spec sections/OQ items not present on main.
- `loadV8` tests are dist-gated with visible skips (same policy as the dist gate); floating promises fixed.
- README documents `./v8`, `loadV8()`, and the type-only import escape hatch.

ADR for the new runtime dependency: #1157 (separate PR per review workflow).

Local verification: 90/90 tests pass, coverage 100%, lint and `typecheck:tests` clean.

VanessaPC (approved):

VanessaPC (approved):

packages/protocol/src/test/dist-laziness.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+// Skipped (not omitted) when dist/ is absent, so a run without a prior build
+// still reports these as visible skips rather than silently vanishing —
+// run `yarn build && yarn test` to green them.
+//
+// The file reads happen lazily inside each `it` body (not in the `describe`
+// body) so that `describe.skipIf` actually prevents them from running when
+// dist/ is absent — vitest still executes a `describe` callback during test
+// collection even when `skipIf` is true; only the nested `it`s are skipped.
+describe.skipIf(!distIndexExists)('dist laziness gate', () => {
```
kapke: Are skips reported as errors in CI? IF not - it should be orchestrated in a way, which ensures the dist exists

packages/protocol/package.json
```diff
@@ -74,6 +78,7 @@
     "@midnight-ntwrk/compact-js": "2.5.5-rc.7",
     "@midnight-ntwrk/compact-runtime": "0.18.0-rc.1",
     "@midnight-ntwrk/platform-js": "3.0.0",
+    "@midnightntwrk/ledger-v8": "8.1.1",
```
kapke, line 81: why using exact versions instead of `^`-prefixed or `~`-prefixed?

sp-io: Because we had in the past multiple issues with incompatible ledger between patch versions sadly...

packages/protocol/src/test/dist-laziness.test.ts
sp-io: Right — vitest skips are reported as passes, so the gate was silently self-disabling. Fixed in 878c4afd: the skip is gone.

Orchestration already guarantees `dist/`: the turbo `test` task has `dependsOn: ["build", "^build"]`, and CI runs `yarn build` before `yarn test` (ci-base.yml). So a missing `dist/` is a real failure, and the suite now treats it as one — a bare `vitest run` without a prior build fails with:

> dist/index.js is missing: this suite inspects the rollup output. Run "yarn build" first, or run "yarn turbo test", which builds before testing.

Verified: without `dist/` → 5 failed, 0 skipped; after `yarn build` → 118 passed, 0 skipped, coverage 100%.

sp-io: Follow-up: the sibling gate had the same shape (`dist-error-identity.test.ts`, `skipIf(!bundlesExist && !process.env.CI)`), so it is fixed the same way in #1155 (ec5f6970) and merged down here (15bf019f). No `skipIf` is left in `packages/protocol/src/test/`.

On this branch after the merge: 118 passed, 0 skipped, coverage 100%; without `dist/` the two dist gates fail loudly instead of skipping.

packages/protocol/package.json
kapke: I see. I believe these packages are going to behave much better in this regard now as there is the mainnet out there and the rust packaging adopted separate-crate-per-era model too, so there's much less of an incentive for them to not follow semver now.

sp-io: For now I prefer a defensive approach "just in case" :)
Once the HF feature will be added, I will proceed as suggested.

sp-io: Tracked in #1217 — switch these to caret (`^`) ranges once the hard-fork stack lands and the deps are GA. Short version of the reasoning there: caret rather than tilde (for the 0.x deps the two are identical anyway, and inside one ledger era the minor line is additive by construction), prereleases stay exact because a range on an `-rc` accepts `rc.4` without ever reaching the next minor, and `zkir-v2` stays exact because the engine and the `WasmProver` params have to come from the same zkir release.

PR #1164, opened 2026-08-18
feat(midnight-js): add ledger-8 engine core with envelope, down-convert and instance guards

sp-io:
# Overview

**PR 1004-D of the MJS-01 series (#1004)** - adds the ledger-8 engine core to `@midnight-ntwrk/midnight-js-protocol`. Stacked on #1159 (HF fixtures), which stacks on #1156 <- #1155. Merge order: #1155 -> #1156 -> #1159 -> this.

**Now also carries #1186** (golden-fixture assertions), merged into this branch rather than landing as a separate PR.

## Summary

- **`src/lib/engine/envelope.ts`** - version-aware `extractEncodedStateValue(raw, version, ledger8ContractState)`: decodes v8-era envelopes (pre-fork `contract-state[v6]`) and post-fork v9 envelopes into the shared `EncodedStateValue` POJO. `version` is validated against the decoder table before dispatch; an unknown value throws the coded `UnknownLedgerVersionError` rather than resolving through the prototype chain. Any decode failure is a typed `DownConvertFailedError` (`MIDNIGHT_JS_P_DOWN_CONVERT_FAILED`), never a silently empty state.
- **`src/lib/engine/down-convert.ts`** - `downConvertForExecution(state, ledger8Runtime)`: v9->v8 state bridge over the WASM packages' own `encode()`/`decode()` POJO round-trip. Structural integrity is checked by re-encoding and deep-comparing against the source. `assertMerkleTreesRehashed` walks the state and asserts every bounded Merkle tree has a readable root via `checkRoot`, throwing `MerkleNotRehashedError` (`MIDNIGHT_JS_P_MERKLE_NOT_REHASHED`); an unrecognised `StateValue` variant is a hard failure, not a silent skip.
- **`src/lib/engine/instance-guard.ts`** - dual-WASM fail-fast: `assertSharedLedger8Instance(axis, probeA, probeB)` compares two **class bindings** obtained from two different acquisition paths (a namespace object is not a valid probe - a re-export produces a fresh one over a single physical copy). Rejects nullish probes; throws `Ledger8InstanceMismatchError`. The single genuine axis is `onchain-runtime-v3`.
- **`src/errors.ts`** - four new error classes on the registry: `Ledger8InstanceMismatchError`, `DownConvertFailedError`, `MerkleNotRehashedError`, `UnknownLedgerVersionError`. Each error's `code` field is typed as its own literal, so `instanceof` and `code` cannot drift and consumers can `switch` exhaustively.
- **Golden fixtures (from #1186)** - `src/test/engine-golden-fixtures.test.ts` asserts the bridge against the committed hard-fork goldens: a real migrated on-chain state down-converts byte-identically to its pre-migration v8 form, a golden Merkle state comes out with a readable root, and every tampered golden fails closed without leaking bytes. `packages/protocol/turbo.json` declares the fixture directory as a test input so editing a golden invalidates the test cache.
- **Coverage** - per-glob 100/100/100/100 thresholds for `src/version.ts` and all engine modules; package totals stay 100 and always-on.
- **Deps** - `@midnight-ntwrk/onchain-runtime-v3@3.1.0` as a protocol dependency, now also pinned in the root `resolutions` alongside every other vendor runtime. The devDep npm alias `onchain-runtime-v3-alt` gives the dual-instance negatives a real second physical WASM copy.

Engine modules are intentionally **not** exported from the root entry or `./v8` - the root stays v9-static; export surfacing lands with the execution legs (PR 1004-E) as a single lazy facade, `loadLedger8Engine()`.

## Review fixes applied on this branch

A multi-agent review of the original diff surfaced these; all are fixed here.

1. **Published error message named one npm scope twice.** The dual-publish (`.github/scripts/publish-public-npm.mjs`) rewrites the old scope to the new one inside built `.js`/`.d.ts` files, not only in `package.json`. `AXIS_PACKAGE_NAMES` held both scoped names as literals, so the public-npm artifact collapsed them into one - and no test could see it, since the suite runs against unrewritten source. Scopes are now held apart from the `/` and joined at use; a test asserts the module carries no scoped literal at all.
2. **`ENVELOPE_DECODERS` was fail-open.** A plain object literal indexed by `version` resolved unexpected keys through `Object.prototype`: `'constructor'` handed the caller's own raw bytes back as an `EncodedStateValue`, `'toString'` returned `'[object Object]'` - neither throwing, both reachable from untyped JS. The table is now null-prototype and frozen, and `version` is validated before dispatch. Validating first also keeps `stage` inside its closed union.
3. **The Merkle walk silently skipped unknown variants.** The `switch` had no `default` and returns `void`, so a variant outside the five it names fell through having checked nothing and recursed into nothing. The compile-time exhaustiveness guard is now folded into the `switch` as a `never` default that also throws at runtime - the pinned `.d.ts` cannot vouch for a caller-injected runtime's WASM.
4. **`checkRoot` mishandled a fallible vendor binding.** The wasm-bindgen shim for `root()` rethrows a Rust `Err`; that throw escaped uncoded and was relabelled `DOWN_CONVERT_FAILED`, pointing the caller at its envelope bytes when the fix is to rehash. A throw and a `null` are now both reported as `MERKLE_NOT_REHASHED`, carrying the cause.
5. **`structurallyEqual` compared key counts, not key sets** - two objects sharing no key compared equal whenever the diverging key's value was `undefined`. Unreachable in today's algebra; guarded so it stays that way.
6. **Instance-guard contract narrowed to a class binding.** A re-export - the case the error message names first - produces a fresh namespace object over one physical copy, so a namespace comparison would report a mismatch on a healthy install.
7. **Tests were single-codec where production is cross-codec.** Every container test built both sides of the re-encode comparison with onchain-runtime-v3; the only genuinely cross-codec case used a bare cell. Added v9-native multi-entry-map and nested-array cases, the v8 decoder's truncated/trailing-bytes paths, the v9 extraction stage, and an explicit pin on Map order-sensitivity. #1186's goldens close the rest.
8. **Two comments claimed more than they delivered** - the `dist-laziness` onchain-runtime-v3 case passes today because no build entry reaches the engine at all, and the vendor drift detectors are checked by the pre-push hook only, never by CI. Both now say so.

## Second review round (VanessaPC), fixes applied

Five of the eight inline comments are fixed in `fc8f22d`; 20 tests red first, all green after. Package stays 239/239, coverage 100/100/100/100, lint clean.

1. **`downConvertForExecution` validated nothing about `ledger8Runtime`.** A runtime missing a binding this function dereferences raised a bare `TypeError` that the `catch` relabelled `DOWN_CONVERT_FAILED` - an error whose message sends the caller to audit input bytes that are fine. It now throws `LEDGER8_RUNTIME_INVALID` naming the absent member, matching the check the envelope seam already had. Only the two members it actually calls are checked; `ContractState` is on the interface to pin the era and is never dereferenced here.
2. **`Ledger8RuntimeInvalidError`'s remediation pointed at the wrong package.** It told callers to pass the `ContractState` that `loadLedger8()` exposes. `loadLedger8()` resolves `@midnightntwrk/ledger-v8`, which (verified against the published 8.1.1 tarball: zero dependencies, `ledger-v8.d.ts:765`) exports its *own* `ContractState` with a matching `static deserialize` on a separate WASM instance. Following the message therefore yielded a class that passes the duck-typed guard and decodes on the wrong copy. The message now names onchain-runtime-v3 and says plainly that this package exposes no accessor for it.
3. **`UnknownLedgerVersionError` shipped undocumented.** Its doc block sat immediately before a second doc block, so it was discarded and the class emitted with no typings comment. Moved down to the class it documents; confirmed in `dist/errors.d.ts`.
4. **A nullish instance probe was diagnosed as a dual-instantiation.** A caller who omitted a probe was told two physical copies exist and sent to `npm why` after a duplicate that is not there. Nullish probes now throw `LEDGER8_RUNTIME_INVALID` - the same code the envelope seam uses for a binding handed over incomplete - and the mismatch error is reserved for an actual reference-equality failure.
5. **The axis package-name table was fail-open.** It kept its prototype and `axis` was never validated, so a non-axis string rendered an `Object.prototype` member as the npm package to trace (`@scope/function Object() { [native code] }`). The table is now null-prototype and frozen, and `axis` is validated against a closed, `satisfies`-checked set in `instance-guard.ts` before any probe is compared, throwing the new `UnknownLedger8AxisError` (`MIDNIGHT_JS_P_UNKNOWN_LEDGER8_AXIS`). The axis set lives in the engine module rather than in `errors.ts`, which is a build entry - nothing outside needs to test an axis, and `satisfies` fails the build if the union grows.

Not applied:

- **`vitest.config.ts:52`** - the comment is correct as written. The claim that a glob-matched file leaves the global aggregate does not hold for the pinned Vitest 4.1.7: `resolveThresholds` adds *every* file to the global map, under its own comment `// Global threshold is for all files, even if they are included by glob patterns`. That was the behaviour in older majors, not this one, so the per-glob entries are an extra gate rather than a replacement - which is what the comment says.
- **`down-convert.ts:212/215/219`** (the `!` assertions) - the consistency argument holds, but each new guard needs a fake-runtime test to keep the per-file 100% branch floor, so it is a follow-up rather than part of this round.
- **`envelope.ts:115`** (v9 requiring the pre-fork runtime) - kept as it is, docblock corrected in `64a5068`. The comment defended the unconditional requirement by the cost to a v9-only consumer; there is no such consumer. `extractEncodedStateValue` has one call site in the whole stack - `createLedger8Engine()` in #1168/#1165 - and it has already awaited onchain-runtime-v3 two lines earlier to assemble the runtime for `downConvertForExecution`. The v9 decoder is there so the bridge can read a post-fork envelope before down-converting it; a caller with no pre-fork runtime reads ledger-v9 directly and never reaches this function. Note also that overloads (or a discriminated parameter object) would not compile against that call site, which passes `version` as a `LedgerVersion` union value rather than a literal. The docblock now gives the real reason and records the condition that flips it: if this seam is ever surfaced beyond the engine, make the parameter optional and move the check into the `'v8'` decoder.

**Downstream:** `assertSharedLedger8Instance` and `downConvertForExecution` change only which coded error a misuse produces, and no signature moved, so #1168 and #1165 need a merge-forward but no code change.

## Dependent PRs: already synced

The stack has been merged forward, so #1168 and #1165 already carry these fixes.

- **#1168 (`feat/1004-engine-execution`)** - 270/270, lint clean, coverage 100/100/100/100. Two conflicts needed a decision rather than a union. Both branches had hardened the same prototype-chain hole in `extractEncodedStateValue`; this branch's resolution won, because a distinct `UNKNOWN_LEDGER_VERSION` code lets a consumer tell a bad version from bad bytes, and the offending string stays in a field instead of being interpolated into the message. The `'envelope extraction'` member of `DownConvertStage` is unreachable under that resolution and was dropped. #1168's prototype-member test was superseded by the wider one here, so the duplicate went and its two unique contributions were folded into the survivor: the `'v7'` case, and the note that `Ledger8Engine.extractState` is the reachable path.
- **#1165 (`feat/1004-engine-legs`)** - 301/301, lint clean, coverage 100/100/100/100, no test changes needed. Its `dist-laziness` version was kept wholesale: the comment here says no build entry reaches `lib/engine`, which is true on this branch and false there, since #1165 has the `./engine` entry.

Both call `extractEncodedStateValue(raw, version, ocrt3.ContractState)` and `assertSharedLedger8Instance('onchain-runtime-v3', ocrt3.ChargedState, glue.ChargedState)` - class bindings, matching the narrowed contract - so no signature changes were needed downstream.

Also worth knowing for 1004-E: `onchain-runtime-v3` stays a runtime `dependency` (not a devDependency), because #1168 imports it dynamically in production.

## Test plan

- [x] Tests written first (TDD), verified they fail before implementation (17 red before the fixes, all green after)
- [x] All existing tests pass: protocol 212/212, coverage 100/100/100/100
- [x] Dependent PR #1168 verified by trial merge (206/210, deltas enumerated above)

## Submission Checklist

- [x] Self-reviewed the diff
- [ ] Reviewer requested
- [x] Update README.md file (if relevant) - n/a
- [x] Update documentation (if relevant) - n/a
- [x] No new todos introduced

## Links

- Part of #1004 (MJS-01). Depends on #1155, #1156, #1159. Absorbs #1186.


packages/protocol/src/lib/engine/down-convert.ts
```diff
@@ -0,0 +1,298 @@
...
+ * What it does not cover: anything the pre-fork encoder re-derives rather than
+ * carries. A bounded Merkle tree encodes as its height and leaves, never its
+ * node hashes, so a tree whose internal hashes were re-materialized
+ * differently across the fork boundary still re-encodes byte-identically here.
+ * {@link assertMerkleTreesRehashed} establishes only that each tree has a
+ * readable root, not that the root matches the source's. Comparing roots
+ * across the boundary needs the source-side tree, which this function never
+ * sees — it belongs at the envelope seam.
+ */
+export const downConvertForExecution = (
```
VanessaPC, line 285: **medium — unguarded runtime seam**

This validates nothing about `ledger8Runtime`, while the sibling seam validates precisely so that a runtime fault isn't reported as a data fault:

```ts
// envelope.ts:115
if (typeof ledger8ContractState?.deserialize !== 'function') {
  throw new Ledger8RuntimeInvalidError('ContractState.deserialize');
}
```

A runtime handed over without `StateValue` (the hand-assembled case `Ledger8RuntimeInvalidError`'s own message calls out) produces `TypeError: Cannot read properties of undefined (reading 'decode')` at line 281, which the `catch` at line 292 re-throws as `DownConvertFailedError('state down-convert', cause)`. That error's message then tells the caller to read the cause because it "distinguishes a tag mismatch ... from truncated, trailing, or empty input bytes" — sending them to audit input bytes that are fine. That's exactly the mis-diagnosis the new error code was added to prevent, at the one seam left unguarded.

A `typeof ledger8Runtime?.StateValue?.decode !== 'function'` / `ChargedState` check before the `try` would close it.

packages/protocol/src/errors.ts
```diff
@@ -77,14 +81,221 @@ export class UnknownProtocolVersionError extends Error {
...
+ * serves. `missingMember` names the absent binding; it is one of this module's
+ * own literals rather than caller-supplied text, so exposing it leaks nothing.
+ */
+export class Ledger8RuntimeInvalidError extends Error {
+  readonly code = PROTOCOL_ERROR_CODES.LEDGER8_RUNTIME_INVALID;
+
+  constructor(readonly missingMember: string) {
+    super(
+      'The ledger-8 runtime handed to the down-convert engine cannot be used: the binding it needs is missing. ' +
+        'Acquire the runtime with loadLedger8() and pass the pre-fork ContractState class it exposes instead of ' +
```
VanessaPC: **medium — this remediation points callers into a dual-instantiation, and the guard won't catch it**

`Ledger8ContractState.deserialize` (`envelope.ts:43`) is typed to return `@midnight-ntwrk/onchain-runtime-v3`'s `ContractState`, and every real call site passes `ocrt3.ContractState`. But `loadLedger8()` (`src/lib/load-v8.ts`) dynamically imports `src/v8.ts` → `@midnightntwrk/ledger-v8`. From the published 8.1.1 tarball:

```
package/ledger-v8.d.ts:765  export class ContractState {
                            ...
                              static deserialize(raw: Uint8Array): ContractState;
package/package.json        dependencies: none, peerDependencies: none
```

So `ledger-v8` exports its *own* `ContractState`, and with zero dependencies it cannot be re-exporting onchain-runtime-v3's. A caller who follows this message hands over a class that **satisfies the `typeof deserialize === 'function'` guard above** and then decodes envelopes on a second, unrelated WASM instance — the exact dual-instantiation `instance-guard.ts` exists to detect. So this isn't only a misleading string: it routes past the check meant to catch the failure it causes.

Related: the package exposes no sanctioned lazy accessor for onchain-runtime-v3 at all (which is why it had to become an unconditional runtime `dependency`). Worth saying that plainly here instead of pointing at `loadLedger8()`.

packages/protocol/src/lib/engine/down-convert.ts
```diff
@@ -0,0 +1,298 @@
...
+  switch (variant) {
+    case 'boundedMerkleTree':
+      checkRoot(sv.asBoundedMerkleTree()!);
+      return;
+    case 'array':
+      sv.asArray()!.forEach((element) => assertMerkleTreesRehashed(element));
+      return;
+    case 'map': {
+      const map = sv.asMap()!;
+      map.keys().forEach((key) => assertMerkleTreesRehashed(map.get(key)!));
```
VanessaPC, line 219: **low/medium — the `!` assertions are unsound on this docblock's own terms**

The docblock at line 202 says the vendor typings "are not authoritative about definedness here (`asCell()` is declared non-optional yet returns `undefined` for a `null` state value)", and the `default` arm at line 225 exists because a caller-injected runtime "can emit a tag the pinned `.d.ts` does not declare". Granting both premises, the assertions can't be relied on:

- a runtime whose `type()` reports `'map'` while `get()` fails to resolve a key marshalled out by `keys()` → `assertMerkleTreesRehashed(undefined)` → bare `TypeError` on `undefined.type()`
- same shape for `sv.asArray()!` (line 215) and `sv.asBoundedMerkleTree()!` (line 212) → `checkRoot(undefined)`

Inside `downConvertForExecution` that's wrapped, but `assertMerkleTreesRehashed` and `checkRoot` are both exported and documented as throwing only `MerkleNotRehashedError` / `DownConvertFailedError`. A direct caller therefore gets an uncoded error out of a seam whose whole contract is code-based discrimination. An explicit `undefined` check throwing `DownConvertFailedError` is one line each.

packages/protocol/src/errors.ts
```diff
@@ -77,14 +81,221 @@ export class UnknownProtocolVersionError extends Error {
...
+ * to match; `errors.test.ts` holds that line.
+ */
+const PUBLISHED_SCOPES = Object.freeze(['@midnight-ntwrk', '@midnightntwrk'] as const);
+
+/**
+ * The unscoped npm name of each axis. Both published copies of an axis carry
+ * this same name under a different scope, so naming only one scope would point
+ * every consumer installed from the other at a package not in their tree.
+ */
+const AXIS_BARE_PACKAGE_NAMES: Readonly<Record<Ledger8InstanceAxis, string>> = Object.freeze({
```
VanessaPC: **low — fail-open prototype hole, the same one this PR deliberately closed for `ENVELOPE_DECODERS`**

This table is `Object.freeze`d but keeps its prototype, and `axis` is never validated before `axisPackageNames` indexes it at line 135. For the untyped-JS consumers these errors are explicitly written for:

- `assertSharedLedger8Instance('constructor', a, b)` → message reads `Trace @midnight-ntwrk/function Object() { [native code] } and @midnightntwrk/function Object() { [native code] } with your package manager's ``why`` command`
- `'__proto__'` or any other non-axis string → `@midnight-ntwrk/undefined`

`envelope.ts:71` gets exactly this right (`Object.create(null)` + validate-before-dispatch, with a documented rationale). Same treatment here — `Object.create(null)` plus a guard in `assertSharedLedger8Instance` — closes it.

packages/protocol/src/lib/engine/instance-guard.ts
```diff
@@ -0,0 +1,78 @@
...
+ * are always `===` to each other, so a caller that optional-chained a missing
+ * export on both sides (or simply forgot to pass a probe) would otherwise
+ * pass this fail-fast safety net by accident instead of failing it.
+ *
+ * An axis earns an assertion only when the package genuinely reaches this
+ * process through two acquisition paths — see {@link Ledger8InstanceAxis} for
+ * why `'onchain-runtime-v3'` is the only member today.
+ */
+export const assertSharedLedger8Instance = (axis: Ledger8InstanceAxis, probeA: unknown, probeB: unknown): void => {
+  if (probeA == null || probeB == null || probeA !== probeB) {
```
VanessaPC: **low — nullish probes are reported with a factually false diagnosis**

Rejecting nullish probes before the `===` is right, and the reasoning in the docblock is sound. The problem is the error they get: `Ledger8InstanceMismatchError`'s message asserts "Detected two physically distinct copies of onchain-runtime-v3 loaded into the same process (a dual-instantiation)" and sends the reader to `npm why` to hunt a duplicate install.

A caller who optional-chained an export that moved, or simply omitted a probe, is told something untrue about their dependency tree and pointed at a hunt that will find nothing. This PR adds `Ledger8RuntimeInvalidError` for precisely the "handed over incomplete" fault at the envelope seam; as it stands the two seams diagnose the same class of caller error differently.

packages/protocol/src/errors.ts
```diff
@@ -77,14 +81,221 @@ export class UnknownProtocolVersionError extends Error {
...
+      'Attempted to read the root of a bounded Merkle tree before it was rehashed. This usually means the ' +
+        'tree was built or updated in memory and never rehashed — call rehash() on it before executing ' +
+        'against it.',
+      { cause }
+    );
+    this.name = 'MerkleNotRehashedError';
+  }
+}
+
+/**
```
VanessaPC: **low — this doc block is dropped, and `UnknownLedgerVersionError` ships undocumented**

Two JSDoc blocks sit back to back: this one (lines 242–257) documents `UnknownLedgerVersionError` — it names `version`, `LEDGER_VERSIONS` and `requestedVersion` — and line 258 opens a second block documenting `Ledger8RuntimeInvalidError`.

A doc comment immediately followed by another doc comment is discarded, so the class actually declared at line 290 (`UnknownLedgerVersionError`) ships with no TypeDoc output and no IDE hover, while the block that *does* survive describes a different class. Looks like residue from the #1168 conflict resolution described in the PR body. Moving this block down to line 289 fixes both halves.

VanessaPC (commented):
Reviewed the three new engine modules, the error classes, and the test suites against the base (`feat/1004-v8-subpath-loadv8`).

The core is solid, and I checked the parts that would be easy to get wrong rather than taking the docblocks' word for it: `structurallyEqual`'s key-set guard and its `Map`/`Uint8Array`/`BigInt`/empty-container edges, the null-prototype decoder table against `constructor` / `toString` / `__proto__`, the exhaustiveness `never` plus its non-redundant runtime throw, `DownConvertStage` staying inside its union, `readHexFixture`'s whole-hex length check, the golden-fixture decode matrix (all four golden assertions match the recorded outcomes, all five referenced fixtures exist), the split-scope defence in `errors.ts` against the pack-time rewrite (which does match on the scope *with* its slash, so it holds), and the `onchain-runtime-v3-alt` alias genuinely producing a second physical copy under `nodeLinker: node-modules`.

Eight comments inline. Two I'd like resolved before merge:

- **`down-convert.ts:276`** — `downConvertForExecution` doesn't validate `ledger8Runtime`, so a runtime fault surfaces as `DownConvertFailedError` telling the caller to audit input bytes that are fine. That's the mis-diagnosis `Ledger8RuntimeInvalidError` was added to prevent, at the one seam that skipped the check.
- **`errors.ts:283`** — the remediation tells callers to pass the `ContractState` that `loadLedger8()` exposes. I pulled the published `@midnightntwrk/ledger-v8@8.1.1` typings: it exports its own `ContractState` with a `static deserialize`, and has zero dependencies, so it cannot be onchain-runtime-v3's. Following the message yields a class that *passes* the `typeof deserialize === 'function'` guard and decodes on a second WASM instance — the dual-instantiation `instance-guard.ts` exists to catch.

The rest are low: three one-to-three-line fixes (`errors.ts:130` prototype hole, `instance-guard.ts:75` false dual-instantiation claim, `errors.ts:242` dropped JSDoc), plus the `!` assertions at `down-convert.ts:212/215/219` on exported functions, the v9 path requiring an unused runtime, and an inverted comment premise in `vitest.config.ts`.

packages/protocol/src/lib/engine/envelope.ts
```diff
@@ -0,0 +1,124 @@
...
+export const extractEncodedStateValue = (
+  raw: Uint8Array,
+  version: LedgerVersion,
+  ledger8ContractState: Ledger8ContractState
+): EncodedStateValue => {
+  const decoder = ENVELOPE_DECODERS[version];
+  if (typeof decoder !== 'function') {
+    throw new UnknownLedgerVersionError(String(version));
+  }
+  if (typeof ledger8ContractState?.deserialize !== 'function') {
```
VanessaPC, line 133: **low — the v9 path requires a runtime it provably never uses**

The v9 decoder (line 75) never reads `ledger8ContractState`, yet this check rejects its absence for every version. The docblock defends that as keeping "a v9 caller from drifting into a v8 call that has no runtime to reach for", but the cost runs against the goal this same file states twice — that "a v9-only consumer would pay for a runtime it never calls".

Concretely: a node-2.x-only consumer must acquire the retained pre-fork onchain-runtime-v3 runtime purely to extract a post-fork envelope, and `extractEncodedStateValue(raw, 'v9', undefined)` throws `Ledger8RuntimeInvalidError` for an argument that is dead on that path.

An optional third parameter, checked inside the `'v8'` decoder (or right after the version validation, only when `version === 'v8'`), gives the same accurate diagnosis without the forced acquisition.

packages/protocol/vitest.config.ts
```diff
@@ -34,7 +34,28 @@ export default defineConfig({
...
+        // instrumentation ever times out against a large/slow WASM call in one
+        // of these suites, exclude that suite from instrumentation (vitest
+        // coverage.exclude) instead — never fix a timeout by padding the test
+        // timeout.
+        //
+        // Two properties of this mechanism are worth knowing before editing:
+        // a glob matching no file is ignored silently, so renaming or moving
+        // one of these files deletes its floor without any warning; and while
+        // the global thresholds above are also 100, these entries are
+        // redundant. They exist so that a future lowering of the global floor
```
VanessaPC, line 52: **low — the comment's premise is inverted, which matters exactly when someone edits this**

In Vitest, files matched by a `thresholds` glob key are *removed from the global threshold aggregate* and judged only against their own entry. So these four entries aren't redundant with the global 100s — they change which files the global gate covers.

No behavioural difference while every number here is 100. But this comment is what a future editor would rely on when lowering one of these floors, and at that moment the file also silently stops contributing to the package-wide gate — the opposite of the protection the comment promises ("a future lowering of the global floor cannot quietly take these files down with it").

The neighbouring warning about a glob matching no file being silently ignored is accurate and worth keeping.

packages/protocol/src/lib/engine/down-convert.ts
sp-io: Fixed in `fc8f22d`. `downConvertForExecution` now validates before the `try`:

```ts
if (typeof ledger8Runtime?.StateValue?.decode !== 'function') {
  throw new Ledger8RuntimeInvalidError('StateValue.decode');
}
if (typeof ledger8Runtime.ChargedState !== 'function') {
  throw new Ledger8RuntimeInvalidError('ChargedState');
}
```

Only the two members this function actually dereferences are checked. `ContractState` is on `Ledger8CompactRuntime` to pin the era and is never called here, so guarding it would reject a runtime that works.

Five cases red first, each pinning the code and the `missingMember`: no runtime, `null`, a runtime with no `StateValue`, a `StateValue` whose `decode` is not callable, and a runtime with no `ChargedState`.

packages/protocol/src/errors.ts
sp-io: Verified the factual half independently rather than taking the quote. `npm pack @midnightntwrk/ledger-v8@8.1.1` gives `ledger-v8.d.ts:765 export class ContractState { ... static deserialize(raw: Uint8Array): ContractState }` with `data: ChargedState`, and a `package.json` carrying neither `dependencies` nor `peerDependencies`. So the message did point callers at a different package's class running on its own WASM instance.

Fixed in `fc8f22d`: the remediation now names onchain-runtime-v3, says plainly that this package exposes no accessor for it, and mentions `loadLedger8()` only to rule it out. The test that pinned `/loadLedger8/` was replaced by one asserting the message names onchain-runtime-v3 and the absent accessor.

One part I read differently, and it is why the fix is the string and not new machinery: "routes past the check meant to catch the failure it causes" overstates the exposure. Nothing crossing this seam is a WASM handle. `deserialize(raw).data.state.encode()` returns a POJO, and that POJO is then decoded by onchain-runtime-v3 inside `downConvertForExecution`. The failure `instance-guard.ts` exists for - a wasm-bindgen `_assertClass` rejection, or a `__wbg_ptr` read against the wrong linear memory - is not reachable this way. What is left is narrower: a schema divergence between the two encoders' POJOs, which a guard on class identity would not catch either.

packages/protocol/src/lib/engine/down-convert.ts
sp-io: The inconsistency is real and I am not disputing it: the docblock says the vendor typings are not authoritative about definedness, the `default` arm exists precisely because a caller-injected runtime can emit a tag the pinned `.d.ts` does not declare, and then three `!` trust those same typings.

Not in this round, and not because it is wrong. The package holds a per-file 100% branch floor, so each guard needs a fake-runtime test forcing `asBoundedMerkleTree()` / `asArray()` / `map.get()` to return `undefined` - three branches plus three tests, rather than one line each. Recorded as an explicit follow-up in the PR body instead of being folded into a round that was otherwise message and doc fixes.

packages/protocol/src/errors.ts
sp-io: Fixed in `fc8f22d`. `AXIS_BARE_PACKAGE_NAMES` is now `Object.freeze(Object.assign(Object.create(null), { ... } satisfies Record<Ledger8InstanceAxis, string>))`, and `assertSharedLedger8Instance` validates `axis` against a closed set before either probe is looked at, throwing a new `UnknownLedger8AxisError` (`MIDNIGHT_JS_P_UNKNOWN_LEDGER8_AXIS`) that keeps the offending string in a field rather than the message.

The axis set lives in `instance-guard.ts`, not in `errors.ts`. `errors.ts` is a build entry, so a predicate exported from it becomes public API that nothing outside the engine needs - and `version.ts` sets the precedent of exporting the tuple but no predicate. The one duplicated literal is checked, not trusted: `satisfies Record<Ledger8InstanceAxis, true>` fails the build if the union gains a member.

Covered by `constructor` / `__proto__` / `toString` / `valueOf` / `bogus`, a non-string axis, and a direct pin on the table asserting no `[native code]` or `[object` reaches the message.

One distinction worth recording next to the `ENVELOPE_DECODERS` comparison: that hole was fail-open - it returned a value and threw nothing. This one only garbled the text of an error that was already being thrown, and only when a bogus axis coincided with a genuine probe failure. Same fix, different severity.

packages/protocol/src/lib/engine/instance-guard.ts
sp-io: Fixed in `fc8f22d`. Nullish probes now throw `Ledger8RuntimeInvalidError` with `missingMember: 'onchain-runtime-v3 instance probe'`, and `Ledger8InstanceMismatchError` is reserved for an actual reference-equality failure, so the two seams diagnose the same class of caller error the same way.

The guard is now three ordered checks - axis, then nullish, then `!==` - which also means `axis` is known to be one of this package's own literals before it reaches `missingMember`, keeping that field's "never caller-supplied text" property intact.

The three existing nullish cases were flipped to assert the new code and to assert the message does *not* contain `dual-instantiation`, so a regression that reintroduced the false diagnosis fails rather than passing on a matching `axis`.

packages/protocol/src/errors.ts
sp-io: Fixed in `fc8f22d`. The block moved down to the class it documents; confirmed in the emitted typings rather than assumed - `dist/errors.d.ts` now carries it directly above `declare class UnknownLedgerVersionError`.

Also swept the rest of the file for the same shape: no other back-to-back doc blocks. `UnknownLedger8AxisError` was added in the same round and sits above `UnknownLedgerVersionError`, so it has its own block and does not re-orphan this one - which it did on the first attempt, before the sweep caught it.

packages/protocol/src/lib/engine/envelope.ts
sp-io: Kept as it is; the docblock is corrected in `64a5068`.

Two things came out of checking this one. First, the stated cost does not exist on this stack. `extractEncodedStateValue` has exactly one call site anywhere in it - `createLedger8Engine()` in #1168 / #1165 - and that factory has already awaited onchain-runtime-v3 two lines earlier to assemble the runtime it hands to `downConvertForExecution`. The v9 decoder is there so the *bridge* can read a post-fork envelope before down-converting it, not to serve a caller holding no pre-fork runtime; such a caller reads ledger-v9 directly and never reaches this function. So no consumer is currently forced to acquire anything.

Second, the shape of the alternative matters. That call site is `(raw, version) => extractEncodedStateValue(raw, version, ocrt3.ContractState)`, where `version` is a `LedgerVersion` union value rather than a literal - so overloads, or a discriminated parameter object, would not compile against it. The only workable version of your suggestion is the optional third parameter, which trades a compile-time error on the one path that genuinely needs the argument for a runtime one.

What was actually wrong was the justification. The comment defended the requirement by the cost to a v9-only consumer while the same file argues laziness on a different axis (the erased `import type`, which is about bundling, not about the call contract). The docblock now separates those two, gives the call-graph reason, and records the condition that flips the decision: if this seam is ever surfaced beyond the engine, make the parameter optional and move the check into the `'v8'` decoder.

packages/protocol/vitest.config.ts
sp-io: Not changing this one - the comment is correct for the pinned toolchain, and I think the premise here is the inverted one.

Vitest 4.1.7, `node_modules/vitest/dist/chunks/coverage.*.js`, `resolveThresholds`:

```js
// Global threshold is for all files, even if they are included by glob patterns
for (const file of files) {
  const fileCoverage = coverageMap.fileCoverageFor(file);
  globalCoverageMap.addFileCoverage(fileCoverage);
}
```

The loop runs over every file with no exclusion of glob-matched ones; the per-glob entries are pushed as additional threshold sets alongside it. So these four entries are an extra gate on top of the global 100s, not a replacement for them, and lowering one of these floors does not drop that file out of the package-wide gate.

The exclusion behaviour you describe was real in earlier majors and is still what a lot of the circulating documentation says, so this is worth having on the record rather than left to be re-raised. The neighbouring warning about a glob matching no file being silently ignored stands as you say.

VanessaPC (approved):

VanessaPC (approved):

PR #1165, opened 2026-08-19
feat(midnight-js): add v8-native tx composition incl. deploy machinery

sp-io:
# Overview

**PR 1004-E2 of the MJS-01 series (#1004)** — the ledger-8 engine composition legs: v8-native call composition, deploy machinery and constructor execution, extending the `loadLedger8Engine` facade to its full 7-method surface. Third of three PRs carved out of the original 1004-E for reviewability — the three (#1167 E0 fixtures, #1168 E1 execution+facade, this) originally summed to a tree byte-identical with the fully reviewed original head; #1168's review follow-ups are all merged in here — `f002525a` (facade passes `Ledger8InstanceMismatchError` through unwrapped), `0aacc76d`'s pass, and its latest (`fc8f22da`, `64a50686`), see the inherited sections at the end. Merge order: #1155 → #1156 → #1159 → #1164 → #1167 → #1168 → this.

## Summary

- **Public facade — the engine's ONLY surface (owner-approved design):** `loadLedger8Engine(): Promise<Ledger8Engine>` (`src/engine/load-engine.ts`), exported from the root barrel. Lazy end-to-end: engine internals and the retained-era WASM (ledger-v8, onchain-runtime-v3, 0.16 glue) load via dynamic imports only — the root bundle keeps zero static WASM linkage (dist-laziness gate extended and asserting it). Construction acquires only the 0.16 glue and `onchain-runtime-v3`, and runs `assertSharedLedger8Instance` once (single `onchain-runtime-v3` axis) before anything else; it deliberately does NOT acquire ledger-v8, because only the two composition legs need it (see the review follow-up below). Methods (all options-object based): `extractState`, `downConvertForExecution`, `executeCircuit`, `wrapKeepStateCall`, `executeConstructor` are synchronous; `composeCallTx` and `composeDeployTx` return `Promise<Uint8Array>` and acquire the v8 module on first use via the memoized `loadLedger8()`. The 1004-D primitives are its internals.
- **Circuit execution on the retained 0.16 stack** (`src/engine/execute.ts`): raw `createCircuitContext` sequence (no compact-js), transcript POJO output. Executable golden fixture: `testkit-js/.../fixtures/hf/counter-016/` — the spike's 0.16-compiled counter ported byte-verbatim (git-blob-hash-verified; the 0.18-era twin cannot run on the 0.16 glue), plus a self-minted golden transcript (the spike recorded none; execution is deterministic, no proving involved).
- **Keep-state v9 wrap** (`src/engine/wrap-v9.ts`): resolves the circuit's REAL registered operation from the migrated v9 `ContractState` (a blank `ContractOperation` is rejected by this repo's own composition guard — caught in review, fixed); end-to-end tested with a real non-empty transcript crossing ocrt3 → v9 `partitionTranscripts`.
- **v8-native composition + deploy machinery** (`src/engine/compose-v8.ts`, `src/engine/deploy-v8.ts`): call leg (via the shared `assembleCallPrototype` from 1004-E1, stage `'call-operation'` → intent → `fromPartsRandomized`); deploy leg validates the verifier-key map against the state's declared entry points in BOTH directions before registering anything via `setOperation`, then `fromParts` (fixed segment, matching `createUnprovenLedgerDeployTx` in `packages/contracts`). Outputs are tag-prefixed serialized `Uint8Array`s round-tripped through the v8 decoder, and read back apart to assert what was actually composed. No proving in tests.
- **Errors:** two registry codes. `MIDNIGHT_JS_P_LEDGER8_COMPOSE_FAILED` / `Ledger8ComposeFailedError` with a public `stage` discriminant (`'wrap-call' | 'call-operation' | 'call-verifier-key' | 'deploy-verifier-key' | 'deploy-unknown-circuit' | 'deploy-ambiguous-circuit' | 'deploy-verifier-key-blob'`; the last one wraps the ledger's own failure on `cause`), and `MIDNIGHT_JS_P_LEDGER8_COMPOSE_OPTION_INVALID` / `Ledger8ComposeOptionError` with an `option` discriminant (`'contractState' | 'networkId' | 'ttl'`) for options the WASM would otherwise absorb silently.
- **Deps:** `compact-runtime-ledger8` (`npm:@midnight-ntwrk/compact-runtime@0.16.0`) as a protocol dependency — verified to resolve to real 0.16.0 despite root `resolutions` pinning the ident to 0.18.0-rc.1 (alias descriptors are distinct idents for Yarn's resolutions matching).
- **Gates:** ESLint ACL and dist-laziness extended for the glue alias, ocrt3 and the `./engine` subpath (canary-tested); per-glob 100% coverage entries for all new engine files.

### Notes for reviewers (documented deviations & known items)

- **The brief's `parseSerializedTag` round-trip is moot:** that parser no longer exists — `7a6a9342` ("drop parseSerializedTag until a real caller needs it"), earlier in this same stack, removed it and its tests. Tests assert the exact tag literal plus a real-decoder round-trip instead; the consumer-layer story lands with MJS-02.
- **`transaction[v9]` on a v8 transaction is correct:** wire-format numbers lag era numbers in this ecosystem (cf. `contract-state[v6]` ↔ the 0.16 stack). Confirmed ledger-v8-genuine three ways, incl. a dual-acquisition round-trip (facade-acquired module serializes, independently imported module deserializes byte-identically).
- **Deploy-then-call ergonomics:** `composeDeployTx` returns only the serialized tx (brief-mandated); recovering the deployed address/key-registered state is an integration-milestone item.
- **`networkId`/`ttl` policy stays with the caller, but well-formedness does not.** The earlier claim that the WASM rejects malformed values downstream was wrong: `Intent.new(new Date('nope'))` does not throw and reads back as the Unix epoch, and `fromParts('')` is accepted. Both legs now refuse an empty network id and an invalid TTL up front (`Ledger8ComposeOptionError`). Which network and which TTL policy remain integration-milestone decisions.
- Commit `0e3d0384`'s message says "devDependency" for the glue alias; the diff correctly declares it under `dependencies` (runtime import).

### Review follow-up (`c999718e`)

A full review of this PR found nine issues; all are fixed in that commit. Each was reproduced against the pinned ledger-v8 WASM before being changed.

- **Deploy could widen the contract's entry-point set, and silently change its address.** The verifier-key map was validated in one direction only, and `setOperation` *creates* a slot — so a key naming a circuit the state does not declare was registered without complaint, and because `ContractDeploy` derives its address from the initial state, the deploy landed at a different address than the caller's artifacts describe (measured: `df150065…` vs `37dce3c6…`). Now validated both ways before any registration; new stage `'deploy-unknown-circuit'`. This also removed the post-registration re-read, and with it the code's reliance on ledger-v8 reporting a blank slot as `undefined`.
- **The call leg accepted an operation with no verifier key** — precisely what a constructor-built state carries — composing a call no ledger could verify. Both legs now reject it (`'call-verifier-key'`).
- **Two WASM crossings threw raw, uncoded errors** (the contract-state bridge and the verifier-key setter). Both are wrapped with `cause`, mirroring `envelope.ts`; the setter failure names the circuit it belongs to.
- **`String(circuitId)`** rendered a byte-array entry point as decimal bytes, breaking `Ledger8ComposeFailedError`'s own "never renders byte contents" contract. Replaced by `entryPointName()`, with both arms tested and a test pinning the ledger behaviour it exists for.
- **Engine construction no longer instantiates ledger-v8.** Every consumer — including the keep-state path, which never touches the v8 axis — was paying for multi-megabyte WASM and a hard ledger-v8 dependency. The two composition methods are now async and acquire the module per call; `engine-load-engine-v8-laziness.test.ts` pins that contract. Side effect: `assertSharedLedger8Instance` now runs ahead of any v8 acquisition, so a dual-instantiation can no longer lose the `Promise.all` race and be reported as a missing runtime.
- **One test was vacuous.** `communicationCommitmentRandomness()` and `fromPartsRandomized` make *any* two composed transactions differ, so asserting byte-inequality between a registered and a blank operation could not fail. Replaced by capturing the `ContractCallPrototype` argument (via a subclass of the real class — no cast) and asserting it by value, along with `entryPoint`, `keyLocation` and `address`.
- **Coverage gaps closed:** the guaranteed/fallible transcript split (swapping those two arguments previously passed the whole suite), per-circuit key registration, `networkId`/`ttl`, the round-0 assertion that read a bypass call rather than `executeConstructor`'s own result, an AAA violation, two tautological assertions, and the facade's `typeof` sweep (now strict key equality).
- **The dist-laziness gate had a hole of its own:** it could not see bare side-effect imports, and applied its eager-closure walk to only three of its eight assertions.
- **Comment accuracy:** removed a block documenting `parseSerializedTag` (deleted earlier in this stack), the references to the internal spike repository, the unattested `VerifierKeyNotSet` claim, and a "fully deterministic" over-claim; fixed dead `{@link}` targets; documented the `fromParts` / `fromPartsRandomized` asymmetry against its precedent in `packages/contracts`.

Not changed, and why: the two spike references in `execute.ts` belong to #1168 and are left to that PR; `assembleCallPrototype`'s `partitionTranscripts` guard stays an uncoded `Error` but is now documented as an internal ledger-module invariant rather than a caller error; the attribution test uses one verifier key for two circuits because the fixtures commit exactly one real key (the comment says what that does and does not prove).

### Review follow-up (`2eb82345`)

A second review pass found one more hole in the deploy leg, fixed in that commit.

- **A deploy could leave a declared slot blank and create an undeclared one beside it.** ledger-v8 declares `operations()` as `Array<string | Uint8Array>` and accepts either form, so resolving each declared entry point to a *name* and then doing everything through that name is lossy twice over: two declared entry points that resolve to one name collapsed into a single `Set` member — making both directions of the map-vs-declared check agree while one of the two slots went unregistered — and `setOperation` then ran against the resolved name, so a byte-declared slot stayed blank while a second, undeclared entry point was created next to it. That is the same silently-different-address outcome `'deploy-unknown-circuit'` already guards. The `Set` is now a `Map` from resolved name to the declared entry point: the checks still run on names (because `verifierKeys` is keyed by name), registration uses the declared entry point, and a name that resolves twice throws the new `'deploy-ambiguous-circuit'` stage before either check runs.
- Neither case is reachable through ledger-v8 today — it decodes every entry point back to a string, pinned by the `entryPointName` suite — so the new cases reach the byte half of the declared union through a `ContractState` subclass rather than a cast.

## Test plan

- [x] Tests written first (TDD), verified failing before implementation (incl. mutation-grade checks: operation-resolution asserts the captured constructor argument's serialized value)
- [x] All existing tests pass — full repo suite green (`yarn check:core`); protocol 332/332 at 100/100/100/100 coverage; utils 279/279
- [x] Lint, `typecheck:tests` (pre-push hook) and build all clean; laziness gates green against built dist

## Submission Checklist

- [x] Self-reviewed the diff
- [ ] Reviewer requested
- [x] Update README.md file (if relevant) — fixtures README extended (counter-016 provenance)
- [x] Update documentation (if relevant) — n/a beyond in-repo docs
- [x] No new todos introduced

## Links

- Part of #1004 (MJS-01). Depends on #1155, #1156, #1159, #1164.

## Inherited: review fixes from #1168 (`0aacc76d`)

Took the merge of #1168's review pass. Two of its fixes change behaviour on this
branch's own legs:

- **Canonical key location.** `assembleCallPrototype` now builds the
  contract-qualified `encodeContractKeyLocation` form instead of passing the bare
  circuit id, so `composeV8CallTx`'s prototypes are resolvable through
  `ZKConfigRegistry`. `engine-compose-v8.test.ts` pinned the bare id and now pins
  the canonical form.
- **Zswap fail-fast.** `executeCircuit` refuses a circuit that recorded Zswap
  inputs or outputs (`Ledger8ZswapUnsupportedError`) rather than returning a
  transcript that silently omits them — this branch's compose legs would
  otherwise assemble an unbalanced transaction from one.

Conflict resolutions kept this branch's wider surface (six-stage union,
`Ledger8ComposeOptionError`, all seven engine methods) and layered the parent's
corrections on top: the total `MESSAGES` `Record` replaces the if-chain
fallthrough, and the `distV8Exists` gating is dropped because `loadLedger8`
imports `'../v8.js'`, which vitest resolves to `src/v8.ts` — the composition
suites never needed a build.

Protocol suite on this branch at that point: 273/273 at full coverage, lint
and `typecheck:tests` clean.

## Inherited: review fixes from #1168 (`77d0309f`)

Took the merge of #1168's latest review pass (`fc8f22da`, `64a50686`). All of it
is diagnosis and docs on the shared engine seams this branch builds on — no
success path changes, and no conflict with this branch's own legs (the errors
registry and `protocol-acl` merged cleanly, and the new axis guard is called
with the one axis `loadLedger8Engine` already passes).

- `downConvertForExecution` now validates the runtime binding it dereferences
  and fails with `LEDGER8_RUNTIME_INVALID` naming the absent member, instead of
  relabelling a `TypeError` as `DOWN_CONVERT_FAILED` and sending the caller to
  audit input bytes that are fine.
- `Ledger8RuntimeInvalidError`'s message named the wrong package — following it
  passed the duck-typed guard and decoded on a second WASM instance. It now
  names `onchain-runtime-v3` and says this package exposes no accessor for it.
- `UnknownLedgerVersionError`'s doc block was being discarded; moved onto the
  class it documents.
- A nullish instance probe is reported as `LEDGER8_RUNTIME_INVALID` rather than
  as a dual-instantiation, so the caller is not sent to `npm why` after a
  duplicate that is not there.
- The axis package-name table is null-prototype and frozen, and the axis is
  validated against a closed `satisfies`-checked set before any probe is
  compared, throwing the new `UnknownLedger8AxisError`.
- Docs: the unconditional-runtime-argument rationale is restated on the real
  call graph, and records the condition that would flip it.

Verified on this branch after the merge: protocol 332/332 at 100/100/100/100
coverage, lint, `typecheck:tests` and build clean, laziness and dist gates green
against a fresh build.


VanessaPC (approved):

packages/protocol/src/test/dist-engine-errors.test.ts
```diff
@@ -63,19 +63,31 @@ const transcriptForUnregisteredCircuit = () => ({
...
-// CI the build always precedes the tests, and this gate is the only cover for
-// cross-bundle error identity, so a missing artifact must fail there rather
-// than vanish behind a skip.
-describe.skipIf(!distBundlesExist && !process.env.CI)('dist engine error gate', () => {
+// Never skipped when a bundle is absent: a skip is reported as a pass, so the
+// gate would silently stop guarding — and it is the only cover for
+// cross-bundle error identity. Every orchestrated run has the build ahead of
+// it — turbo's `test` task dependsOn `build` — so a missing bundle is a real
+// failure, called out by the first case below before the imports fail.
+describe('dist engine error gate', () => {
```
sp-io, line 71: Found while syncing the no-skip fix down the stack: this gate had the same shape (`skipIf(!distBundlesExist && !process.env.CI)`), so it self-disabled outside CI — and it is the only cover for cross-bundle error identity. Now unconditional, with the missing-bundle case called out by the first `it` before the imports fail.

The merge also kept this branch's stronger closure-based laziness gates (`eagerContents`/`eagerClosureOf`) and routed their reads through the fail-fast `readDistFile()`.

Verified here: 333 passed, 0 skipped, coverage 100%; lint clean.

PR #1166, opened 2026-08-19
fix(midnight-js): serialize LevelDB access in private state provider

### LINKED ISSUE #1144: [level-private-state-provider@5.0.0-beta.6] concurrent contract calls sharing one provider crash with 'Database failed to open'
**Environment:** `@midnight-ntwrk/midnight-js` 5.0.0-beta.6 (`level-private-state-provider`), node 2.0.0-rc.4 / indexer 4.4.0-pre-alpha.16 / proof-server 9.0.0-rc.5_experimental (local stack), Node 22, Ubuntu. Hit while battle-testing concurrent contract calls in a bridge orchestrator prototype (sig-net erc20-vault fork).

## What we hit

Firing two contract circuit calls CONCURRENTLY from one wallet — one shared providers object, hence one `levelPrivateStateProvider` instance — fails the second call inside the private-state LevelDB open:

```
Error: Unexpected error executing scoped transaction '<unnamed>':
Error: Database failed to open
```

The first call proceeds normally. Reproduced deterministically on two separate attempts (`Promise.allSettled` over two calls to the same contract via `findDeployedContract` handles sharing providers). Retrying the failed call sequentially succeeds every time.

## Expected vs actual

Expected: either concurrent calls per provider instance are supported (internal serialisation of the LevelDB access), or the provider documents that call-level concurrency is unsupported so integrators serialise deliberately. Actual: an internal LevelDB error surfaces mid-transaction with no hint that concurrency was the cause — the first diagnosis costs a debugging session, because the error points at the database rather than at the calling pattern.

## Impact

Friction / docs-gap. Any orchestrator or backend multiplexing users over shared provider instances will hit this on its first concurrent load. (For our own state machine it became a requirement: serialise ALL circuit calls per wallet.)

## Suggested fix

Either an internal mutex around the scoped-transaction private-state access, or a documented "one in-flight call per provider instance" contract plus a clear error message naming the concurrent-access cause.

sp-io:
## Summary

- Fixes #1144 — concurrent operations on one `levelPrivateStateProvider` instance failed with `Error: Database failed to open`. Two concurrent contract circuit calls sharing one providers object overlap with near certainty, so this hit any orchestrator on its first concurrent load.
- **Root cause:** `withSubLevel` creates a fresh `Level` instance per operation and does `open() → operate → close()`. All 11 provider operations funnel through it. On the Node.js path `classic-level` holds the database directory exclusively, so a second `open()` — or an `open()` racing another operation's `close()` — rejects. The package already guarded salt initialization and password rotation this way, but never general operation-level concurrency.
- **Fix:** queue database operations per database name (`withDbLock`), so overlapping callers are serialized instead of colliding. Serializing costs no throughput: each operation already held the database exclusively — today the overlap does not run in parallel, it crashes.
- The queue is module-level because the resource it guards is process-wide. A per-provider queue could not see a second provider opening the same database. Entries are removed once a queue drains, so the map holds at most one entry per database in use.
- `migrateSublevel`, which opens a `Level` directly, goes through the same queue and the same open helper.
- Also reports the real reason for a failed database open instead of surfacing a bare `Database failed to open`.

### Why not a shared open handle

A single refcounted `Level` per database would be faster (one open per process instead of per operation), but the handle then needs a lifecycle owner and neither placement works: per-provider means the first provider holds the lock indefinitely and any provider-per-request pattern deadlocks; module-level means the lock is held for the process lifetime with no way to release it short of a new `close()` method. That trades a bug fix for a change in the public contract. Not now.

### Scope

No public API change, no on-disk format change, no change to sequential behavior.

## Review follow-ups (second commit)

A four-agent review found five confirmed defects in the first commit, all fixed here:

- **The new error message was useless.** `abstract-level` reports every open failure as the same constant `'Database failed to open'` (`abstract-level.js:1113`), with the real reason one level down in `cause`. The wrapper interpolated that constant and then asserted a single confident diagnosis ("another process"), so a permissions, disk or corruption failure sent the reader hunting a phantom process. It now reads the reason from `cause` and lists the possible causes.
- **`level.close()` could be skipped.** `finally { await subLevel.close(); await level.close(); }` — a rejecting sublevel close meant the database handle was never closed, keeping the LevelDB lock held for the rest of the process and making *every* subsequent operation fail with exactly the error this branch fixes. Now `level.close()` always runs.
- **Close failures masked the error that mattered.** A close rejection replaced the operation's error — including `Old password is incorrect` and decryption failures, the most important messages in this package. Close failures are now suppressed on the error path and reported on the success path, where they can mean a write never reached disk.
- **Dead code in the queue.** In `pending.then(operation, operation)` the rejection handler is unreachable, because the stored tail can never reject. Two reviewers confirmed it independently; one instrumented both handler slots and the rejection slot ran zero times. Removed, and the comment that documented the dead mechanism corrected.
- **The doc comment was wrong about the browser.** `level` v10 resolves to `browser-level` (IndexedDB) under a bundler, which takes no exclusive lock and has no OS lock file, but the comment stated the exclusive lock as an unconditional property. Now qualified per path. The comment also records that the lock is **not reentrant** — the invariant that makes the current callers safe is that they all resolve encryption keys before entering the lock, and nothing said so.

Test changes from the same review: added coverage for the open-failure path (the only uncovered lines in the diff) and for queue recovery after a rejection; three tests that asserted only "nothing rejected" now assert observable effects; `beforeEach` isolation replaces a shared database across the block; and one nondeterministic test was dropped — a reviewer measured that queue-entry order does not follow call order, because `set` resolves the rotation lock, key derivation and encryption *before* entering the queue, so "last writer wins" cannot be asserted.

## Known limitation, not fixed here

`changePassword` racing `set` can write a value under the pre-rotation key, leaving it permanently undecryptable. `set` passes the rotation-lock check, resolves its encryption key, and only then enters the queue; rotation can complete in that window. This hazard **pre-exists on `main`** — but before this PR the racing write usually died at `Database failed to open`, and now the open succeeds and the corruption lands, so the failure mode changes from loud to silent.

It is deliberately out of scope: the same time-of-check/time-of-use gap affects every write path, including `remove` and the re-encryption writes on the read path, so a guard in `set` alone would give false confidence. It needs its own design and tests, tracked in #1169, which should land before this ships to consumers.

Two smaller documented limitations: the queue key is the configured database name verbatim, so two configs naming one directory differently (`'db'` vs `'./db'`) are not serialized against each other — normalizing would be wrong on the browser path, where the name is not a path; and concurrency across separate processes still contends on the operating system lock.

## Test plan

- [x] Tests written first (TDD), verified they fail before fix — the triage repro reproduced 17/20 rejections; a reviewer independently reverted the source and confirmed 4 of 5 tests go red with `LEVEL_LOCKED` / `lock ... already held by process`
- [x] All existing tests pass (no regressions) — 281 tests in the package, 20 packages / full `yarn test:unit` green
- [x] Coverage headroom restored: lines 93.45 vs 93 required (was 93.07)
- [x] Full CI green on the first commit (45 pass, 4 skipping, including the e2e suite)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ logs
 .DS_Store
 
 midnight-concurrent-test-db
+test-concurrency-db
 .claude
 .worktrees
 .gitnexus
--- a/packages/level-private-state-provider/src/level-private-state-provider.ts
+++ b/packages/level-private-state-provider/src/level-private-state-provider.ts
@@ -131,6 +131,12 @@ export type DatabaseLevel = AbstractLevel<string | Buffer | Uint8Array, string, 
 
 export type LevelFactory = (dbName: string) => DatabaseLevel;
 
+/**
+ * A store as this provider uses it: string keys, string values. Keys that are
+ * branded strings, such as `ContractAddress`, widen to `string` on the way in.
+ */
+type StringSubLevel = AbstractSublevel<DatabaseLevel, string | Uint8Array | Buffer, string, string>;
+
 interface StorageContext {
   readonly dbName: string;
   readonly createLevel: LevelFactory;
@@ -186,6 +192,37 @@ const defaultLevelFactory: LevelFactory = (dbName: string) =>
 const dbAccessQueues = new Map<string, Promise<void>>();
 
 /**
+ * How long one queued operation may run before it is abandoned.
+ *
+ * Without a bound, a single operation that never settles - a `levelFactory`
+ * whose `open()` hangs, a stalled iterator - would wedge every later operation
+ * on that database for the lifetime of the process, with no error to diagnose.
+ */
+const DB_OPERATION_TIMEOUT_MS = 300000; // 5 minutes
+
+const withOperationTimeout = async <A>(dbName: string, operation: () => Promise<A>): Promise<A> => {
+  let timer: ReturnType<typeof setTimeout> | undefined;
+  try {
+    return await Promise.race([
+      operation(),
+      new Promise<never>((_resolve, reject) => {
+        timer = setTimeout(() => {
+          reject(
+            new Error(
+              `Timed out after ${DB_OPERATION_TIMEOUT_MS}ms operating on private state database "${dbName}". ` +
+              `The database handle appears stuck. Operations queued behind it have been released, so the ` +
+              `next one may fail to open until that handle is released.`
+            )
+          );
+        }, DB_OPERATION_TIMEOUT_MS);
+      })
+    ]);
+  } finally {
+    clearTimeout(timer);
+  }
+};
+
+/**
  * Serializes access to one database, so that overlapping callers queue instead
  * of colliding.
  *
@@ -196,6 +233,13 @@ const dbAccessQueues = new Map<string, Promise<void>>();
  * path (`browser-level`, IndexedDB) takes no such lock, but serializing there
  * is still what makes a read-modify-write sequence deterministic. The cost is
  * that throughput is one operation per database at a time.
+ *
+ * The queue slot is claimed synchronously, before this function's first `await`.
+ * That is what makes queue order equal call order, so `set(k, v)` followed by
+ * `remove(k)` applies in that order even when neither is awaited. Callers must
+ * therefore reach this function without awaiting anything first - work that
+ * needs the database, such as resolving an encryption key from the stored salt,
+ * belongs inside `operation`, which runs on the already-open handle.
  *
  * The map is module-level because the resource it guards is process-wide: a
  * per-provider map could not see a second provider opening the same database.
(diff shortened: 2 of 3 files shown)


sp-io:
> *This was generated by AI during triage.*

Triage measurement from #1232, relevant to the "Why not a shared open handle" section.

#1232 reports this bug without concurrency and asks for the per-provider handle this PR declines. Two results from triaging it, both against `main` (`5.0.0-beta.7`):

- **The sequential claim does not hold.** 40 awaited call-shaped sequences on one provider (about 300 open/close cycles), 25 fresh processes with a new provider per call, and 300 raw `level@10` open/use/close cycles all completed with zero failures. An awaited `level.close()` releases the lock before the next `open()`. So the queue in this PR is sufficient, and #1232's "a mutex would not be sufficient" does not follow. (Measured on macOS, not Linux.)

- **The shared-handle option may be cheaper than the section assumes.** Memoizing one `Level` per `dbName` through the existing `levelFactory` seam, while keeping the current per-operation `close()`, survived 80 concurrent operations across 3 runs with zero rejections. That shape is not an always-open refcounted handle: closes still happen per operation, so it does not hold the lock for the process lifetime and needs no new `close()` on the public contract. With one handle per path there is no second handle to collide with, and `abstract-level`'s own state machine absorbs the overlap.

Not an argument to change this PR -- the queue also fixes it, and it additionally covers `migrateSublevel` and the always-run `level.close()` defect. Recorded so the follow-up decision has the numbers.

PR #1168, opened 2026-08-19
feat(midnight-js): add ledger-8 circuit execution and keep-state wrap behind the loadLedger8Engine facade

sp-io:
# Overview

**PR 1004-E1 of the MJS-01 series (#1004)** — circuit execution on the retained 0.16 stack, the keep-state v9 wrap, and the engine's public facade. Second of three PRs carved out of the original 1004-E (the three originally summed to a byte-identical tree with the fully reviewed original; review follow-up `f002525a` since amends the facade to pass `Ledger8InstanceMismatchError` through unwrapped instead of re-wrapping it as `Ledger8RuntimeMissingError`).

## Summary

- **Public facade — the engine's ONLY surface (owner-approved design):** `loadLedger8Engine(): Promise<Ledger8Engine>` (`src/lib/engine/load-engine.ts`), exported from the root barrel. Lazy end-to-end: engine internals and the retained-era WASM load via dynamic imports only — the root bundle keeps zero static WASM linkage (dist-laziness gate extended and asserting it). Construction acquires the 0.16 glue and onchain-runtime-v3, runs the single-axis instance guard once, memoizes with memo-clear-on-failure; v8-ledger acquisition joins construction with the composition legs in 1004-E2. Methods in THIS PR: `extractState`, `downConvertForExecution`, `executeCircuit`, `wrapKeepStateCall` (all options-object based). The composition legs (`composeCallTx`, `composeDeployTx`, `executeConstructor`) land in 1004-E2 (#1165).
- **Circuit execution** (`src/lib/engine/execute.ts`): raw `createCircuitContext` sequence (no compact-js), transcript POJO; tested against the executable `counter-016` fixture and its golden transcript, which has since landed on this PR's base in #1186.
- **Keep-state v9 wrap** (`src/lib/engine/wrap-v9.ts`): resolves the circuit's REAL registered operation from the migrated v9 `ContractState` (a blank `ContractOperation` is rejected by this repo's own composition guard — caught in review); missing operation → `Ledger8ComposeFailedError`. End-to-end tested with a real non-empty transcript crossing ocrt3 → v9 `partitionTranscripts`. A review follow-up extracted the shared assembly into `assembleCallPrototype` (`src/lib/engine/assemble-call.ts`, a structural `CallAssemblyLedger` interface both ledger modules satisfy with inferred type parameters); `wrapKeepStateCall` is now a thin binding of it to ledger-v9.
- **Errors:** new registry code `MIDNIGHT_JS_P_LEDGER8_COMPOSE_FAILED` + `Ledger8ComposeFailedError` with public `stage` discriminant (`'wrap-call' | 'call-operation' | 'deploy-verifier-key'` — the last two are exercised by 1004-E2); the `utils` cross-repo code-union test synced.
- **Deps:** `compact-runtime-ledger8` (`npm:@midnight-ntwrk/compact-runtime@0.16.0`) as a protocol dependency — verified to resolve to real 0.16.0 despite root `resolutions` pinning the ident to 0.18.0-rc.1.
- **Gates:** ESLint ACL + dist-laziness extended for the glue alias, ocrt3 and the `./engine` subpath (canary-tested); per-glob 100% coverage entries.

## Series

Merge order: #1155 → #1156 → #1164 → **this** → #1165 (E2) → #1194 → #1198.

The fixture layer that was `1004-E0` is no longer pending in front of this PR: #1184 landed the hard-fork fixtures on `main` and #1186 landed the 0.16 counter fixture and its golden transcript on `feat/1004-engine-core`, so the two placeholder PRs for them (#1159, #1167) are closed. This branch is merged up to date with its base as of `b824be0a`, which took the base's five-misreporting-seams fix and its docblock correction.

## Test plan
- [x] TDD in the original development (RED/GREEN evidence per task, incl. mutation-grade operation-resolution assertion)
- [x] At this PR boundary, re-run after the `b824be0a` base sync: protocol 298/298 at 100/100/100/100 coverage, nothing skipped; utils 235/235 green; `yarn build`, `yarn lint`, `yarn typecheck:tests` all green (dist-laziness and the two `dist/` suites run inside the protocol suite)

## Submission Checklist
- [x] Self-reviewed the diff
- [ ] Reviewer requested
- [x] Update README.md file (if relevant) — n/a
- [x] No new todos introduced

## Links
- Part of #1004 (MJS-01). Base: #1164. Fixtures this suite executes against: #1184, #1186 (both merged).

## Follow-up: engine errors gated against the built bundles (`2bdf3371`)

Review found that the class-based error discrimination this PR relies on did not
survive the build. The engine is its own rollup entry, so everything it throws
crosses a bundle boundary before a caller sees it — and with the error module
inlined per bundle (fixed in #1155), the classes on either side of that boundary
were different objects. Consequences on the built artifact, both reproduced:

- An error raised inside the chunk was not `instanceof` the root package's class,
  so any consumer `catch (e) { if (e instanceof Ledger8ComposeFailedError) … }`
  fell through to generic handling.
- The facade's own pass-through — `f002525a`, added in this PR's review — was
  ineffective in `dist`: with a **real** duplicate install of
  `onchain-runtime-v3`, the guard fired correctly but the caller received
  `Ledger8RuntimeMissingError` / `MIDNIGHT_JS_P_LEDGER8_RUNTIME_MISSING` with
  `axis` lost, and the remediation text told them to reinstall rather than dedupe.
  The real diagnosis sat one level down in `.cause`.

Every existing engine suite resolves `src/` once, which makes that boundary
invisible — `engine-load-engine-chunk-failure.test.ts` asserts "does not
double-wrap" and passes while the built artifact double-wraps.

`dist-engine-errors.test.ts` therefore drives the BUILT package through its own
exports map: the facade resolves and binds the retained era, an error raised
inside the chunk stays discriminable by class *and* by code from the root, and
neither engine bundle declares its own error classes. Verified failing first —
against the pre-fix build it reported `expected Ledger8ComposeFailedError … to be
an instance of Ledger8ComposeFailedError`. `Ledger8ComposeFailedError` also joins
the code-based recognition rows.

After the fix, on the built artifact: the broken-`./v8` path surfaces a single
unwrapped `Ledger8RuntimeMissingError`, and a real dual install surfaces
`Ledger8InstanceMismatchError` with its code, its `axis` and the `yarn why`
remediation intact. Protocol suite: 187/187 at 100/100/100/100.

## Follow-up: review fixes to the call assembly and the engine guards (`824617df`)

A full review pass over this leg found four functional defects the suite passed
over, plus guards that read as protective without being so. All fixed here;
`feat/1004-engine-legs` (#1165) has taken the merge and inherits them.

**Key location.** `assembleCallPrototype` passed the bare circuit id as the call
prototype's `keyLocation`. Everywhere else in this repo that slot carries the
canonical, contract-qualified form built by `encodeContractKeyLocation` with the
hash of the deployed verifier key — `packages/contracts/src/utils/ledger-utils.ts:142` says why: *"circuit
names alone are ambiguous across contracts"*. A keep-state call therefore
assembled into an `Intent` (all this suite asserted) but `parseContractKeyLocation`
rejects a bare id, so `ZKConfigRegistry.resolveKeyLocation()` could not resolve
it and the DApp-connector path threw. The v8-native leg shares this function and
inherits the fix.

**Verifier key.** The only guard was on a missing operation, though `wrap-v9.ts`
promised a call would never be composed against *"a blank, unverifiable
operation"* — a state carrying a blank `ContractOperation` passed straight
through. (`engine-wrap-v9-partition-failure.test.ts` was itself registering one.)
Ports the guard #1165 already carries, on stage `'call-verifier-key'`, so the two
branches converge rather than conflict. It is also what the key location hashes.

**Zswap.** The real 0.16 `CircuitContext` tracks `currentZswapLocalState`;
`Ledger8CircuitContext` declared neither it nor a way to return it, so a circuit
moving coins had its inputs and outputs dropped silently and composed into an
unbalanced transaction, rejected on submission with nothing in the engine having
reported a problem. Such a call is now refused at execution with
`Ledger8ZswapUnsupportedError`. Full plumb-through is an engine API change and
belongs in its own issue.

**Load diagnostics.** `Ledger8RuntimeMissingError` hardcoded the `/v8` subpath,
so every engine-chunk failure told the operator to investigate an entry point
that had loaded fine, with the real cause only visible under `.cause`. Verified
against the built package: a missing `compact-runtime-ledger8` reported the v8
text. The subpath is now a constructor argument and a public discriminant.

**Prototype-chain lookups.** `impureCircuits` and `ENVELOPE_DECODERS` are plain
object literals, so a bare index resolved `Object.prototype` members: a circuit
named `toString` dispatched into it and failed as an unrelated `TypeError`, and
`extractState(raw, 'toString')` returned the string `'[object Undefined]'` typed
as `EncodedStateValue`, with no throw at all. Both now look up own properties,
and the circuit guard names the circuits the contract does carry.

Also: `Ledger8ComposeFailedError` builds its message from a total `Record`
instead of an if-chain that silently reached the `wrap-call` text for an
unhandled stage; its messages no longer name `composeV8CallTx` /
`composeV8DeployTx`, which this package does not export; the engine option and
result types are re-exported from the root barrel so a consumer holding a
`Ledger8Engine` can name them without a second, gated import; and the stale
`dist/engine.mjs`/`.cjs`, "composition and deploy" and ledger-v8 drift-gate
claims in the TSDoc are corrected.

### Tests

Eight argument-position mutations survived the previous suite, which asserted
only that a prototype was constructed and that `Intent.addCall` accepted it. The
prototype capture now pins all ten constructor arguments, the pre-call state
bridge, the guaranteed/fallible order and fresh commitment randomness per call;
the execute fake pins the address, coin key and injected cost model. Verified by
re-running the mutants against the new suite — the previously surviving
pre/post-state, guaranteed/fallible and address/coinPk mutants are now killed.
The ledger-v8 rows the drift gate was already cited as covering are added, and
they compile.

Three `skipIf` gates on `dist/` artifacts are removed: under vitest the engine
facade resolves to `src/engine.ts` — which is what makes the chunk-failure mock
intercept at all — so those nine tests were skipping for a premise that does not
hold. The two suites that genuinely read `dist/` keep their gate and now fail
rather than skip in CI, matching `dist-error-identity.test.ts`.

Protocol suite: 242/242 at 100/100/100/100, nothing skipped.


VanessaPC (approved):

sp-io:
Synced from #1156 (the no-skip fix for the dist gates). While resolving, the three engine cases this branch added were routed through the same `readDistFile()` helper instead of raw `readFileSync` — otherwise half the suite would keep reading `dist/` without the fail-fast message.

Verified here: 298 passed, 0 skipped, coverage 100%.

PR #1175, opened 2026-08-20
ci: add .d.ts API-report gate for types and barrel

sp-io:
## Summary

Adds the `.d.ts` API-report gate that the ledger-fork type work depends on, **before** any breaking type change lands, so the breaking surface can be proven to equal exactly what is documented.

- `@microsoft/api-extractor` (pinned `7.58.12`) runs over the existing rollup-dts output for `midnight-js-types` and the `midnight-js` barrel; baselines are checked in and stay **frozen** during the breaking window.
- CI regenerates the report and diffs it against the baseline. Every changed block must be covered by an entry under `## Entries` in `build-tools/api-report/allowed-breaking.md`; anything undocumented fails the job.
- The allowlist parser is scoped to the `## Entries` section, so the format documentation's own bullets cannot act as live allow-entries.
- A changed block ends at a blank line. Without this, consecutive declarations folded into one block and a single entry could silently cover an undocumented neighbour — the gate's own unit test reproduces that case.
- `build-tools/api-report/allowed-breaking.md` states the gate's real scope: the `types` surface is compared; anything behind the barrel's namespace re-exports (`contracts`, `utils`, `network-id`) is **not**. A green gate is not evidence the whole public surface is unchanged.

Stacked on #1156. Two follow-up PRs build on this one and add their own allowlist entries.

## Test plan

- [x] Tests written first (TDD), verified they fail before fix — the block-boundary test fails (1 group instead of 2) before the fix, passes after
- [x] Gate proven to fire: undocumented export → fails; same export with a matching entry → passes; a `- ` bullet in the format prose does **not** allow it
- [x] `yarn api-report:test` 7/7 passing, now wired into the CI job
- [x] All existing tests pass (no regressions); build 16/16, lint clean
- [x] Gate reports baselines matching on this branch (no type surface changed here)


sp-io:
Closing: the gate is not worth its keep for this release line.

**Why it was dropped**

1. **It watches the class of break that is already self-reporting.** Every breaking change in this stack is type-visible — `submitTx(tx: FinalizedTransaction)` becomes `submitTx(tx: VersionedFinalizedTransaction)`, `proveTx`/`balanceTx` likewise, and #1177 adds two *required* members to `PublicDataProvider`. Each one is a compile error at the consumer's import site, and 5.0.0 is a major bump that licenses exactly that. The gate adds no signal a downstream `tsc` does not already produce.

2. **The repo already pins the fork surface harder than an API report can.** #1176 ships `expectTypeOf` checks that fail the build if a union arm is removed, a third arm added, or a discriminant widened, with `assertNever` in the default branch. That is a stronger guarantee than a documentation diff, because it breaks CI rather than a report.

3. **The one real leak on this line would have slipped through it.** Between the #1183 and #1193 doc regenerations, seven exported symbols disappeared from `midnight-js-protocol/compact-runtime` (`Bytes32Descriptor`, `CompactTypeJubjubSchnorrSignature`, `ContractAddressDescriptor`, `MaxUint8Descriptor`, `ShieldedCoinInfoDescriptor`, `ShieldedCoinRecipientDescriptor`, `ZswapCoinPublicKeyDescriptor`), carried in by the `build(deps)` compactc/compact-runtime bump in #1190 with no `!` marker. `PACKAGES` in the gate covers only `types` and the barrel, so `protocol` was never compared.

4. **A frozen baseline decays while the window stays open.** Bare-line entries such as `- readonly version: 'v9';` wave through any future block containing that line. The longer the breaking window runs, the less a green gate proves — and no anchor for closing that window was ever fixed.

**What replaces it**

- Require `!` on dependency bumps that move a re-exported surface — the gap that let #1190 through.
- At release time, not as a gate: `git diff --diff-filter=AD -- docs/api` across the 5.0.0 range yields added and removed symbols across *all* packages, feeding `docs/releases/v5.0.0/api-changes.md`.
- Keep pinning unions with `expectTypeOf`, as #1176 does.

**Stack impact**

#1176 has been restacked onto this PR's own base (`feat/1004-v8-subpath-loadv8`) with the six gate commits dropped; #1177 follows on top of it. The allowlist rationale written in both has been moved into their PR descriptions rather than deleted. Neither branch retains any `build-tools/api-report/` content; both build 16/16, lint clean, 27/27 test tasks pass.

PR #1176, opened 2026-08-20
feat(midnight-js)!: add LedgerVersion-discriminated type surface for the fork window

sp-io:
## Summary

The consumer-breaking read-surface and transaction-flow types for the ledger-fork window. Every version-divergent surface becomes a closed, `version`-discriminated union, so a caller cannot reach the wrong era's payload without defeating the type system.

**Read surface**
- `FinalizedTxData` gains `readonly version: 'v9'`; all 14 existing fields are unchanged.
- New `FinalizedTxDataV8` carries the same metadata with a v8 transaction (`import type` from `midnight-js-protocol/v8`, erased at runtime), and `VersionedFinalizedTxData` is the closed union of the two.

**Transaction flow**
- `V8TxBytes` (`{ version: 'v8', txBytes }`) plus `VersionedUnprovenTransaction` / `VersionedUnboundTransaction` / `VersionedFinalizedTransaction`; `proveTx`, `balanceTx` and `submitTx` carry them in both directions. Parameter lists are otherwise unchanged.
- The v8 arm crosses these seams as serialized tag-prefixed bytes because the two ledger runtimes are separate WASM instances — a live v8 object cannot safely cross a provider seam.
- Providers that only speak v9 reject the v8 arm with `V8PayloadUnsupportedError` (registered code `MIDNIGHT_JS_PR_V8_PAYLOAD_UNSUPPORTED`, transitional until the provider work lands); contracts-side narrowing of a v9-only path throws with the already-registered `MIDNIGHT_JS_C_ERA_INVARIANT_VIOLATION`. Every narrowing point is an exhaustive `switch` closed with `assertNever`.

Existing v9-native call sites keep compiling; only consumers of the read surfaces need the documented narrowing. Public examples (`llms.txt`, proof-provider READMEs and TSDoc) are updated to the new shapes.

Every entry of the breaking set is documented in `build-tools/api-report/allowed-breaking.md`; the gate reports 14 changed blocks, all covered.

Stacked on #1156. The API-report gate this PR originally sat on (#1175) was dropped; this branch now sits directly on `feat/1004-v8-subpath-loadv8`.

## Breaking surface

Rationale kept from the API-report allowlist that #1175 would have carried, so it survives into the 5.0.0 migration notes.

**Read surface.** `FinalizedTxData` gains a `version: 'v9'` discriminant, and a new `FinalizedTxDataV8` arm plus the closed `VersionedFinalizedTxData` union let read-surface providers report a v8-era finalized transaction record without silently mislabelling it as v9. Both arms are otherwise identical.

**Provider seams.** The three transaction-flow seams — `ProofProvider.proveTx`, `WalletProvider.balanceTx` and `MidnightProvider.submitTx` — now carry a version-tagged payload in both directions instead of a bare ledger transaction. During the ledger-fork window the two ledger runtimes are separate WASM instances, so a v8-era transaction cannot cross a provider seam as a live object; it crosses as serialized, tag-prefixed bytes (`V8TxBytes`), and the `version` discriminant says which runtime produced the payload. `VersionedTx` is the shared shape behind the three concrete unions.

**Rejection path.** Providers that only speak v9 reject the v8 arm with the new `V8PayloadUnsupportedError`, whose stable `code` comes from the shared error registry in `@midnight-ntwrk/midnight-js-utils`.

**Exported symbols added or changed:** `FinalizedTxDataV8`, `VersionedFinalizedTxData`, `V8TxBytes`, `VersionedTx`, `VersionedUnprovenTransaction`, `VersionedUnboundTransaction`, `VersionedFinalizedTransaction`, `V8PayloadUnsupportedError`, `ProviderErrorCode`, and the `version: 'v9'` field on `FinalizedTxData`.

## Test plan

- [x] Tests written first (TDD), verified they fail before implementation — 9 typecheck failures captured before the types existed
- [x] Compile tests are genuinely discriminating, not tautologies: independent hand-written fixtures compared with `toEqualTypeOf`, discriminant-literal pins, and `@ts-expect-error` proving a bare `Uint8Array` is not assignable at any seam. Each assertion was mutation-tested (retype a field / drop a union arm → the check fails)
- [x] Negative tests assert the specific error `code` via `hasErrorCode`, and that the downstream seam is never reached
- [x] All existing tests pass (no regressions): types 82, http-client 47, dapp-connector 14, contracts 226, utils 283
- [x] Lint clean, build 16/16


sp-io:
Superseded by #1204.

This PR was opened against the API-report gate branch (#1175). That gate has been dropped, but GitHub refuses to re-point the base of a PR that belongs to a stack entity (`Cannot change the base branch because the pull request is part of a stack`), in either direction. #1204 carries the same head branch `feat/1006-types-d14-foundation` with the correct base `feat/1004-v8-subpath-loadv8`.

Same commits, no code lost. This PR had no reviews and no human comments.

PR #1177, opened 2026-08-20
feat(midnight-js)!: add head-version and raw-state queries to PublicDataProvider

sp-io:
## Summary

Adds the two new required `PublicDataProvider` members together with their indexer implementation and the testkit mocks — the interface change is implementer-facing breaking, so the interface, every in-repo implementation and the mocks ship in one PR.

**Interface (`types`)**
- `queryLatestProtocolVersion(options?: { fresh?: boolean }): Promise<number>` — the network head's protocol version.
- `queryRawContractState(address, config?): Promise<RawContractState | null>` — serialized state before any parse, for either era.
- `RawContractState` carries `version`, `protocolVersion` and `raw`. `version` is derived from `protocolVersion` at a single construction point; the docs state plainly that agreement with the serialized envelope is **not** established at this layer.

**Indexer implementation**
- Head version reads `Block.protocolVersion` from the argument-less `block` root query.
- A monotonic era latch caches the head only once the newer era is **corroborated**, never on a bare integer: either the composed response carries a newer-era head together with a matching state envelope, or a self-decoded finalized record proves the era and the provider then takes the value from one real head read. A lagging replica does not latch, a pinned (offset) read does not latch, the latch never lowers, and `{ fresh: true }` always bypasses it.
- Corroboration is best-effort end to end: neither an unrecognized version integer nor a failing head fetch can break the finalization read it rides along with.
- `queryRawContractState` composes the head field and the state into one request. That buys one round trip, **not** one consistent snapshot — the indexer resolves root fields independently — and the code says so; the era cross-check remains the correctness backstop.
- The serialized tag is parsed **before** any decode, in the indexer's real hex wire encoding. Malformed input — missing second separator, unknown tag, tag/content mismatch, oversized prefix — fails with `MIDNIGHT_JS_U_TAG_PARSE_FAILED` and never reaches a deserializer. Error messages never echo payload bytes.
- A raw state arriving without a block to date it now fails fast instead of being reported as "no state at this address".

Fixtures are minted at test time through `protocol`'s sanctioned surfaces (`ledger` for v9, `loadLedger8()` for v8), so no `ledger-v8` dependency leaks outside `protocol`.

**`parseSerializedTag` returns here**

The base branch removed `parseSerializedTag`, `TagParseError` and `UTILS_ERROR_CODES` from `utils` (`7a6a9342`, `c5ffe986`) on the rule that a constant returns with the PR that first throws with it, and that the parser "comes back with its first real caller". This is that caller: `contractStateEnvelopeVersion` reads the envelope tag before any byte reaches a deserializer, so the module, its error code and its registry entry are restored in this PR. Both objections that motivated the removal are answered here:

- **Grammar.** The charset accepts the brackets and parentheses the ledger runtimes actually emit (`midnight:contract-state[v8]:`), not the invented `namespace:version:` shape the original pattern assumed.
- **`vN` is not the era.** No caller reads it as one. `codec.ts` matches whole tags through an explicit allowlist in which the **v8** ledger writes `contract-state[v6]` and the **v9** ledger writes `contract-state[v8]` — the mapping is pinned by a test that serializes a state with each runtime, so a runtime bump that changes the format version fails loudly instead of silently mis-reading an envelope.

## Stack and base sync

Stacked on #1156 → #1204 (D14 unions). The API-report gate this stack originally sat on (#1175) was dropped; #1176 was reopened as #1204 to get the correct base. Interface contract per the types plan; implementation per the providers plan.

Two merges bring the base in. Conflict resolutions are spelled out in each merge commit message.

`a1bbe87a` merges the 31 commits the base moved since this branch forked. The substantive resolution is the `parseSerializedTag` restore above. That merge also silently took the base's deletion of `UTILS_ERROR_CODES` while keeping the module that throws with it, so the code, its type, its registry entry and the `./serialized-tag` re-export were put back and the `utils` error-code test re-covers the group.

`4b0d2b6a` merges the 6 commits the base moved since. Nine files conflicted, two of them more than textually:

- **The transitional v9 arm is gone.** `mapping.ts` and `provider.ts` each carried a hardcoded `version: 'v9'` under a comment promising per-record dispatch "lands later". The base landed it: `requireV9Era` resolves the discriminant from the record's own `protocolVersion` and reports `EraUnsupportedError` / `EraUnresolvableError` rather than mislabelling a record it cannot decode. The base's arms win, and the indexer README's "Transaction Data" section with them — its "per-record era dispatch arrives later" claim is now false.
- **Two head-latch tests are restated, not deleted.** Both called `watchForTxData` with a record the v9-only read path now refuses — one v8-era, one carrying an unresolvable `protocolVersion`. Their subject was never the read succeeding; it was that a read which cannot establish the era must leave the version latch exactly as cold as it found it. Both now assert the typed rejection and then that same latch property: no corroboration request, and the next head reading still goes to the network. The second test's old premise — that an unresolvable version "must not turn the caller's finalization read into a failure" — is the opposite of what the base decided, so it is restated rather than preserved.

The remaining seven are additive: `types` re-exports both `./raw-contract-state` and `./unwrap-v9`; `public-data-provider.ts` imports `RawContractState` alongside `VersionedFinalizedTxData` and drops `FinalizedTxData`, which the base's retyped watch methods left unused; `UTILS_ERROR_CODES` and `UtilsErrorCode` rejoin the union under the base's corrected doc comment; and the registry test and both READMEs keep each side's additions.

One follow-on rename: the base moved compile-level tests to vitest's `*.test-d.ts` convention and dropped the hardcoded typecheck `include`, so a renamed or added file cannot silently stop being guarded. This branch's `public-data-provider.test.ts` is entirely compile-level, so it becomes `public-data-provider.test-d.ts`. Verified picked up — the typecheck pass reports it as 7 TS tests.

## Breaking surface

Rationale kept from the API-report allowlist that #1175 would have carried, so it survives into the 5.0.0 migration notes.

`PublicDataProvider` gains two members for reading across the ledger fork. `queryLatestProtocolVersion` reports the protocol version of the network's head block, so callers can tell which ledger era the network is on. `queryRawContractState` returns a contract's state as the serialized bytes the network actually sent, still in their envelope, paired with the era they belong to — the new `RawContractState` record, whose `version` field is typed by `LedgerVersion` from `@midnight-ntwrk/midnight-js-protocol`.

Both members are **required**, so every implementation of the interface — in this repo and downstream — must add them.

**Exported symbols added:** `queryLatestProtocolVersion`, `queryRawContractState`, `RawContractState`, and — back from the base's removal — `parseSerializedTag`, `TagParseError`, `UTILS_ERROR_CODES`.

## Test plan

- [x] Tests written first (TDD) per commit, verified failing before implementation
- [x] Latch behaviour tested by observable network-request counts, not private state: engagement per corroboration route, bare-int negative, lagging-replica negative, fresh bypass, never-lowers, fetch-failure tolerance
- [x] Adversarial parse matrix built from real minted bytes mutated at the hex level, asserting the typed code and that no deserializer was reached
- [x] Both envelopes round-trip byte-for-byte through the real runtimes
- [x] Post-merge full run, no regressions: 31/31 turbo test tasks, all force-run rather than replayed from cache — indexer 271 passed / 2 skipped, level-private-state 276, utils 275, contracts 238, testkit-js 126, protocol 118, http-client 49, node-zk-config 43, fetch-zk-config 35, compact 24, dapp-connector 15, midnight-js 9, logger 7, network-id 3, types 87 (including the three `*.test-d.ts` typecheck suites)
- [x] `yarn install --immutable` clean (both merged lockfiles needed no change), build 16/16, lint clean, `typecheck:tests` clean
- [ ] **Integration tests are written but have not been executed** — no Docker in the authoring environment. They assert the head version against the live indexer field and must be run before merge.


packages/indexer-public-data-provider/src/query-definitions.ts
```diff
@@ -275,6 +286,23 @@ export const CONTRACT_STATE_QUERY = gql(
   }`
 );
 
+// The call path's single request: the block whose protocol version dates the
+// state, and the state itself, in one document. Composition buys ONE ROUND
+// TRIP, not one consistent snapshot — the indexer resolves Query-root siblings
+// concurrently, from independent reads, so the two fields can still disagree.
+// With no offset the block field resolves to the head block.
+export const RAW_CONTRACT_STATE_QUERY = gql(
```
sp-io, line 294: **`protocolVersion` coverage audit across this file — 4 of 16 documents select it.**

I checked every document in `query-definitions.ts` against `schema.graphql`. Selecting it:

- `HEAD_PROTOCOL_VERSION_QUERY` (new)
- `RAW_CONTRACT_STATE_QUERY` (this one)
- `TX_ID_QUERY` (`Transaction.protocolVersion`)
- `DEPLOY_TX_QUERY` (all three inline fragments)

Not selecting it, **although the schema exposes it at that point**:

| Document | Where the era is reachable | Feeds an era-sensitive parse? |
|---|---|---|
| `CONTRACT_AND_ZSWAP_STATE_QUERY` | `block { ... }` is **already in the selection set** | Yes — `parseHexZswapState`, `parseHexContractState`, `parseHexLedgerParameters` (`provider.ts:412-414`) |
| `TXS_FROM_BLOCK_SUB` | `blocks { ... }` and `transactions { ... }` **already selected** | Yes — states extracted per block |
| `CONTRACT_STATE_QUERY` | not on `Contract`; needs `block(offset: $offset)` | Yes — `parseHexContractState` (`provider.ts:375`) |
| `DEPLOY_CONTRACT_STATE_TX_QUERY` | not on `ContractAction`; via `transaction { protocolVersion }` | Yes — `parseHexContractState` (`provider.ts:483`) |
| `CONTRACT_STATE_SUB` | not on `ContractAction`; via `transaction { protocolVersion }` | Yes — `parseHexContractState` (`provider.ts:489`) |
| `CONTRACT_EVENTS_QUERY` / `CONTRACT_EVENTS_SUB` | `ContractEvent.protocolVersion` exists **next to** the `version` already selected | Not in-provider (`raw` is carried verbatim), but the consumer that decodes `raw` gets no era |
| `BLOCK_QUERY` | `Block.protocolVersion` | No bytes returned |
| `LATEST_CONTRACT_TX_BLOCK_HEIGHT_QUERY` | `Transaction` / `Block` | No bytes returned |

Genuinely not applicable (no such field on the selected types): `UNSHIELDED_BALANCE_QUERY`, `QUERY_UNSHIELDED_BALANCES_WITH_OFFSET`, `UNSHIELDED_BALANCE_SUB`.

**Assessment.** Not a defect in this PR — the two new members are self-consistent, and `codec.ts` parsers are deliberately pinned to the v9 arm for now ("Transitional: always the v9 arm, whatever `protocolVersion` says" — `provider.ts:543`, `mapping.ts:103`), so today the field would have no consumer on those paths. What the PR does leave is an asymmetry: `queryRawContractState` reads the era from the network, while five older paths hand the same class of era-sensitive bytes to an era-pinned parser with no era in the response at all. Once the network carries both eras, those paths mis-read silently rather than failing.

Two of them are free to close: `CONTRACT_AND_ZSWAP_STATE_QUERY` and `TXS_FROM_BLOCK_SUB` already select the parent node, so it is one field each — no extra round trip, no change in request shape.

Follow-up PR stacked on this branch will widen these documents and give the new field a real consumer (fail fast when the reported era is one the pinned parsers cannot read), rather than selecting data nothing reads.

packages/indexer-public-data-provider/src/query-definitions.ts
```diff
@@ -275,6 +286,23 @@ export const CONTRACT_STATE_QUERY = gql(
...
+    block(offset: $offset) {
+      protocolVersion
+    }
+    contract(address: $address, offset: $offset) {
+      state
+    }
+  }`
+);
+
 export const CONTRACT_STATE_SUB = gql(
```
sp-io, line 306: `ContractAction` carries no `protocolVersion` in the schema, so this document cannot date its own bytes at all — the era is only reachable through `transaction { protocolVersion }`. Same shape as `DEPLOY_CONTRACT_STATE_TX_QUERY`, and both feed `parseHexContractState` (`provider.ts:483`, `provider.ts:489`), which is currently pinned to the v9 arm.

Noting it here rather than asking for a change in this PR; picked up in the stacked follow-up.

sp-io (commented):
Review of `protocolVersion` coverage in the indexer query documents (requested check). Findings inline; nothing here blocks this PR.

packages/indexer-public-data-provider/src/query-definitions.ts
sp-io: Follow-up is open: #1207, based on this branch.

It widens the six documents the schema allows and gives the field a consumer rather than leaving it as dead data: `parseHexContractState` now requires the dating protocol version and refuses to decode unless the reported era and the state envelope agree on an era this client can read. Side effect worth noting here - `queryContractState` and `queryZSwapAndContractState` now carry exactly the evidence this PR's head-version latch asks for, so they engage it; until now only `queryRawContractState` could, and nothing calls it yet.

`toRawContractState` is left exactly as it is here, agreement still being the caller's to establish.

PR #1180, opened 2026-08-20
feat(midnight-js)!: publish every package as ESM-only

### LINKED ISSUE #1173: [Task] Replace rollup with tsc and go ESM-only across all published packages
### Component

midnight-js

### Summary

Replace rollup with plain `tsc` as the build for every published package, and drop the CJS half of the output matrix (ESM-only).

Rollup is not bundling anything in this repo. `build-tools/rollup.config.factory.mjs:8-11` sets `external: [/node_modules/]`, so every runtime dependency stays external — verified on the artifacts: `packages/indexer-public-data-provider/dist/index.mjs` opens with 12 external imports (`@apollo/client`, `rxjs`, `graphql-ws`, …) and `packages/utils/dist/index.mjs` imports `@noble/hashes` from outside the bundle. No minification, no browser target (no package declares a `browser` field). Rollup's only two jobs here are concatenating a package's own source files into one file and rolling up its `.d.ts`.

We pay the full price of a bundler for concatenation nobody consumes.

`packages/compact` is the existence proof for the target: `"type": "module"`, `module`/`moduleResolution: NodeNext`, plain `tsc -p tsconfig.build.json`, `.js` extensions on relative imports, per-file ESM output. It builds in **0.85 s**.

### Why now

- `5.0.0` has not shipped GA (`5.0.0-beta.6`). Dropping CJS is a breaking change; doing it inside the existing beta line costs nothing, doing it after GA costs a major.
- `engines` is already `node >=22`. Node ≥ 22.12 supports `require(esm)` unflagged, so CJS consumers keep working through the interop rather than through a second build.
- ESM-only is what makes the `tsc` migration cheap. Dual-format `tsc` needs two tsconfigs, two output trees and `package.json` type shims per directory; single-format `tsc` needs one tsconfig and nothing else.

### Current state (measured on `main`, 8949082)

| Fact | Value |
|---|---|
| Cold full build (`turbo run build --force`) | 34.8 s wall / 80 s CPU for ~17k LOC |
| `packages/protocol` build | **9.8 s** for 164 LOC — 10 entries × 2 rollup passes = 20 passes |
| `packages/compact` build (`tsc`) | **0.85 s** |
| Rollup configs | 16, of which 3 (`protocol`, `midnight-js`, `testkit-js-e2e`) bypass the shared factory with a copied `entries.flatMap(...)` |
| `rollup` in `devDependencies` | 2 of 16 packages. `yarn workspace @midnight-ntwrk/midnight-js-contracts build` fails with `command not found: rollup`; builds only work via turbo from the root. Violates "Missing devDependencies" in CLAUDE.md |
| Dead plugins | `resolve()` and `commonjs()` in the factory can never fire — `external` pre-empts them |
| Output files per entry | 5 (`.mjs`, `.cjs`, `.d.mts`, `.d.cts`, `.d.ts`) |

Correctness cost, not hypothetical: `388ceee5` ("share one error module across protocol bundles"). Each rollup entry is bundled independently, so a relatively-imported `src/errors.ts` was inlined into every entry and each declared its own `class …Error`. A class only ever equals itself, so `instanceof` answered `false` across entries of the same package — silently, exactly where a caller distinguishes failure modes. The fix needed a bespoke `share-error-module` rollup plugin, `Symbol.hasInstance` on the error classes, and a dist-level identity test. Under per-file `tsc` output that class of bug cannot exist: module identity is preserved 1:1.

### Target

Per published package:

- `"type": "module"`
- `build: "tsc -p tsconfig.build.json"`
- `module`/`moduleResolution`: `NodeNext` (drop `bundler`)
- `exports` map keeps its current subpath keys, each pointing at `./dist/<name>.js` + `./dist/<name>.d.ts`
- drop `main`-vs-`module` duality, drop `.cjs` / `.d.cts` / `.d.mts` outputs
- `engines.node` → `>=22.12`

Subpath entries survive unchanged in count and laziness: `protocol` (10 subpaths, incl. the lazy ledger-era accessors) and `midnight-js` (5) each emit one file per source module, so lazy loading of the heavy WASM eras becomes easier to reason about than it is today, not harder.

### Scope

15 packages need work (`compact` is already there):

`contracts`, `dapp-connector-proof-provider`, `fetch-zk-config-provider`, `http-client-proof-provider`, `indexer-public-data-provider`, `level-private-state-provider`, `logger-provider`, `midnight-js`, `network-id`, `node-zk-config-provider`, `protocol`, `types`, `utils`, `testkit-js`, `testkit-js-e2e`

Suggested phases, each independently mergeable:

1. **Spike on one leaf package** (`network-id`, 5 specifiers, 64 LOC). Prove the tsconfig, the `exports` shape and the published-artifact resolution end to end. Record before/after build time.
2. **Import-extension codemod.** 491 extensionless relative specifiers across the 15 packages (`contracts` 186, `indexer-public-data-provider` 82, `testkit-js` 54, `utils` 41, `types` 36, `testkit-js-e2e` 29, `level-private-state-provider` 20, `protocol` 10, rest ≤ 8). **Zero of them are directory imports**, so this is a pure "append `.js`" rewrite with no `/index.js` special cases. Land with an ESLint rule (`import/extensions`) so it cannot regress.
3. **Per-package migration**, leaves first, following the dependency layering in CLAUDE.md: `types` → `network-id`/`utils` → providers → `contracts` → `midnight-js` barrel → testkit.
4. **Cleanup**: delete `build-tools/rollup.config.factory.mjs`, all 16 `rollup.config.mjs`, the `share-error-module` plugin and the `Symbol.hasInstance` workaround (keep the dist identity test as a regression gate), rollup deps from the root `package.json`, the `rollup` resolution pin, and the `.rollup.cache` entries in `clean` scripts.

### Acceptance criteria

- [ ] Every published package builds with `tsc` only; no rollup left in the repo (`grep -r rollup` clean outside lockfile history)
- [ ] `yarn workspace <pkg> build` works standalone for every package, not just via turbo from the root
- [ ] Cold `turbo run build --force` measurably faster; number recorded in the PR
- [ ] Published artifact resolves for an ESM consumer **and** for a `require()` consumer on Node ≥ 22.12 — verified against a packed tarball (`yarn pack`), not against the workspace
- [ ] `exports` subpath keys unchanged for `protocol` and `midnight-js`; no internal module reachable that was not reachable before
- [ ] Deep-import of an internal file is still blocked by the `exports` map
- [ ] Lazy ledger-era loading still lazy (existing `dist-laziness` gate passes)
- [ ] Error-class identity across subpaths still holds (existing dist identity gate passes) with the `share-error-module` plugin removed
- [ ] `yarn check`, full unit suite, integration and e2e green
- [ ] TS5069 warning in `testkit-js-e2e` gone
- [ ] `docs-api` (typedoc) still generates — it reads sources via `entryPointStrategy: packages`, so it should be unaffected; confirm rather than assume
- [ ] CHANGELOG entry flags the CJS drop as breaking, with the `require(esm)` migration note for consumers

### Risks and open questions

1. **`moduleResolution: bundler` → `NodeNext` is the real risk, not the codemod.** `bundler` is permissive about dependency `exports` maps; `NodeNext` is not. Needs a spike per risky dependency: `@apollo/client/*` subpaths (`/core`, `/link/http`, `/link/retry`, `/link/subscriptions`, `/utilities`), `isomorphic-ws`, `cross-fetch`, `graphql-ws`, and the WASM packages `@midnightntwrk/ledger-v9`, `@midnightntwrk/onchain-runtime-v4`, `@midnight-ntwrk/compact-runtime`, `@midnight-ntwrk/platform-js`.
2. **CJS interop shifts.** `packages/indexer-public-data-provider/src/config.ts:18-23` documents that `isomorphic-ws` (`module.exports = require('ws')`, no `exports` map) exposes only `default` to the ESM build and named imports work *only* in the CJS build. Removing the CJS build removes that escape hatch — the workaround must be verified as the permanent path.
3. **`__DEBUG__` loses its build-time substitution.** Only consumer is `packages/contracts/src/submit-tx.ts` (declared at :28, used at :60 and :87), fed today by `@rollup/plugin-replace` off `process.env.CI`. Needs a decision: runtime flag, or keep a minimal transform. Note the current mechanism ships *different code* to npm depending on the CI env var — worth revisiting on its own merits.
4. **Breaking change for CJS consumers.** Anyone on Node < 22.12, or on a bundler that will not follow `require(esm)`, breaks. Needs a product call — sign-off required before phase 3. `docs/releases/v3.0.0/new-features.md:325` still shows a `require()` example and would need updating.
5. **`importHelpers: true` in `tsconfig.base.json:7` with no `tslib` dependency anywhere.** A no-op at `target: es2024`, but it should be removed rather than left as a trap.
6. **The in-flight `.d.ts` API-report gate** (on `feat/1006-types-d14-api-report-gate`) reads a rolled-up single `.d.ts`. Per-file declarations change its input shape; the gate needs reworking in the same series.

### Sequencing

Must land **after** the HF v8→v9 MJS-01 stack (#1004 and PRs #1155 / #1156 / #1159 / #1164). A 491-specifier codemod plus 15 `package.json` rewrites would conflict with everything currently in flight, and `protocol` is the package that stack is actively reshaping.

### Non-goals

- Introducing a different bundler (tsdown, tsup, unbuild). Evaluated; rejected for now. Nothing here needs bundling, and `packages/compact` already proves `tsc` is sufficient. Revisit only if a browser-targeted, dependency-inlining artifact becomes a requirement.
- Changing any public API surface. This is a packaging and toolchain change only.

sp-io:
## Summary

Closes #1173. Moves **every published package** to ESM-only and deletes the dual-format build. Packaging only — no public API changes.

- Every published package is `"type": "module"` and ships one `dist/<name>.js` + `.js.map` + `.d.ts` per entry. `main`, `module`, `.cjs`, `.d.cts` and `.d.mts` are gone.
- `exports` **subpath keys are unchanged**, so no consumer import specifier moves. Each value is now `{ "types": ..., "default": ... }`.
- `createRollupConfig` is rewritten into a single factory. There is no dual-format function left alongside it.
- `rollup` is added to each package's devDependencies.
- `packages/compact` is deliberately untouched — see **Decision needed** below.
- Two packages that were publishing their own source and test fixtures to npm are now limited to `dist/` — see **Publishing hygiene** below.
- The e2e CJS/ESM loader probe is deleted — see **Obsolete test** below.
- `testkit-js-e2e` follows too, so **no dual-format producer is left in the repo** — see **The last dual-format package** below.

### Why one PR and not a per-package rollout

The CJS bundles `require()` their siblings:

```
contracts/dist/index.cjs                    -> network-id, types, utils
midnight-js/dist/index.cjs                  -> contracts, network-id, types, utils
indexer-public-data-provider/dist/index.cjs -> types, utils
```

So flipping one package leaks the break sideways: the moment `network-id` is ESM-only, the published CJS artifact of `contracts` is `require()`ing an ESM-only package while `contracts` itself announced nothing. Moving together announces the break exactly once and leaves no intermediate state.

### The factory

`build-tools/rollup.config.factory.mjs` exports exactly one function.

- **Entries are derived from the `exports` map** instead of a hand-written array. The published surface is the single source of truth. This matters most for `protocol`, whose subpath keys deliberately differ from its file names (`./platform-js/effect/Configuration` → `dist/platform-effect-configuration.js`); there is no longer a parallel list that can drift. A malformed entry throws rather than silently dropping a subpath.
- **`resolve()` and `commonjs()` dropped.** Nothing is bundled from node_modules, so they had nothing to do.
- **`external` is now a syntactic rule**: a specifier that is not relative or absolute is a runtime import. This removed the need for `protocol`'s bespoke config, whose only difference was an extra scope in `external`.
- **`replace()` is opt-in** via a `define` option. `__DEBUG__` exists only in `contracts/src/submit-tx.ts`, so only that package asks for it.
- **`declaration`/`declarationMap` off** for the source pass; declarations come from `rollup-plugin-dts`. This removes the stray per-module `.d.ts` and `.d.ts.map` files the dual build left in `dist` and never exported (`utils/dist` had 16 of them).

12 of the 14 package `rollup.config.mjs` files are byte-unchanged — only their meaning changed, via the factory.

> **Note for a separate decision:** `define` publishes **different code to npm depending on `process.env.CI`** (`__DEBUG__: process.env.CI !== 'true'`). This PR preserves that behaviour exactly and does not change it, but it deserves its own look.

### The `@/` alias removal (separate commit)

`testkit-js` resolved its internal modules through a `@/* -> ./src/*` tsconfig alias. That blocks the `external` rule above, because `@/utils` looks bare while being internal. Removing it also **fixed a live bug**: `rollup-plugin-dts` does not honour tsconfig `paths`, so the published `testkit-js/dist/index.d.ts` shipped `import ... from '@/test-environment'` — unresolvable for any consumer. Those declarations were already broken.

## Breaking change

Consumers need **both**:

| Requirement | Why |
|---|---|
| **Node >= 22.12** | Where `require(esm)` works unflagged. `engines.node` is now `>=22.12`. |
| **TypeScript >= 5.8**, `module` `node20`/`nodenext` or `moduleResolution: bundler` | Older module settings refuse the import at compile time. |

Measured against tsc 6.0.3, CommonJS consumer importing `@midnight-ntwrk/midnight-js/utils` and `@midnight-ntwrk/midnight-js-protocol/ledger`:

| `module` | `moduleResolution` | Result |
|---|---|---|
| `node16` | `node16` | `TS1479` |
| `node18` | (default) | `TS1479` |
| `node20` | (default) | compiles |
| `nodenext` | `nodenext` | compiles |
| `preserve` | `bundler` | compiles |
| `commonjs` | `node10` | `TS5107` (`node10` unsupported in TS 6) |

`node20` is the lowest CommonJS setting that works — 5.8 added it to model `require(esm)`. Documented in `docs/releases/v5.0.0/{breaking-changes,migration-guide}.md`.

## Test plan

- [x] Tests written first (TDD): the protocol export-surface gate lands in its own commit **before** any flip, and was mutation-tested — retargeting one subpath fails test 1, deleting a subpath fails test 2.
- [x] All existing tests pass (no regressions): `yarn test:unit` 20/20 tasks, forced no-cache; `testkit-js` 77/77.
- [x] Lint clean, build succeeds.

### Repo gates (all exit 0)

```
yarn clean && yarn turbo run build --force   16/16 tasks, 0 rollup warnings
yarn turbo run test --force (packages)       20/20 tasks
testkit-js unit suite                        77/77 tests
yarn workspace @midnight-ntwrk/testkit-js-e2e build   exit 0 (standalone)
yarn lint                                    clean
yarn typecheck:tests                         clean (now covers testkit-js/*)
yarn constraints                             clean
yarn commitlint --from 949b5a9d --to HEAD    0 problems
```

### Byte comparison against the dual-format build

Every emitted bundle was compared against the `.mjs` it replaces. **27 of 27 JS bundles byte-identical**, the only difference being the sourcemap file name (`index.mjs.map` → `index.js.map`), which must change. All declarations identical except `testkit-js`, which improved (see above).

### Artifact shape

Exactly 3 files per entry (`.js`, `.js.map`, `.d.ts`) with no `.cjs` / `.mjs` / `.d.cts` / `.d.mts` / `.d.ts.map` anywhere: `protocol` 30 files / 10 entries, `midnight-js` 15 / 5, the rest 3 / 1.

### Standalone builds

`yarn workspace <pkg> build` exits 0 for all 14 packages. Verified the `rollup` devDependency is load-bearing, not boilerplate: removing it from `logger-provider` and reinstalling gives `command not found: rollup`.

### Packed tarballs (the acceptance criterion)

All 13 packages packed with `yarn pack`, installed into a clean project from `file:` tarballs with resolutions forcing siblings to the new builds, then exercised on Node v22.13.0.

ESM (`import`):

```
PASS  protocol/ledger  LedgerParameters.initialParameters() -> LedgerParameters
PASS  protocol/platform-js/effect/Configuration  4 exports, e.g. Keys, Network, configProvider
PASS  protocol (bare)  5 exports
PASS  midnight-js/network-id  getNetworkId throws unconfigured, then round-trips
PASS  midnight-js/utils  assertIsHex accepts hex and rejects non-hex
PASS  contracts  submitTx + createContractMaintenanceTxInterface present
PASS  level-private-state-provider  classic-level encrypted set/get round-trip
PASS  deep import .../dist/ledger.js -> ERR_PACKAGE_PATH_NOT_EXPORTED
```

CJS (`require`, i.e. Node's `require(esm)` interop):

```
PASS  protocol/ledger  LedgerParameters.initialParameters() -> LedgerParameters
PASS  protocol/platform-js/effect/Configuration  4 exports
PASS  midnight-js/network-id  setNetworkId/getNetworkId round-trip
PASS  midnight-js/utils  assertIsHex accepts hex and rejects non-hex
PASS  contracts  submitTx present
PASS  deep require .../dist/ledger.js -> ERR_PACKAGE_PATH_NOT_EXPORTED
PASS  level-private-state-provider  classic-level encrypted set/get round-trip
```

`level-private-state-provider` wraps `classic-level`, a native CJS addon — the least-travelled path here. Tested for real under both loaders, not reasoned about.

`protocol` was exercised through two subpaths including `./platform-js/effect/Configuration`, whose key does not match its file name.

### Build time

No measurable change. Packages-only cold build, `--force`: **19.7s** before (main), **19.6-20.0s** after. Full cold build including `compactc` circuits: 3m12s before, 3m13s after. The rollup pass count per entry is unchanged; only the number of written outputs drops, and `tsc`/`dts` dominate.

## Correctness win: module-level state is no longer duplicated

Under the dual build, `network-id`'s `let currentNetworkId` and `level-private-state-provider`'s `encryptionInitPromises` / `passwordRotationLocks` / `encryptionCache` existed **twice** in a mixed-format process. The level maps are coordination primitives, so a duplicated lock map did not serialize access across the two copies. One format means one copy. (`utils` holds only immutable lookup sets; nothing else has module-level mutable state.)

## Decision needed: `packages/compact`

Left untouched. It is already `"type": "module"` but has **no `exports` map at all**, and its `bin` points at `dist/run-compactc.cjs`. Adding an `exports` map changes what consumers can reach, and the `.cjs` bin needs its own call — neither was decided, and both are public-surface changes, which this PR excludes. Happy to do it here if you want it in scope.

## Obsolete test: the CJS/ESM loader probe

`testkit-js-e2e/test/nodejs.it.test.ts` spawned `dist/counter.mjs` expecting exit 0 and `dist/counter.cjs` expecting exit **1**. It asserted a dual-format property — the same fixture succeeds as ESM and fails as CJS. With the framework ESM-only there is no second format to compare against, so the assertion no longer describes anything. Deleted, along with its only reason to exist:

- `src/node/counter.ts`, the spawned fixture. Nothing imports it; its `CounterConfiguration` was a third copy of a class `counter-api.ts` and `double-counter-api.ts` already export, and every other test imports those.
- The third rollup chunk that built it to `dist/counter.cjs` / `dist/counter.mjs`. That chunk ran without `declaration: false`, so it was also the source of the stray per-module `.d.ts` / `.d.ts.map` files in `dist` that the package never exported. Verified gone after a clean rebuild.
- The now-dead `dist/counter.*` line in `.npmignore`.

The e2e matrix is generated with `ls *.test.ts`, so CI drops the job on its own; no workflow lists the file.

What is genuinely lost: the only test that drove the testkit from a plain built Node script rather than from vitest. The deploy-and-increment path it covered is exercised by `contracts.it.test.ts` and friends.

## The last dual-format package

`testkit-js-e2e` is `private: true`, so it was initially left alone. Once the loader probe was gone, nothing exercised its CJS output at all, so it now uses the shared factory like everything else. Three defects went with it:

- **Missing `rollup` devDependency.** `yarn workspace @midnight-ntwrk/testkit-js-e2e build` failed with `command not found: rollup`; it only ever worked under turbo, which puts the root bin on PATH. Identical on `main`.
- **The repo's last rollup warning.** Its bespoke config set `declaration: false` without `declarationMap: false`, so every build printed TS5069. A forced clean build of all 16 tasks is now **warning-free**.
- **Stray per-module declarations** its old config emitted into `dist`.

Its `@/` alias was removed first, for the same reasons as `testkit-js`: it blocks the factory's external rule, and `rollup-plugin-dts` had been leaking unresolvable `from '@/types/events-types'` imports into `dist/events-api.d.ts` and `dist/block-time-api.d.ts`. 93 specifiers across 28 files.

With no dual-format config left anywhere, `@rollup/plugin-node-resolve` and `@rollup/plugin-commonjs` have no remaining reference in the repo and are dropped from the root devDependencies.

Verification: `main` / `module` fields — none anywhere; `.cjs` / `.d.cts` / `.d.mts` outputs — none anywhere; the 10 Compact-generated `src/contract/compiled/**/index.js` files were confirmed already-ESM before flipping `"type": "module"`.

## The testkit projects never type-checked

`tsc --noEmit -p testkit-js/testkit-js/tsconfig.json` reported **16** TS6059 errors and the e2e project **27**: `rootDir` was `./src` while `include` also covers `test/**` and the root-level `*.ts` config files. `rootDir` constrains emit, and these projects only type-check, so it moved to `tsconfig.build.json` next to the narrowed `include` the build actually uses.

TS6059 was not cosmetic — it stopped those files being checked at all. Fixing the config surfaced **four real errors**, present on `main` and never seen because `typecheck:tests` only looped over `packages/*/`. Confirmed pre-existing by reproducing them on `main` with `rootDir` removed:

| File | Error | Fix |
|---|---|---|
| `dapp-connector-wallet-adapter.ut.test.ts` | 2x `TS2352` casting mocks to `WalletFacade` / `UnshieldedKeystore` | Both are wallet-sdk classes with private members, so no literal satisfies them and a direct assertion is rejected. Routed through one named helper that states the unsoundness once instead of a cast per member. |
| `dapp-connector-proving.it.test.ts` | `TS2322` `NodeZkConfigProvider<string>` not assignable to `ZKConfigProvider<"noop">` | Keyed by `SimpleCircuit`, the contract's own circuit ids. |
| `utils/events-test-utils.ts` | `TS2345` `MidnightProviders<string, string, unknown>` vs `EventsProviders` | Annotated the binding so the generics come from the contract. |

`typecheck:tests` now covers `testkit-js/*`, so this cannot regress. It iterates over tsconfig files rather than directories (`testkit-js/env` has none) and passes `--composite false` — a composite `--noEmit` run writes `dist/tsconfig.tsbuildinfo`, which would have landed in the published `testkit-js` tarball.

## Publishing hygiene: two packages were leaking their own source to npm

Surfaced by the tarball test. `level-private-state-provider` and `node-zk-config-provider` were the only packages without a `files` field, so npm packed everything not ignored — `src/`, the whole `src/test/**` suite, `tsconfig*.json`, `tsconfig.tsbuildinfo`, `rollup.config.mjs`, `typedoc.json`, `vitest.config.ts` and stale `.turbo/*.log` build logs. Pre-existing on `main`.

`node-zk-config-provider` was also publishing its ZK test fixtures (`set_topic.prover`, `set_topic.verifier`, `set_topic.bzkir`), which accounted for essentially its whole tarball:

| Package | Before | After | Saved |
|---|---|---|---|
| `node-zk-config-provider` | 7 321 321 B | 5 685 B | **99%** |
| `level-private-state-provider` | 106 318 B | 24 083 B | **77%** |

Both now pack the same 5 entries as every other package. Verified nothing depended on those files being published:

- Unit tests still pass — `files` affects packing, not the repo: 276/276 and 43/43.
- Neither `dist` bundle contains `__dirname` or `import.meta.url`, so no shipped code can reach into the package's own tree. `node-zk-config-provider` resolves every artifact path from the caller-supplied directory.
- The `.prover` / `.verifier` / `.bzkir` strings in `dist` are file-extension constants, not fixture paths.
- Sourcemaps list `../src/*.ts` in `sources` with `sourcesContent` null — identical to the other twelve packages, so this makes the two consistent rather than removing a capability the others have.
- Acceptance suite re-run against the repacked tarballs: ESM and CJS both still green, including the `classic-level` round-trip.

## Found along the way, deliberately not fixed here

- **`949b5a9d` missed `network-id`** when it raised `engines.node` repo-wide, so `yarn constraints` fails at that commit. Fixed in the ESM commit rather than by rewriting the earlier one.

## Out of scope

`moduleResolution` `bundler` → `NodeNext`, and migrating the build from rollup to plain `tsc`. Separate risks, separate PRs.


testkit-js/testkit-js-e2e/test/unshielded.cross-wallet-transfer.it.test.ts
```diff
@@ -29,11 +29,11 @@ import {
 import path from 'path';
 import { afterAll, beforeAll, beforeEach, describe, expect, test } from 'vitest';
 
-import { CompiledUnshieldedContract } from '@/contract';
+import { CompiledUnshieldedContract } from '../src/contract';
 import {
   type UnshieldedContractCircuit,
   type UnshieldedContractProviders
-} from '@/types/unshielded-types';
+} from '../src/types/unshielded-types';
```
iangregsondev, line 36: There are a lot of these relative dirs going back to ..src/ what was wrong with the previous alias ?

iangregsondev (approved):

PR #1194, opened 2026-08-25
feat(midnight-js)!: dispatch the two ledger eras behind a protocol facade

sp-io:
# Overview

**PR 1004-G of the MJS-01 series (#1004)** - a ledger-era facade in `protocol`, so era dispatch happens once, in the only package that holds both ledger runtimes, instead of in every consumer above it.

Stacked on #1165 (`feat/1004-engine-legs`). Merge order: #1155 -> #1156 -> #1159 -> #1164 -> #1167 -> #1168 -> #1165 -> this.

## Why

MJS-01 delivered the engine as a v8-only facade (`loadLedger8Engine`), so every consumer that touches ledger bytes carries its own two-case switch: `contracts` (MJS-02) would own artifact-era, head-era and key-set dispatch, and the indexer provider already owns its own era decisions and latch.

The era-agnostic core was in fact already here, discovered during implementation rather than designed: `lib/engine/assemble-call.ts` declares a structural `CallAssemblyLedger` interface satisfied by **both** runtimes (`wrap-v9.ts` binds it to ledger-v9, `compose-v8.ts` to ledger-v8), and `extractEncodedStateValue(raw, version, ...)` already dispatches on a `LedgerVersion`. This PR names that abstraction and gives it a door.

This is **not** the unified entry API. `submitCallTx` / `deployContract` stay in `contracts` - they need provider interfaces from `types`, and `types` already depends on `protocol`, so moving them down would close a package cycle. The version-agnostic entry has to *be* the entry dApps already call.

## Summary

- **`loadLedgerEra(version): Promise<LedgerEra>`** (`src/lib/era/`), exported from the barrel. Four methods, both eras: `extractState`, `decodeContractState`, `composeCallTx`, `composeDeployTx`. Memoised per era, memo cleared on failure. Dispatch is a closed `switch` on `LedgerVersion` with a `never` default that also throws at runtime.
- **Only symmetric operations join the facade.** The fork-crossing operations stay on `Ledger8Engine`, which goes from 7 methods to 4: `downConvertForExecution`, `executeCircuit`, `executeConstructor`, `wrapKeepStateCall`. `executeCircuit` deliberately does not join - v9 execution runs through compact-js in `contracts`, so a symmetric signature there would be a lie in the type.
- **Bytes and POJOs cross the boundary, never a live WASM object.** Two runtimes are two WASM instances; class identity does not survive a bundle boundary (the reason `dist-engine-errors.test.ts` exists); and the version-tagged provider seams already made this call. Mechanised in the tests by `structuredClone` over every value the facade returns.
- **v9 composition arm brought down into `protocol`** (`era/compose-v9.ts`, `era/unshielded.ts`): call and deploy legs over ledger-v9, with the Zswap offer taken as an **input parameter** - offer construction stays in `contracts`, where the per-recipient encryption-key resolver lives.
- **Era-neutral error codes.** `Ledger8ComposeFailedError` -> `ComposeFailedError` (`MIDNIGHT_JS_P_COMPOSE_FAILED`), `Ledger8ComposeOptionError` -> `ComposeOptionError` (`MIDNIGHT_JS_P_COMPOSE_OPTION_INVALID`), plus the new `StateDecodeFailedError` (`MIDNIGHT_JS_P_STATE_DECODE_FAILED`). Each carries `version` as a field. Registry 11 -> 12 keys by this PR's own change, 13 in total now that #1168's inherited `UNKNOWN_LEDGER8_AXIS` has merged down. The option error had to be renamed too: `assertComposeEnvelope` is shared by both eras' legs, so its `Ledger8*` name would have been thrown out of a purely v9 operation.
- **`Ledger8*` names are kept on the fork-crossing layer** - there the era is the subject, not a parameter.
- **Parity table** (`era-parity.test.ts`): `it.each(['v8','v9'])` runs the same scenario through both arms. `extractState` is asserted **deep-equal** across the migrated / pre-migration golden pair; the rest is shape parity, because the two eras emit different serialization tags.
- **Gates extended:** `dist-laziness` asserts on the built package that `loadLedgerEra('v9')` links no `ledger-v8` or `onchain-runtime-v3` and that the v8 arm reaches its chunks only through the dynamic imports; `dist-engine-errors` asserts the new classes stay discriminable by class **and** by code across the chunk boundary.

`packages/contracts` is untouched: `utils/ledger-utils.ts` remains the production v9 path. Repointing it at the facade is an MJS-02 decision, behind the existing e2e suite.

## Decisions taken during implementation

- **`loadLedgerEra('v8')` acquires ledger-v8 up front**, so the facade methods stay synchronous. Asking for the v8 era by name *is* the observation of v8 that the lazy-WASM requirement gates on. The old `engine-load-engine-v8-laziness.test.ts` premise dies with the compose legs and is replaced by `era-load-era-v8-laziness.test.ts`, which asserts `loadLedgerEra('v9')` never calls `loadLedger8` and `loadLedgerEra('v8')` calls it exactly once.
- **No `./era` exports subpath.** The barrel already re-exports `ledger-v9` and `compact-js` statically - the facade's only static dependencies - so a separate entry saves a consumer no linkage and costs one more bundle. Follows the `loadLedger8Engine` precedent: `src/lib/` is not a build entry.

## Deviations worth a reviewer's attention

1. **`composeDeployTx` returns `{ transaction, contractAddress, initialState }`, not bytes.** `ContractDeploy` mints a fresh nonce, so the same state deploys to a different address on every construction - measured: three constructions, three addresses. The address cannot be recomputed from the inputs and must travel with the transaction. A test pins that two deploys of one state differ.
2. **The v8 legs' option shapes changed with them.** `ComposeV8CallOptions.contractState` was a live WASM handle and the call pre-state wrapped a live `ocrt3.ChargedState`; both are now bytes / `EncodedStateValue`, and `composeV8DeployTx` returns the same record as v9. Keeping the old shape would have meant dragging `onchain-runtime-v3` into the v8 era arm purely to rebuild a handle. Every assertion in both v8 suites survives; only their option builders changed. This is the highest-risk part of the diff.
3. **`ComposeCallOptions.transcript` is a two-shape union.** The v8 leg partitions inside `assembleCallPrototype`; the v9 production path receives transcripts already partitioned by compact-js. Neither subsumes the other, and both eras expose `partitionTranscripts`, so both arms accept both shapes. This is where the remaining asymmetry between the eras actually lives.
4. **`engine-load-engine-missing-runtime.test.ts` is deleted.** It existed only to pin the engine's per-call v8 acquisition failure, which no longer exists. Its assertion is re-homed into `era-load-era-failure.test.ts` and strengthened from class matching to object identity.

## Zswap on the retained era

Review (#discussion_r3902578726) established that the earlier refusal rested on a false premise, and it is gone. The spike (`shieldedtech/spike-dapp-hf`, island-3) composes shielded circuits on the retained era with real proofs, and ledger-v8 exposes the same `ZswapOffer` / `Transaction.fromParts*` API v9 does.

- `executeCircuit` no longer refuses a coin-moving circuit. `TranscriptPojo` carries `zswapLocalState` - the post-call Zswap local state, decoded through the injected 0.16 glue - which is what a caller turns into the transaction's segmented offer (`zswapStateToSegmentedOffer` in `contracts`, where the encryption-key resolver lives). `Ledger8ZswapUnsupportedError` and its code are removed.
- Both era arms read `guaranteedZswapOffer` / `fallibleZswapOffer` and carry the offer into the transaction; unreadable bytes raise `ComposeOptionError(version, 'zswapOffer')` with the decoder failure on `cause`, identically on both eras. `era-parity.test.ts` pins the symmetry.
- The v8 arm still composes exactly one call, now stated for the real reason: a cross-contract call is a ledger-9-only feature a pre-fork contract cannot emit, not a consequence of refusing coin effects.

Wiring the carried state through `contracts` (building the offer from a retained-era call) is not in this PR - the seam it needs is what this change adds.

## Test plan

- [x] Tests written first (TDD), RED recorded before every implementation. Where a suite passed on first write (the parity table), RED was forced by breaking the v8 arm's operation resolution and then reverted
- [x] Every new coverage glob proven to match by raising its floor to 101 and seeing the named failure
- [x] `protocol` 35 files / 482 tests, 0 skipped, coverage 100/100/100/100 — re-verified after the #1168 merge and the review fixes below
- [x] The three `dist-*` suites confirmed running rather than skipping (14 + 4 + 17 tests)
- [x] Dist gates verified by sabotage: a deliberate static `export * from './v8'` plus two uncoded throws, rebuilt, three assertions red; restored and rebuilt green
- [x] `yarn lint`, `yarn typecheck:tests`, `yarn build` (16/16), `yarn test:unit` (28/28) all clean
- [x] No regressions: every pre-existing v8 and v9 engine suite passes

## Submission Checklist

- [x] Tests are provided
- [x] Self-reviewed the diff
- [ ] Reviewer requested
- [x] Update README.md file - `packages/protocol/README.md` covers the facade, the boundary rule and where the two eras actually differ
- [x] No new todos introduced

## Inherited: review fixes from #1168 (`331cbd74`)

Took the merge of #1165 after it picked up #1168's latest review pass
(`fc8f22da`, `64a50686`) — diagnosis and docs on the fork-crossing engine seams
this facade sits on top of. Three conflicts, all in the error layer, because
this PR renames the compose errors while that pass adds one:

- **`errors.ts`** — kept both additions side by side: this PR's
  `StateDecodeFailedError` and the parent's new `UnknownLedger8AxisError`. The
  `UnknownLedgerVersionError` doc block kept this PR's rewrite (it now names
  `loadLedgerEra` as a second raise site, and says why the class carries no
  `version` field) rather than the parent's envelope-only wording, which this PR
  had already superseded.
- **`errors.test.ts`** and **`protocol-acl.test.ts`** — both import lists and the
  exported-surface assertion take both names.

Nothing in the pass touches the era facade itself: the axis guard is called with
the one axis `loadLedger8Engine` already passes, `downConvertForExecution`'s new
runtime-binding check is on the fork-crossing side that stays on
`Ledger8Engine`, and no era-arm signature moved. Verified after the merge:
protocol 33 files / 461 tests at 100/100/100/100, the three `dist-*` suites
still running (14 + 4 + 17) against a fresh build, lint, `typecheck:tests` and
`yarn check:core` (28/28) all clean.

## Review fixes on this PR

**[r3904100036](https://github.com/midnightntwrk/midnight-js/pull/1194#discussion_r3904100036) — `era/unshielded.ts` doc block** (`4024c110`). A contract-addressed claimed unshielded spend is a normal, expected case once cross-contract calls exist, and the doc block read as if it were an oddity. It now says plainly why skipping it is safe: a cross-contract transfer is never materialized as a UTXO — the ledger settles it against the callee's own balance update under the `real_unshielded_spends` superset check, and `UtxoOutput.owner` takes a user address in the first place. Comment-only.

**[r3906584082](https://github.com/midnightntwrk/midnight-js/pull/1194#discussion_r3906584082) — the recorded query context never reached the partitioner** (`f321ec93`). Confirmed, and wider than `comIndices`.

`resolvePartition` rebuilt the `QueryContext` from `preState` alone, which restores the **state and nothing else**. compact-js's own v9 leg carries three more pieces across the same boundary — `ContractExecutable.js`: `asLedgerQueryContext` sets `block` and `effects`, `partitionAllTranscripts` folds the commitment indices — so this leg was missing a step the vendor performs, not missing an optimisation.

- **`CallTranscriptSource`'s unpartitioned variant now carries a required `PartitionContext`**: `block` and `effects` from the **pre-call** context (the partitioner recomputes both from the program it replays, so post-call values would be counted twice) and `comIndices` from the **post-call** one (the runtime registers a received coin's commitment as the circuit produces it). Required, not optional: an optional field would let a producer silently reproduce the defect.
- **`executeCircuit`** reads them off the retained runtime's own context — `block`/`effects` before the circuit runs, because the 0.16 glue swaps `currentQueryContext` on every coin it registers. **`assembleCallPrototype`** writes them onto the era's `QueryContext` before partitioning, in the vendor's order.
- **Not v8-only**, as flagged: `wrapKeepStateCall` builds the same unpartitioned source, so the fix sits in the shared seam and both era arms get it.
- **New stage `'call-partition-context'`.** A context the era rejects is a different caller defect from state bytes it cannot decode, and it comes from a different producer. No new error code — the stage is a field on `ComposeFailedError`, and both the message table and the test's stage record are total `Record<ComposeStage, …>`, so the addition is compiler-enforced. The `'call-partition'` remediation now names the missing-commitment case as well, since it previously pointed only at hand-built op sequences.
- **The commitment index is deliberately NOT recomputed** from `zswapLocalState`. Measured: `insertCommitment` accepts `-1n` and `2^70` without complaint, so a re-derived index rule fails silently rather than loudly — and the spike's own version of that rule is off by N when a call emits both user- and contract-owned outputs (it skips user-owned before incrementing; the glue increments on every output). Carrying the runtime's own map is the only guard.

Tests, RED first in every case:

- `era-assemble-call-partition-context.test.ts` — effects carried into the composed call's transcript (observable end-to-end through `Intent.addCall` → `ContractCall.guaranteedTranscript`), commitment indices and block folded onto the context actually handed to the partitioner, and three well-typed-but-runtime-invalid context shapes reported as `'call-partition-context'`.
- Both halves of the bridge re-verified by **sabotage**: dropping the `effects` assignment reds 5 tests across three suites, dropping the fold reds 2.
- `era-parity.test.ts` asserts the bridge on **both** eras through the composed transaction; the pre/post sourcing split is pinned in both directions in `engine-execute.test.ts` (swapping either field reds it); the real-0.16 fixture asserts the context the glue itself built.
- Drift gate extended to `CallContext` and `Effects` across all three axes — the bridge now depends on their structural identity, and the WASM setters reject a mismatch at runtime with no compile error to warn first.

**Gap closed in `df6679d0`** — the end-to-end case is now covered, and the reported failure is reproduced in-repo rather than argued from vendor code.

`testkit-js/.../fixtures/hf/coin-receiver-016/` is twelve lines of Compact: `receiveShielded(coin)` then `pot.writeCoin(coin, kernel.self())` into a `QualifiedShieldedCoinInfo` cell. Qualifying that coin needs the index its commitment was recorded at, so a transcript from this circuit cannot be partitioned unless the recorded indices travel with the call — which is precisely the defect.

- **Compiled in this repo, not ported.** The `compact` toolchain manager offers **0.31.1**, which reports `language-version 0.23.0` / `runtime-version 0.16.0` — the same triple `counter-016` was built with. `compiler/contract-info.json` is committed **and asserted**, so a recompile with the repo's pinned `compactc` (0.34.x → async 0.19 codegen) fails loudly instead of yielding an artifact that never reaches the seam under test.
- **Not the spike's `micro-dao`.** Same receive path, but 239 KB and 11 circuits with witnesses, a Merkle tree, minting and voting phases that nothing here tests. This fixture is 26 KB and its source fits on one screen.
- **`era-partition-received-coin.test.ts`** runs the real 0.16 execution and asserts both directions on **both** eras: the call composes when the recorded commitments travel with it, and is refused with stage `'call-partition'` when they are dropped.

Remaining, unchanged: **`PreTranscript`'s third argument (`comm_comm`) is not wired.** Both eras declare `constructor(context, program, comm_comm?)`; the structural type here declares two, so an unpartitioned **C2C callee** is not composable. Masked today by the v8 arm's one-call limit, and adding the parameter with no producer would be dead code — documented as a limit instead.

## Links

- Part of #1004 (MJS-01). Depends on #1165.


VanessaPC (approved):

packages/protocol/src/lib/era/adapt-v8.ts
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * Raised as {@link ComposeOptionError} naming the offending option, not as
+ * `Ledger8ZswapUnsupportedError`: no circuit ran here and none produced
+ * anything, so the class whose message says a circuit produced Zswap effects
+ * would be describing something that did not happen. That class stays on the
+ * execution leg, where a circuit really did. Keeping them apart is what lets a
+ * caller tell "my circuit moves coins, it cannot run on v8" from "I passed
+ * offer bytes to a v8 composition".
+ */
+const refuseZswapOffer = (offers: readonly (Uint8Array | undefined)[]): void => {
```
kapke: This exists based on a false premise and restricts functionality that should not be ever restricted.
The spike shows how to downcast the state to allow for shielded tokens interactions: https://github.com/shieldedtech/spike-dapp-hf.

packages/protocol/src/lib/era/unshielded.ts
```diff
@@ -0,0 +1,138 @@
...
+  version: LedgerVersion,
+  circuitId: string
+): UtxoOutput[] => {
+  if (transcript === undefined) {
+    return [];
+  }
+
+  const outputs: UtxoOutput[] = [];
+  for (const [[tokenType, publicAddress], value] of transcript.effects.claimedUnshieldedSpends) {
+    if (publicAddress.tag !== 'user') {
```
kapke, line 101: But contract 2 contract calls render a possibility to have such situation.

sp-io: Before changing this I checked it against the ledger types and the spike. `PublicAddress` is a two-member union (`user | contract`), and the spike's assembler skips a contract recipient for the same reason this does — `assemble-token.ts`: "A spend to a CONTRACT recipient stays inside the contract (no tx output)". A contract-addressed claimed spend is settled by the ledger's `real_unshielded_spends ⊇ claimed_unshielded_spends` check against the callee's own balance update, and `UtxoOutput.owner` only takes a user address, so there is nothing this seam could emit for it.

So I want to be sure which part you are flagging:

1. the docblock reads as if a contract-addressed spend were an oddity, and should instead say plainly that contract-to-contract calls make it a normal, expected case;
2. the bare `continue` should become an exhaustive dispatch that fails CLOSED if the vendor union ever grows a third tag — the same argument the paragraph right below makes for `TokenType`;
3. something about C2C this seam gets outright wrong, in which case tell me what the expected output is.

Happy to do any of them — I would rather ask than guess. The Zswap thread on `adapt-v8.ts` I am fixing as you described: the v8 arm will read and carry the offer exactly like the v9 arm, and the execution leg will carry the post-call `ZswapLocalState` instead of refusing the call.

packages/protocol/src/lib/era/adapt-v8.ts
sp-io: You are right, and the premise was false in both places, so both are gone (6cc60bbc).

What I checked first: island-3 of the spike composes the four shielded circuits on the retained era with real 8.0.3 proofs (`shielded-hf.test.ts` -> `assembleShieldedTokenCallV8` -> `assembleDaoCallV8`), building the offer from the circuit's `currentZswapLocalState`. And `ledger-v8.d.ts` exposes `ZswapOffer.deserialize` and `Transaction.fromParts*(networkId, guaranteed, fallible, intent)` with the same signatures as v9 — the v8 leg was simply passing `undefined, undefined`.

Execution leg:
- `TranscriptPojo` now carries `zswapLocalState`, the post-call state decoded through the injected 0.16 glue (`decodeZswapLocalState`), so a caller feeds it straight into `zswapStateToSegmentedOffer` in `contracts` without acquiring the retained glue itself.
- `Ledger8ZswapUnsupportedError` and `MIDNIGHT_JS_P_LEDGER8_ZSWAP_UNSUPPORTED` are deleted; `executeCircuit` runs a coin-moving circuit like any other.

Composition legs:
- The v8 arm reads both offers exactly as the v9 arm does and carries them into the transaction. Bytes its own decoder rejects raise `ComposeOptionError('v8', 'zswapOffer')` with the decoder failure on `cause` — the same fault, the same remediation, on both eras. `era-parity.test.ts` now pins that symmetry in both directions (offer carried; unreadable bytes refused identically).
- The single-call refusal stays, but with its real reason: a cross-contract call is a ledger-9-only feature a pre-fork contract cannot emit (spike, `island-1/spec.md` §Out of scope), not a consequence of refusing coin effects. The old wording in `errors.ts` and the README said the latter.

Not in this PR: wiring the carried state through `contracts` so a retained-era call builds its own offer. This change adds the seam that needs.

packages/protocol/src/lib/era/unshielded.ts
kapke: I thought C2C works slightly differently, hence flagging. I think the best course for this one would be by correcting the doc block by saying that it's safe to do in this case, because cross-contract transfers won't be materialized as UTxOs.

packages/protocol/src/lib/engine/assemble-call.ts
```diff
@@ -85,100 +90,189 @@ export interface VerifiableOperation {
...
+    if (transcript.guaranteed === undefined && transcript.fallible === undefined) {
+      throw new ComposeFailedError(version, 'call-transcript-empty', circuitId);
+    }
+    return [transcript.guaranteed, transcript.fallible];
+  }
+
+  let queryContext: TQueryContext;
+  try {
+    const stateValue = ledger.StateValue.decode(transcript.preState);
+    queryContext = new ledger.QueryContext(new ledger.ChargedState(stateValue), contractAddress);
```
kapke, line 237: `comIndices` is never populated here, so a call that receives a coin in-contract
  can't be partitioned — and this is not v8-only.

  `resolvePartition` builds a fresh `QueryContext` from `preState` and partitions
  immediately. Nothing calls `insertCommitment` — it doesn't appear anywhere under
  `packages/`. Per the `onchain-runtime-v3` docblocks, that map is exactly what
  makes a received coin spendable later:

  - `insertCommitment(comm, index)` — "Register a given coin commitment as being
    accessible at a specific index, for use when receiving coins in-contract, and
    needing to record their index to later spend them"
  - `qualify(coin)` — "upgrades an encoded `ShieldedCoinInfo` to an encoded
    `QualifiedShieldedCoinInfo` using the inserted commitments"

  The spike hits this and works around it —
  `island-3/driver/src/contracts/micro-dao/assemble-dao.ts:279-291`:

  ```ts
  // comIndices (PR #648) fix: a `receiveShielded` coin op in the transcript reads
  // the received coin's commitment at a recorded index. partitionTranscripts fails
  // ("expected a cell, received null") unless that commitment is registered first.
  for (const out of zls.outputs) {
    if (out.recipient.is_left) continue;              // user-owned: skip
    const co = B.ZswapOutput.newContractOwned(shieldedCoin(out.coinInfo), GUARANTEED, contractAddr);
    queryContext = queryContext.insertCommitment(co.commitment, comIdx);
    comIdx += 1n;
  }
```

  Failure scenario. A pre-fork DAO-shaped buy_in (pays a shielded coin into
  the contract pot) now executes fine and returns a zswapLocalState carrying a
  contract-owned output. composeCallTx then throws
  ComposeFailedError('call-partition') — whose remediation currently points at a
  non-rehashed Merkle tree, so the diagnosis misroutes as well. That's 3 of the
  spike's own proven DAO flows (buy_in, set_topic, vote_commit).

  It's shape-specific rather than universal: token-transfers/assemble-token.ts:82-88
  partitions with no insertion at all, because its four shielded circuits produce
  user-owned outputs only. This bites receive-into-contract shapes.

File: packages/protocol/src/lib/era/unshielded.ts (hunk 2 above)
Comment (sp-io) [reply packages/protocol/src/lib/era/unshielded.ts:101]: Fixed as recommended

Comment (sp-io) [reply packages/protocol/src/lib/era/unshielded.ts:101]: Done in `4024c110` — the doc block now says it plainly:

> A contract-addressed claimed spend is skipped, and cross-contract calls make that a normal, expected case rather than an oddity. It is safe to skip because a cross-contract transfer is never materialized as a UTXO: the ledger settles it against the callee's own balance update, under the `real_unshielded_spends` superset check, and `UtxoOutput.owner` takes a user address in the first place. Emitting an output for one would add a payout the transaction cannot cover.

The paragraph below it also gained "not a **UTXO** payout at all", so the contrast with the refused user-addressed case stays sharp. Comment-only change.

File: packages/protocol/src/lib/engine/assemble-call.ts (hunk 3 above)
Comment (sp-io) [reply packages/protocol/src/lib/engine/assemble-call.ts:237]: Confirmed, and it turned out wider than `comIndices`. Fixed in `f321ec93`.

**The decisive find:** compact-js's own v9 leg already does the missing step — `ContractExecutable.js:65`, `partitionAllTranscripts`:

```js
new PreTranscript(
  Array.from(entry.finalQueryContext.comIndices)
    .reduce((qc, e) => qc.insertCommitment(...e), asLedgerQueryContext(entry.initialQueryContext)),
  entry.publicTranscript, entry.commCommData?.commComm)
```

and `asLedgerQueryContext` (:49-56), with the vendor's own comment *"The above method of converting to ledger query context only retains the state. So, we have to set the settable properties manually"*, also copies `block` and `effects`. Our `resolvePartition` did only the `StateValue.decode`. So this was a step the vendor performs and we skipped — and `comIndices` was one of three things dropped:

| Dropped | Effect |
|---|---|
| `comIndices` (post-call) | your finding — `qualify` has nothing to lift, partition fails |
| `block` (`balance`, `secondsSinceEpoch`, `parentBlockHash`, `caller`, `lastBlockTime`) | replay sees defaults, not what the circuit ran on |
| `effects` | measurable: the partitioner **starts** a transcript's effects from the context's own, so a composed call declared no claimed receives even when the execution recorded them |

**Also not v8-only, as you said** — `wrapKeepStateCall` builds the same `unpartitioned` source, so the fix sits in the shared seam (`assemble-call.ts` + `compose-types.ts`) and both era arms get it. Partitioning had to stay on a ledger module either way: `onchain-runtime-v3` exports no `partitionTranscripts`.

**What landed.** `CallTranscriptSource`'s unpartitioned variant now carries a required `PartitionContext` — `block`/`effects` from the **pre-call** context (the partitioner recomputes both from the program it replays, so post-call values would count twice) and `comIndices` from the **post-call** one. `executeCircuit` reads them off the retained runtime's own context; `assembleCallPrototype` writes them onto the era's `QueryContext` before partitioning, in the vendor's order. Required rather than optional, so no producer can silently reproduce the defect. A context the era rejects raises the new stage `'call-partition-context'` (a field on `ComposeFailedError`, so no new error code), and the `'call-partition'` remediation now names the missing-commitment case too — it pointed only at hand-built op sequences before.

**One thing I deliberately did NOT copy from the spike.** Re-deriving the commitment from `zswapLocalState` is a trap:

- `insertCommitment` accepts `-1n` and `2n ** 70n` without complaint (measured), so a wrong index fails **silently** — it surfaces later as `'call-partition'` or as a mismatched transcript, not at the point of the mistake.
- The spike's rule is off by N: `assemble-dao.ts` `continue`s on user-owned outputs **before** `comIdx += 1n`, while the glue increments `currentIndex` on *every* output (`compact-runtime-ledger8/dist/zswap.js:114,117`, single insert site, `currentIndex` starts `0n` and nothing else moves it — same in 0.18 at `:122`). A call that both pays a user and funds the contract pot would shift.

So we carry the runtime's own map instead of recomputing it.

**Tests**, RED first throughout, both halves of the bridge re-verified by sabotage (dropping the `effects` assignment reds 5 tests across three suites; dropping the fold reds 2). Parity asserted on **both** eras through the composed transaction; the pre/post sourcing split pinned in both directions; drift gate extended to `CallContext`/`Effects` on all three axes.

**Two limits worth your eyes, both stated in the PR body rather than papered over:**

1. **End-to-end receive-into-contract stays uncovered.** The HF fixtures carry no shielded circuit (`counter`/`twin-contract` only), so a real RED for the whole path needs a new `.compact` with `receive(...)` compiled by the era-0.16 toolchain. Happy to raise that as a fixture ticket.
2. **`PreTranscript`'s third argument (`comm_comm`) is not wired.** Both eras declare `constructor(context, program, comm_comm?)`; our structural type declares two parameters, so an unpartitioned **C2C callee** is not composable. Today that is masked by the v8 arm's one-call limit, and adding the parameter without a producer would be dead code — so it is documented as a limit instead. Say the word if you would rather it went in now.

Adjacent, same root cause, **unproven**: `executeCircuit` hands `createCircuitContext` a `ChargedState`, so `createInitialQueryContext` sets `balance = new Map()` (`circuit-context.js:42`) — a pre-fork circuit always sees a zero self-balance on our leg. The bridge closes the partition-side half of that; whether any real transcript trips over the execution-side half I could not test without a fixture.

sp-io: Follow-up on the coverage limit I flagged above — it is closed, and your reported failure is now **reproduced in this repo** rather than argued from compact-js's code. `df6679d0`.

I did not port the DAO. The `compact` toolchain manager offers **0.31.1**, which reports `language-version 0.23.0` / `runtime-version 0.16.0` — the same triple `counter-016` was built with — so the fixture is twelve lines compiled here:

```compact
export ledger pot: QualifiedShieldedCoinInfo;
export ledger pot_has_coin: Boolean;

export circuit receive_coin(coin: ShieldedCoinInfo): [] {
  receiveShielded(disclose(coin));
  pot.writeCoin(disclose(coin), right<ZswapCoinPublicKey, ContractAddress>(kernel.self()));
  pot_has_coin = true;
}
```

The `writeCoin` into a `QualifiedShieldedCoinInfo` cell is the whole point — same shape `buy_in` uses, without the witnesses, Merkle tree, minting and voting phases (239 KB, 11 circuits) that none of this tests.

Measured on the real 0.16 execution: 23 ops in the public transcript, one Zswap output, one recorded commitment at index `0n`. Then, on **both** eras:

| Recorded commitments | Result |
|---|---|
| carried | call composes |
| dropped | `ComposeFailedError`, stage `'call-partition'`, circuit `receive_coin` |

`compiler/contract-info.json` is committed and asserted (`runtime-version 0.16.0`, `compiler-version 0.31.1`), so a recompile with the repo's pinned `compactc` — 0.34.x, async 0.19 codegen — fails loudly instead of quietly producing an artifact that never reaches this seam.

One limit stays, unchanged: `PreTranscript`'s optional third argument (`comm_comm`) is still not wired, so an unpartitioned **C2C callee** is not composable. Masked today by the v8 arm's one-call limit; adding the parameter with no producer would be dead code. Say the word if you want it in now.

kapke (approved):

PR #1203, opened 2026-08-27
fix(types): replace unsafe Option.some cast with Option.fromNullable in asEffectOption

### LINKED ISSUE #1202: asEffectOption wraps null/undefined as Some instead of None, bypasses type safety with unsafe unknown-to-T cast
### Components

component:midnight-js

### Version affected

main branch, HEAD (also present in latest published release; file: packages/types/src/contract.ts:127-129)

### Context & setup

Discovered via static code audit of midnight-js repository.
No specific environment needed to reproduce — the bug is in the function logic itself.
Toolchain: TypeScript, Effect library (Option module).
Tested by reading source and writing a standalone reproduction script.

### Steps to reproduce

import { Option } from 'effect';

// Reproducing asEffectOption's behavior from packages/types/src/contract.ts:127-129
const asEffectOption = <T>(obj: unknown): Option.Option<T> => {
  return Option.some(obj) as Option.Option<T>;
};

// Step 1: Pass null
const result = asEffectOption<string>(null);
console.log('isSome(null):', Option.isSome(result));  // true
console.log('isNone(null):', Option.isNone(result));  // false

// Step 2: Pass undefined
const result2 = asEffectOption<string>(undefined);
console.log('isSome(undefined):', Option.isSome(result2));  // true

// Step 3: Pass wrong type
const result3 = asEffectOption<string>(42);
console.log('isSome(42 as string):', Option.isSome(result3));  // true, contains number not stringasEffectOption(null) returns Some(null) — passes Option.isSome() as true.
asEffectOption(undefined) returns Some(undefined) — passes Option.isSome() as true.
asEffectOption<string>(42) returns Some(42) typed as Option<string> — no runtime error, silent type violation.

The function unconditionally calls Option.some(obj) and casts to Option<T>.
Nullish values are never converted to None. The unknown input type is never validated.

Used in governance code at packages/contracts/src/governance/unproven-tx.ts:68
as asEffectOption(newAuthority). While the current call site is type-protected
at compile time (newAuthority is required SigningKey), any JavaScript consumer
or runtime nullish value would silently produce Some(null) that passes all
downstream Option.isSome() checks.

### Actual behaviour

asEffectOption(null) returns Some(null) — passes Option.isSome() as true.
asEffectOption(undefined) returns Some(undefined) — passes Option.isSome() as true.
asEffectOption<string>(42) returns Some(42) typed as Option<string> — no runtime error, silent type violation.

The function unconditionally calls Option.some(obj) and casts to Option<T>.
Nullish values are never converted to None. The unknown input type is never validated.

Used in governance code at packages/contracts/src/governance/unproven-tx.ts:68
as asEffectOption(newAuthority). While the current call site is type-protected
at compile time (newAuthority is required SigningKey), any JavaScript consumer
or runtime nullish value would silently produce Some(null) that passes all
downstream Option.isSome() checks.

### Expected behaviour

asEffectOption(null) should return None (Option.isNone() === true).
asEffectOption(undefined) should return None.
asEffectOption<string>(42) should either be rejected by the type signature
or at minimum not silently cast unknown to T.

The fix is to replace Option.some(obj) with Option.fromNullable(obj)
and narrow the input type from unknown to T | null | undefined:

export const asEffectOption = <T>(obj: T | null | undefined): Option.Option<T> => {
  return Option.fromNullable(obj);
};

This returns None for null/undefined and Some(value) for everything else,
matching the semantic contract of the Option type.

Reference: packages/types/src/contract.ts:127-129
Usage: packages/contracts/src/governance/unproven-tx.ts:68

Emmanuellsensai:
## Summary

Fixes `asEffectOption` to correctly return `None` for null/undefined
instead of wrapping them as `Some(null)` / `Some(undefined)`.

Fixes #1202

## Problem

`asEffectOption` in `packages/types/src/contract.ts:127-129` unconditionally
calls `Option.some(obj)` and casts `unknown` to `Option<T>`. This means
null/undefined pass `Option.isSome()` checks as true, which defeats the
purpose of the Option type.

## Solution

- Replace `Option.some(obj) as Option.Option<T>` with `Option.fromNullable(obj)`
- Narrow input type from `unknown` to `T | null | undefined`

`Option.fromNullable` returns `None` for null/undefined and `Some(value)`
for everything else, matching the semantic contract of the Option type.

## Testing

The existing governance call site (`packages/contracts/src/governance/unproven-tx.ts:68`)
passes `SigningKey` (non-nullable), so this change is backward-compatible
for current callers. Any caller that was passing null and relying on getting
`Some(null)` was already in a broken state.

### DIFF
--- a/packages/types/src/contract.ts
+++ b/packages/types/src/contract.ts
@@ -122,10 +122,10 @@ export const exitResultOrError: <A, E>(exit: Exit.Exit<A, E>) => A =
  * Wraps an object into an `Option.some`.
  *
  * @param obj The value that should be wrapped into an `Option`.
- * @returns An `Option.some` for `obj`.
+ * @returns An `Option.some` for `obj`, or `Option.none` if `obj` is null or undefined.
  */
-export const asEffectOption = <T>(obj: unknown): Option.Option<T> => {
-  return Option.some(obj) as Option.Option<T>;
+export const asEffectOption = <T>(obj: T | null | undefined): Option.Option<T> => {
+  return Option.fromNullable(obj);
 }
 
 /**


Emmanuellsensai:
Hi @iangregsondev Would appreciate a review when you get a chance. This fixes the asEffectOption bug reported in #1202.

sp-io:
Thanks for the fix and for the clear write-up on #1202. Sorry for the wait -- I'm working against a tight deadline on a large feature, so reviews are slower than they should be right now. I'll get to this as soon as I can. Thanks for your patience.

PR #1206, opened 2026-08-30
fix(level-private-state-provider): eliminate PBKDF2 re-derivation on cache hits

### LINKED ISSUE #1174: [Bug] level-private-state-provider: encryption cache re-derives PBKDF2 on every hit, so it saves nothing
### Component

midnight-js

### Version affected

`@midnight-ntwrk/midnight-js-level-private-state-provider@5.0.0-beta.6` (`main` @ 8949082a)

### Context & setup

macOS (Apple silicon), Node 22.13.0, `yarn test` in `packages/level-private-state-provider`.
Found while investigating why the unit test suite of this package takes ~47 s while every other package finishes in ~1 s.

### Steps to reproduce

1. Instrument `crypto.subtle.deriveBits` in a Vitest setup file to count calls and accumulate wall time.
2. Run `yarn vitest run` in `packages/level-private-state-provider`.
3. Read the counters.

Equivalent production repro: call `provider.get(...)` twice in a row on one provider instance with the same password and time each call.

### Actual behaviour

`getOrCreateEncryption` treats a warm cache exactly as expensively as a cold one.

`packages/level-private-state-provider/src/level-private-state-provider.ts:279`

```ts
const cached = encryptionCache.get(cacheKey);
if (cached && cached.saltHex === saltHex) {
  const password = await getPasswordFromProvider(passwordProvider);
  if (await cached.encryption.verifyPassword(password)) {   // <-- full PBKDF2, 600 000 iterations
    return cached.encryption;
  }
  ...
```

`StorageEncryption.verifyPassword` (`src/storage-encryption.ts:139`) derives the key from scratch with `PBKDF2_ITERATIONS_V2 = 600000` and compares it to the stored one. So a cache hit costs one full key derivation — the same as a cache miss. The cache eliminates zero PBKDF2 work, contradicting the comment that documents it (`src/level-private-state-provider.ts:226`):

> This cache avoids repeated PBKDF2 key derivation (600,000 iterations) on each operation.

Measured cost of one derivation via Web Crypto on this machine: **~71 ms**. Every `get`/`set`/`getSigningKey`/`setSigningKey` pays it, warm cache or not.

Measured in the unit suite (`--coverage.enabled=false`):

| | calls | time |
|---|---|---|
| `level-private-state-provider.test.ts` (188 tests) | 617 | 43.7 s |
| `storage-encryption.test.ts` | 39 | 2.7 s |
| **total PBKDF2** | **618** | **43.9 s of 45.5 s test time (96%)** |

LevelDB file I/O accounts for ~1.7 s; coverage and the html/junit reporters are noise (47 s with coverage vs 48 s without). The suite is not slow because of I/O or test count — it is slow because it pays the production KDF cost ~3.3 times per test.

Of those 618 derivations, **303 are cache hits, i.e. pure waste**. Patching only `verifyPassword` to a cheap comparison (probe, not a proposed fix) took the package from **47 s to 25 s** with no other change.

### Expected behaviour

A cache hit performs no key derivation. Verifying that the supplied password matches the cached instance must be cheap.

Suggested approach: alongside the cached `StorageEncryption`, store a keyed fingerprint of the password — `HMAC-SHA256(processKey, salt || password)` where `processKey` is random per process and never persisted. Compare it in constant time. The fingerprint is worthless offline, and the derived AES key is already resident in the same memory, so this widens nothing. Behaviour on a wrong password is unchanged: fingerprint mismatch falls through to the existing `StorageEncryption.create` path and fails on the GCM auth tag as it does today.

Impact if fixed:

- production: ~71 ms removed from every private-state read/write on a warm cache;
- CI and local: this package 47 s -> 25 s, whole-repo unit suite ~72 s -> ~50 s.

The remaining 25 s are genuine cache misses driven by test design (DBs wiped, passwords rotated, per-`accountId` level names). That is a separate, test-only lever — see the follow-up comment.

Ansh-Sonkusare:
## Summary

Fixes #1174.

`getOrCreateEncryption` called `StorageEncryption.verifyPassword()` on every warm-cache access. `verifyPassword` re-derives the PBKDF2 key at 600 000 iterations (~71 ms on Apple silicon) — the same cost as a cache miss. The cache eliminated zero KDF work, contradicting the comment at line 226 that documents it.

**Root cause:** the cache stored only the `StorageEncryption` instance and the salt hex; there was no cheap way to verify the incoming password matched the cached key, so `verifyPassword` ran a full derivation on every access.

**Fix:** store an `HMAC-SHA256(PROCESS_KEY, password:saltHex)` fingerprint alongside the cached `StorageEncryption`. Cache hits now do a constant-time fingerprint comparison (~microseconds) instead of a full PBKDF2 derivation.

Security properties:
- `PROCESS_KEY` is 32 random bytes generated once at module load, never persisted or logged.
- The fingerprint is worthless outside the current process.
- A wrong password (fingerprint mismatch) falls through to `StorageEncryption.create()`, which fails on the GCM auth tag exactly as before.
- The constant-time `timingSafeEqual` (already in this package) is used for the comparison.

Measured on Apple silicon, Node 22:

| | Before | After |
|---|---|---|
| Warm cache access | ~71 ms (full PBKDF2) | ~0 ms (HMAC compare) |
| `level-private-state-provider` unit suite | ~47 s | ~25 s |

The remaining ~25 s are genuine cold-cache derivations driven by test design (DBs wiped, passwords rotated, per-`accountId` level names) — addressed separately per the issue.

## Test plan

- [x] Existing unit suite passes (`packages/level-private-state-provider`) — 276/276 tests pass
- [x] Password-rotation tests still pass (wrong password → fingerprint mismatch → new derivation)
- [x] `invalidateEncryptionCache()` continues to work (cache cleared → cold path on next access)

### DIFF
--- a/packages/level-private-state-provider/src/level-private-state-provider.ts
+++ b/packages/level-private-state-provider/src/level-private-state-provider.ts
@@ -46,8 +46,10 @@ import {
   decryptValue,
   getPasswordFromProvider,
   type PrivateStoragePasswordProvider,
-  StorageEncryption
+  StorageEncryption,
+  timingSafeEqual
 } from './storage-encryption';
+import { hmac } from '@noble/hashes/hmac.js';
 
 /**
  * The default name of the indexedDB database for Midnight.
@@ -205,6 +207,14 @@ const METADATA_KEY = '__midnight_encryption_metadata__';
 
 const DEFAULT_MAX_ROTATION_ENTRIES = 10000;
 
+// Per-process key for password fingerprinting — random at startup, never persisted.
+// Cache hits are validated with HMAC-SHA256(PROCESS_KEY, password + ":" + saltHex)
+// instead of re-running PBKDF2 (600 000 iterations) on every access.
+const PROCESS_KEY = randomBytes(32);
+
+const computePasswordFingerprint = (password: string, saltHex: string): Uint8Array =>
+  hmac(sha256, PROCESS_KEY, new TextEncoder().encode(`${password}:${saltHex}`));
+
 const encryptionInitPromises = new Map<string, Promise<Buffer>>();
 
 const passwordRotationLocks = new Map<string, Promise<void>>();
@@ -220,6 +230,7 @@ export interface PasswordRotationOptions {
 interface EncryptionCacheEntry {
   readonly encryption: StorageEncryption;
   readonly saltHex: string;
+  readonly passwordFingerprint: Uint8Array;
 }
 
 /**
@@ -279,21 +290,16 @@ const getOrCreateEncryption = async (
   const cacheKey = `${ctx.dbName}:${levelName}`;
   const salt = await getOrCreateSalt(ctx, levelName);
   const saltHex = salt.toString('hex');
+  const password = await getPasswordFromProvider(passwordProvider);
+  const fingerprint = computePasswordFingerprint(password, saltHex);
 
   const cached = encryptionCache.get(cacheKey);
-  if (cached && cached.saltHex === saltHex) {
-    const password = await getPasswordFromProvider(passwordProvider);
-    if (await cached.encryption.verifyPassword(password)) {
-      return cached.encryption;
-    }
-    const encryption = await StorageEncryption.create(password, { existingSalt: salt, cryptoBackend: ctx.cryptoBackend });
-    encryptionCache.set(cacheKey, { encryption, saltHex });
-    return encryption;
+  if (cached && cached.saltHex === saltHex && timingSafeEqual(fingerprint, cached.passwordFingerprint)) {
+    return cached.encryption;
   }
 
-  const password = await getPasswordFromProvider(passwordProvider);
   const encryption = await StorageEncryption.create(password, { existingSalt: salt, cryptoBackend: ctx.cryptoBackend });
-  encryptionCache.set(cacheKey, { encryption, saltHex });
+  encryptionCache.set(cacheKey, { encryption, saltHex, passwordFingerprint: fingerprint });
   return encryption;
 };
 
--- a/packages/level-private-state-provider/src/test/level-private-state-provider.test.ts
+++ b/packages/level-private-state-provider/src/test/level-private-state-provider.test.ts
@@ -1772,6 +1772,7 @@ describe('Level Private State Provider', (): void => {
 
     test('encryption instance is reused for multiple operations (caching works)', async () => {
       const verifyPasswordSpy = vi.spyOn(StorageEncryption.prototype, 'verifyPassword');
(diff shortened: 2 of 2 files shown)


efekrbas:
Hi @Ansh-Sonkusare, excellent work! 👏

This is a very clean and impactful fix. Eliminating the 600k PBKDF2 re-derivation on warm cache hits resolves a major bottleneck in `level-private-state-provider` without compromising the security model.
---
### 🛡️ Security & Cryptographic Evaluation
- **Ephemeral Process Key (`PROCESS_KEY`):** Generating a 32-byte CSPRNG secret on module load ensures HMAC fingerprints are scoped strictly to the runtime process lifecycle and are meaningless outside of memory.
- **Timing Attack Resistance:** Using `timingSafeEqual` for fingerprint comparison ensures constant-time validation across warm lookups.
- **Delimiter / Domain Separation:** Because `saltHex` is fixed-length (64 hex characters) and hex-encoded, concatenation (`${password}:${saltHex}`) is safe from delimiter injection/collision even if `password` contains colons.
- **Graceful Fallback:** When a password rotation occurs, fingerprint mismatch naturally falls through to `StorageEncryption.create()`, correctly triggering key re-derivation and AES-GCM tag validation failure on stale ciphertexts.
---
### 💡 Suggestions & Minor Enhancements
1. **Strengthen Password-Rotation Test Assertion:**
   In `packages/level-private-state-provider/src/test/level-private-state-provider.test.ts` under `'cache is invalidated when password changes'`:
   It would be great to also assert `createSpy` count here (e.g. verifying `StorageEncryption.create` was invoked exactly 2 times across the password change), mirroring the approach added in the caching test:
   ```typescript
   const createSpy = vi.spyOn(StorageEncryption, 'create');
   // ... after password change and rejected get() ...
   expect(createSpy).toHaveBeenCalledTimes(2);
Potential Future Optimization (Concurrent Cold Misses): Currently, encryptionInitPromises only deduplicates the salt creation promise (getOrCreateSalt). If multiple asynchronous operations (e.g., Promise.all([db.get('a'), db.get('b')])) hit a cold cache simultaneously, both may trigger parallel StorageEncryption.create() calls before the cache is populated. Deduplicating the in-flight encryption derivation promise could save redundant PBKDF2 runs during cold startup races.
Overall, the logic is solid, conforms to repo conventions, and provides massive performance gains. 🚀

LGTM!

sp-io:
Thanks for this -- eliminating the PBKDF2 re-derivation on warm cache hits is a worthwhile fix, and the write-up made it easy to follow.

Apologies for not reviewing it yet: this touches the encrypted private-state path, so it needs a careful look, and a large feature under a tight deadline is currently taking all of my review time. I'll come back to it as soon as I can. Thanks for your patience.

PR #1207, opened 2026-08-31
fix(midnight-js)!: date indexer reads with the ledger era they belong to

sp-io:
## Summary

Follow-up to the `protocolVersion` coverage review on #1177 ([review](https://github.com/midnightntwrk/midnight-js/pull/1177#pullrequestreview-5064575104)). That audit found 4 of 16 indexer documents selecting `protocolVersion`, while five paths handed era-sensitive bytes to an era-pinned deserializer with no era in the response at all. This PR closes that gap and gives the field a real consumer.

**Queries widened**

| Document | Change | Cost |
|---|---|---|
| `CONTRACT_AND_ZSWAP_STATE_QUERY` | `block { protocolVersion }` | free - `block` already selected |
| `TXS_FROM_BLOCK_SUB` | `blocks { protocolVersion }` | free - `blocks` already selected |
| `CONTRACT_STATE_QUERY` | new `block(offset:)` root field | same round trip |
| `DEPLOY_CONTRACT_STATE_TX_QUERY` | `transaction { protocolVersion }` per arm | one sub-field |
| `CONTRACT_STATE_SUB` | `transaction { protocolVersion }` | one sub-field |
| `CONTRACT_EVENTS_QUERY` / `_SUB` | `protocolVersion` beside `version` | one field |

`BLOCK_QUERY` and `LATEST_CONTRACT_TX_BLOCK_HEIGHT_QUERY` are deliberately untouched: they return no serialized bytes, so the field would be dead data. `ContractAction` carries no era in the schema, which is why two documents reach it through `transaction`.

`protocol-version-coverage.test.ts` states the whole surface exhaustively and in both directions - a document that decodes era-sensitive bytes and fails to date them breaks the build, and so does a document that starts paying for a field it has no use for. The table also fails if a new document is added without being classified.

## What the era check is, and what it is not

It is **not** protection from a silent mis-decode. `ContractState.deserialize` checks the header tag before it reads the body, so genuine v8 bytes (tag `midnight:contract-state[v6]:`) are rejected by the v9 runtime with a named error. They can never be mis-decoded into a plausible-looking wrong answer. `era-cross-check.test.ts` pins that fact against real bytes minted by both runtimes, so the reasoning cannot rot into folklore and justify a stricter guard than the problem warrants.

What it **is**: a typed, actionable failure in place of a header-tag error from deep inside a decoder, era dating on contract events for consumers that decode `raw` themselves, and the evidence the head-version latch from #1177 needs.

## The rule: the envelope decides, the block bounds

The two signals are not symmetric, and the rule follows from how each is obtained.

- The **envelope** comes off the bytes. It is attacker-controlled and proves nothing about the network - but it is also what the deserializer reads, so a wrong envelope produces a failure, never a wrong answer. It therefore decides decodability: only v9 is decodable on this path.
- The **reported `protocolVersion`** dates the *read*, not the bytes. The indexer serves the latest contract action at or before the requested block, so a state is routinely older than the block that dates it; once the network has forked, every contract dormant across the fork is exactly that. It is an upper bound and nothing more.

So an older envelope under a newer block is the ordinary case, settled on the envelope alone. The reverse - an envelope *newer* than its dating block - cannot happen **while both signals describe the same block**, and that direction is reported as an era disagreement.

That qualifier is load-bearing. `block` and `contract` are Query-root fields the indexer resolves concurrently, from independent reads, and on an unpinned read both follow the chain tip - so a block indexed between the two leaves them on either side of a fork, giving post-fork bytes dated by a pre-fork block with nothing wrong anywhere. `parseHexContractState` therefore takes the bound as an explicit input: `enforced` (the default, so no call site loses the check by omission) where the version is read from the same resolution as the state or both root fields are pinned to one offset, and `withheld` on the unpinned sibling reads - `queryContractState` and `queryZSwapAndContractState` with no offset, and `watchForContractState`, whose poll is hardwired to an unpinned query. Withholding drops only the comparison; the envelope still decides decodability, which is what keeps a wrong-era payload away from the decoder either way.

An unresolvable protocol version withholds the bound and nothing else, rather than stranding a readable state behind an integer this client happens not to map. `toRawContractState` still fails closed on one, because `RawContractState.version` cannot be constructed without an era.

**The check pays for itself.** An unpinned state read now carries the evidence the head-version latch from #1177 asks for - a head reading plus a v9 envelope - so `queryContractState` and `queryZSwapAndContractState` engage that latch. Until now only `queryRawContractState` could, and nothing calls it yet, so in practice `queryLatestProtocolVersion` always paid for a request. `corroborateFromDecodedState` re-checks the head reading itself: a v9 envelope does not make the reported integer resolvable, and `corroborateV9` treats an unresolvable one as a call-site bug, so cache warming a read was only doing incidentally must not fail that read.

`toRawContractState` is otherwise untouched. `RawContractState` still reports what the network said, with agreement between the signals left to its caller, exactly as #1177 documents.

## Contract events

`ContractEventBase` gains a required `protocolVersion`. The provider never decodes `raw`, so the beneficiary is the consumer that does: it can now tell which runtime wrote those bytes. The interface satisfies `VersionedRecord` from `protocol`, so `versionOfRecord(event)` resolves the era with no local version table - pinned by a test.

## Error taxonomy

A state served with no block to date it now raises its own `undated-state` kind. It was reusing `missing-head-block`, whose message tells the caller to wait for the indexer to index a block - not the situation at hand. Applied to `queryRawContractState` too, where that reuse came from.

## Stack

Stacked on #1177 (`feat/1006-types-d14-provider-members`). Merge that first.

## Breaking surface

- `ContractEventBase` gains a required `protocolVersion` member, so every producer of `ContractEvent` must supply it.
- The exported `parseHexContractState` helper takes a second `protocolVersion` argument. Required rather than optional on purpose: the era the read is dated to is part of the question being asked.
- `queryContractState`, `queryDeployContractState`, `queryZSwapAndContractState`, `watchForContractState` and `contractStateObservable` reject an undecodable era with a typed `IndexerDataError` instead of the decoder's header-tag error. Those refusals are precisely the ones the tag-strict decoder already rejected, so nothing that decodes today stops decoding.
- The only reads that start failing are **pinned** reads where the indexer serves genuinely inconsistent data for one fixed block: bytes from a newer runtime than the block the caller pinned. Such a read decodes today (v9 bytes, v9 decoder) and now raises `era-disagreement` instead. Unpinned reads cannot reach that refusal at all - see the qualifier above.
- `parseHexContractState` gains an optional third argument, `{ upperBound?: 'enforced' | 'withheld' }`, defaulting to `enforced`. The `EnvelopeUpperBound` type is exported alongside it.
- A state served with no block to date it fails fast (`undated-state`), matching the call #1177 made for the raw path.

## Follow-ups, deliberately not in this PR

- `RawContractState.version` is derived from the same read-viewpoint block, while its docs tell callers to narrow on it and then hand the bytes to that era's deserializer. For a contract dormant across a fork that points at the wrong runtime. The type does hedge ("not a verified statement about the bytes"), but the primary instruction should be reworded - it belongs to #1177.
- `nightly-e2e.yml` is the only coverage for eras other than v9 (it rotates `TESTKIT_DOCKER_ENV` across all five image sets) and it has been `startup_failure` with zero jobs launched for at least six consecutive nights. Every e2e signal on this stack is devnet, i.e. node 2.x / ledger v9 only.

## Test plan

- [x] Tests written first; the coverage table went 7 red -> green with the document changes, and the era rule was rewritten test-first (5 red -> green)
- [x] The withheld bound written test-first (4 red -> green): an unpinned read whose state is newer than the block dating it decodes on `queryContractState`, `queryZSwapAndContractState` and `watchForContractState`, and at unit level on `parseHexContractState`
- [x] The bound's survival pinned in the same pass: the two pinned-read tests were green *before* the change and stayed green, so the check was narrowed rather than removed; withholding it still refuses an undecodable envelope
- [x] Era matrix built from real minted bytes - both runtimes via `ledger` and `loadLedger8()` - covering both disagreement directions, the consistently-v8 case, older bytes under a newer block, an unresolvable version integer with a decodable and an undecodable envelope, a non-contract-state payload and non-hex input
- [x] Tag strictness pinned directly: genuine v8 bytes handed to `deserializeCompactContractState` raise a `DeserializationError` whose cause names `midnight:contract-state[v6]:`
- [x] Every refusal asserts it is not a `DeserializationError`, i.e. nothing reached a decoder
- [x] Latch engagement, non-engagement on a pinned read, and non-engagement on an unresolvable head version all tested by observable request counts, not private state
- [x] Full run, no regressions: 31/31 turbo test tasks, indexer 307 passed / 2 skipped
- [x] `yarn build` 16/16, `npx eslint .` clean, `yarn typecheck:tests` clean
- [x] Integration and e2e were green on the previous commit of this branch ([run](https://github.com/midnightntwrk/midnight-js/actions/runs/33375964407)) - all 26 e2e suites against a live indexer. **Not re-run locally for the rule change** (no Docker in the authoring environment); that commit only narrows refusal paths, and CI re-runs on this push.


VanessaPC (approved):

VanessaPC (approved):

packages/indexer-public-data-provider/src/codec.ts
```diff
@@ -113,6 +116,116 @@ export const contractStateEnvelopeVersion = (raw: Uint8Array): LedgerVersion =>
...
+ * @throws {IndexerDataError} When the state is not hex-encoded, when its
+ *   envelope is newer than the block that dates it, or when its era is not
+ *   decodable here.
+ * @throws {TagParseError} When the payload carries no supported contract-state
+ *   envelope.
+ */
+export const parseHexContractState = (hexState: string, protocolVersion: number): ContractState => {
+  const { raw, envelopeVersion } = stateBytesAndEnvelopeVersion(hexState);
+  const reportedVersion = reportedEra(protocolVersion);
+  if (reportedVersion !== undefined && isNewerEra(envelopeVersion, reportedVersion)) {
```
VanessaPC: **medium** — The `eraDisagreement` branch treats "envelope newer than the dating block" as impossible on a consistent indexer, but for an unpinned read (`offset === null`) `block` and `contract` are independently-resolved sibling Query-root fields. This package documents exactly that at `provider.ts:319-323` (`queryRawContractState`): "the indexer resolves Query-root siblings concurrently, from independent reads, so they can still come from different blocks".

Scenario: the indexer indexes the fork block between resolving the two root fields. `queryContractState()` then gets a pre-fork `block.protocolVersion` (v8) alongside post-fork v9 bytes from `contract`, and throws a hard `IndexerDataError` telling the operator the indexer is serving state and block data from different eras — on a read that succeeds today. That contradicts the PR description's "No read that succeeds today starts failing".

`CONTRACT_STATE_SUB` already shows the race-free alternative (date the state by the action's own `transaction { protocolVersion }`); `Contract.actions` would allow the same for `CONTRACT_STATE_QUERY`. The window is narrow (one block, once per fork), but the outcome is a hard throw with a remediation message that sends the operator hunting a fault that is not there.

sp-io: Fixed in b95adce.

The direction was right, the qualifier was not: an envelope newer than its dating block is impossible only while both signals describe the same *block*, which an unpinned read does not give. `parseHexContractState` now takes the bound as an explicit input — `enforced` where the version comes from the same resolution as the state (`queryDeployContractState`, the three `observables.ts` paths) or where both root fields are pinned to one offset, `withheld` on the unpinned sibling reads.

Two things your comment led me to:

- The exposed path is the default one, not an edge case: `queryContractState(address)` with no config is reached from `findDeployedContract` and the three governance flows.
- `queryZSwapAndContractState` carried the same defect in a comment — "callers pin `offset` to a specific block, so both fields resolve at the same anchor with no race between them" — but `getPublicStates` reaches it with no offset whenever its own caller passes no block hash. Corrected in the same commit.

I did not take the `Contract.actions` route. The schema does not say whether `actions` is scoped by the parent's `offset`; if it is not, a pinned read would end up dated by the newest action, and the bound would then fire wrongly on the path that is currently sound. Since the bound buys diagnostics rather than safety — the decoder is tag-strict, which `era-cross-check.test.ts` pins directly — narrowing it was the cheaper correct move. Restructuring the query is still open if you want the bound back on head reads; it needs a look at the indexer to confirm the semantics.

Where it still fires: a pinned read whose bytes are newer than the block the caller pinned. That is a genuinely inconsistent indexer, and the description no longer claims that no read starts failing.

PR #1208, opened 2026-08-31
fix(types): honor proveTxConfig.timeout in createProofProvider

### LINKED ISSUE #1061: createProofProvider silently drops proveTxConfig.timeout — same defect class as #974, on the canonical factory used by dappConnectorProofProvider

Ansh-Sonkusare:
## Problem

`createProofProvider` in `@midnight-ntwrk/midnight-js-types` is the simple factory that wraps any `ProvingProvider` into a `ProofProvider`. Its `proveTx` implementation did not declare the `proveTxConfig` parameter:

```typescript
// Before
async proveTx(unprovenTx: UnprovenTransaction): Promise<UnboundTransaction> {
  return unprovenTx.prove(provingProvider, costModel);
}
```

This silently drops any caller-supplied `timeout`:

```typescript
const proofProvider = createProofProvider(httpClientProvingProvider(url, zkConfig));
// timeout is never forwarded — the 5-second limit is lost
await proofProvider.proveTx(tx, { timeout: 5_000 });
```

The `ProofProvider` interface requires `proveTx(unprovenTx, proveTxConfig?)`, so this is a silent contract violation.

Closes #1061.

## Fix

Accept `proveTxConfig` and, when a timeout is set, build a thin wrapper around the underlying provider that forwards the per-call override to each `check`/`prove` call:

```typescript
async proveTx(unprovenTx, proveTxConfig?) {
  if (proveTxConfig?.timeout !== undefined) {
    const timeout = proveTxConfig.timeout;
    const pp = provingProvider as unknown as { check(..., t?: number): ...; prove(..., t?: number): ...; };
    const wrapped: ProvingProvider = {
      check: (preimage, key) => pp.check(preimage, key, timeout),
      prove: (preimage, key, binding) => pp.prove(preimage, key, binding, timeout),
      lookupKey: (key) => provingProvider.lookupKey(key),
    };
    return unprovenTx.prove(wrapped, costModel);
  }
  return unprovenTx.prove(provingProvider, costModel);
}
```

This mirrors the pattern already established in `httpClientProofProvider` by #1054. Providers that expose a per-call timeout (e.g. `TimeoutAwareProvingProvider` returned by `httpClientProvingProvider`) will honour the override; plain `ProvingProvider` implementations safely ignore the extra argument, preserving backward compatibility.

## Tests

Adds `packages/types/src/test/proof-provider.test.ts`, mirroring the spy-based pattern used in `packages/http-client-proof-provider/src/test/http-client-proof-provider.test.ts`. Covered cases:

- [x] `proveTxConfig.timeout` threaded to underlying `check` and `prove` (per-call timeout forwarding, issue #1061)
- [x] Omitted `proveTxConfig` — no override reaches `check`/`prove`
- [x] Explicit `{ timeout: undefined }` — no override reaches `check`/`prove`
- [x] Multiple `proveTx` calls each thread their own resolved timeout to the same underlying provider
- [x] No wrap when timeout absent — underlying provider passes through by referential identity (plain `ProvingProvider` implementations are unaffected)
- [x] Wrap when timeout set — captured provider is not the original; `lookupKey` delegates to the underlying provider
- [x] Backward compatibility — a plain `ProvingProvider` whose `check`/`prove` have no override-timeout parameter accepts the extra trailing argument at runtime without error

### DIFF
--- a/packages/types/src/proof-provider.ts
+++ b/packages/types/src/proof-provider.ts
@@ -61,12 +61,33 @@ export interface ProofProvider {
  * @param provingProvider - The underlying proving provider used to generate proofs.
  * @param costModel - Optional cost model to use for proof generation. Defaults to the initial cost model if not provided.
  * @returns A {@link ProofProvider} that delegates proof generation to the given proving provider.
+ *
+ * @remarks
+ * When `proveTxConfig.timeout` is set and `provingProvider` is a timeout-aware implementation
+ * (e.g. the result of `httpClientProvingProvider`), the timeout is forwarded to each circuit-level
+ * call. Plain {@link ProvingProvider} implementations that do not declare an override timeout
+ * parameter safely ignore the extra argument at runtime.
  */
 export const createProofProvider = (
   provingProvider: ProvingProvider,
   costModel: CostModel = CostModel.initialCostModel()
 ): ProofProvider => ({
-  async proveTx(unprovenTx: UnprovenTransaction): Promise<UnboundTransaction> {
+  async proveTx(unprovenTx: UnprovenTransaction, proveTxConfig?: ProveTxConfig): Promise<UnboundTransaction> {
+    if (proveTxConfig?.timeout !== undefined) {
+      const timeout = proveTxConfig.timeout;
+      // Cast to the extended signature accepted by timeout-aware providers (e.g. httpClientProvingProvider).
+      // The extra argument is ignored by implementations that don't support per-call timeout overrides.
+      const pp = provingProvider as unknown as {
+        check(preimage: Uint8Array, keyLocation: string, overrideTimeout?: number): Promise<(bigint | undefined)[]>;
+        prove(preimage: Uint8Array, keyLocation: string, overwriteBindingInput?: bigint, overrideTimeout?: number): Promise<Uint8Array>;
+      };
+      const wrapped: ProvingProvider = {
+        check: (preimage, keyLocation) => pp.check(preimage, keyLocation, timeout),
+        prove: (preimage, keyLocation, overwriteBindingInput) => pp.prove(preimage, keyLocation, overwriteBindingInput, timeout),
+        lookupKey: (keyLocation) => provingProvider.lookupKey(keyLocation),
+      };
+      return unprovenTx.prove(wrapped, costModel);
+    }
     return unprovenTx.prove(provingProvider, costModel);
   }
 });
--- a/packages/types/src/test/proof-provider.test.ts
+++ b/packages/types/src/test/proof-provider.test.ts
@@ -1,0 +1,169 @@
+/*
+ * This file is part of midnight-js.
+ * Copyright (C) Midnight Foundation
+ * SPDX-License-Identifier: Apache-2.0
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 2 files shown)


Ansh-Sonkusare:
Closing as a duplicate of #1063, which I hadn't seen when I opened this. #1063 has broader scope (also fixes the dapp-connector branch called out in #1061) and uses an explicit brand symbol that's more robust than my unconditional wrap. Apologies for the parallel work.

PR #1211, opened 2026-09-01
fix(testkit-js): increase health check timeout from 1000ms to 15000ms

Isaaco3349:
Fixes #<issue-number-you-filed-earlier>. Increases the hardcoded 1000ms axios timeout on testkit-js's 4 environment health checks (node, indexer, proof server, faucet) to 15000ms. 

The original timeout is too tight for real-world network conditions, especially WSL2/Docker Desktop setups, causing intermittent false-negative environment startup failures even when all endpoints are genuinely healthy.

Verified locally by patching the compiled bundle before submitting this source fix.

### DIFF
--- a/testkit-js/testkit-js/src/client/faucet-client.ts
+++ b/testkit-js/testkit-js/src/client/faucet-client.ts
@@ -43,7 +43,7 @@ export class FaucetClient {
    */
   async health() {
     const url = buildUrlWithPath(this.faucetUrl, '/api/health');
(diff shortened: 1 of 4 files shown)


sp-io:
Thanks for spotting this and opening the PR. Sorry for the delay -- I'm heads-down on a large feature with a tight deadline, so reviews are lagging. I'll take a proper look as soon as that clears. Thanks for your patience.

PR #1219, opened 2026-09-02
ci: resolve CI tooling through yarn.lock instead of npm

sp-io:
## Summary

Alternative to #1135, reached from the same starting point: several CI steps fetch their tooling from the npm registry at run time, with no lockfile behind them.

#1135 hardens those fetches by pinning the npm CLI to 12.0.2 everywhere and adding an `.npmrc` cooldown. This PR removes most of the fetches instead, so there is far less left to harden.

**What was actually unpinned**

| Step | Call | Pinned? |
|---|---|---|
| Coverage merge (`ci-base.yml`) | `npx --yes nyc` | no version at all |
| Commit lint (`commit-lint.yml`) | `npm install -g @commitlint/cli@18.6.1` | top level only |
| Pre-publish validation (`validate-packages`) | `npx publint@latest` | `@latest`, in the **publish** job |

In every case the top-level version was pinned at best. Transitive dependencies resolved fresh on each run against floating ranges, with install lifecycle scripts enabled — precisely the vector behind the keyv/cacheable worm.

**Change**

Route all three through the workspace, so they are pinned by `yarn.lock` and integrity-checked:

- add `nyc` and `publint` as root devDependencies
- `ci-base.yml`: `npx --yes nyc` → `yarn nyc`
- `validate-packages`: `npx publint@latest` → `yarn publint`
- `commit-lint.yml`: install with `yarn install --immutable`, run `yarn commitlint`

**Bonus fix.** The commitlint job installed `@commitlint/cli@18.6.1` globally while the husky hook uses the root devDependency at `20.4.3` — CI and local developers were linting commits with different major versions. Both now use `20.4.3`.

**Where npm stays, on purpose**

`e2e_tests_reports` in `ci-testkit-js-e2e.yml` merges artifacts only. It has no `yarn install` and pulls three report CLIs ad hoc: `allure-commandline`, `ctrf`, `junit-report-merger`. Giving it a full workspace install would mean installing the whole testkit tree (wallet SDK, ledger WASM) in a job that currently finishes in seconds.

So this job keeps `npx`, and gets `#1135`'s hardening scoped to itself: its own `setup-node` plus a pinned `npm@12.0.2`, which refuses to run dependency lifecycle scripts by default. The blast radius is a single report — a bad npm release cannot touch a build, test or publish.

`junit-report-merger@9.0.4` pulls `commander`, `tinyglobby` and `xmlbuilder2` on floating ranges, so the pin is load-bearing there. `allure-commandline@2.43.0` has zero dependencies and no install scripts, so for that one the pin is defence in depth rather than active protection.

**Relationship to #1135**

- The `Update NPM` steps in `ci-base.yml` and `commit-lint.yml` become unnecessary — those jobs no longer use npm.
- The `.nvmrc` bump #1135 needs (npm 12 requires node `^22.22.2 || ^24.15.0 || >=26`, `.nvmrc` is at `24.11.1`) is not needed either; the one job that runs npm 12 pins its own node.
- The `.npmrc` cooldown from #1135 is **not** carried over. Dependencies here are installed by Yarn 4, which does not read `.npmrc` at all, so `min-release-age` would have covered only the remaining `npx` calls in the E2E report job. If you want it, it can land on top — but the Yarn-native equivalent is `enableScripts: false` in `.yarnrc.yml` with a `dependenciesMeta.<pkg>.built: true` allowlist, which is a separate piece of work.

**Not addressed**

`validate-packages` still ends both loops with `|| true`, swallowing publint and pack failures. Fixing that would change what the release job accepts and belongs in its own PR.

## Test plan

- [x] `yarn install` — `nyc@18.0.0` and `publint@0.3.24` resolve, `yarn.lock` committed
- [x] `yarn constraints` clean
- [x] `yarn nyc merge` + `yarn nyc report --reporter=html` verified locally against two synthetic `coverage-final.json` files: both entries survive the merge, HTML report generated
- [x] `yarn publint ./packages/types --strict` → `All good!`, exit 0
- [x] `yarn commitlint --from origin/main~25 --to origin/main` under 20.4.3: no new failures. The single reported error is a >100-char line inside a code block in a GitHub squash-merge body, which CI never lints (it lints `BASE_SHA..HEAD_SHA`, i.e. the PR's own commits) and which `config-conventional@18` flags identically
- [x] `yarn build`, `yarn lint`, `yarn typecheck:tests` all pass
- [ ] Workflow changes verified by this PR's own CI run


sp-io:
Closing — we are not going this way.

The trade this PR makes is not worth it. Routing `nyc` through the workspace means adding a package with 27 direct dependencies to the root `devDependencies`, so every developer pays for it on every `yarn install`, to serve a single CI step. Relying on `yarn` to launch CI tooling is also not something we want to lean on further.

#1135 stays the approach: keep the tooling on npm and harden npm itself, rather than migrating the tooling to yarn.

Three findings from the review of this branch are worth carrying over to #1135, since they apply to it too:

1. **The `Update NPM` step fails today.** `npm@12.0.2` requires node `^22.22.2 || ^24.15.0 || >=26.0.0`; `.nvmrc` is `24.11.1`, so `npm install -g npm@12.0.2` exits with `EBADENGINE`. This is the red check on #1135.
2. **`e2e_tests_reports` in `ci-testkit-js-e2e.yml` has no `setup-node` at all**, so its `Update NPM` step will fail the same way once the e2e matrix finishes and the job actually runs.
3. **An unguarded pin step is worse than none.** Every report step in that job is `if: always()`. A pin step without a guard is skipped when an earlier step fails, while the `npx` fetches below it still run — on whatever npm the runner ships. The pin has to sit above the artifact download and carry `if: always()`, or the guarantee is void exactly on the failure path.

Also noted while reviewing, unrelated to the npm question and worth separate tickets:

- `.github/actions/validate-packages/action.yml` ends both loops with `|| true`, so the step named "Validate Packages Before Publishing" cannot block a publish. Its second loop also never uses `$dir`, packing the monorepo root N times.
- `commitlint.config.js` downgrades `type-enum`, `scope-enum`, `subject-empty`, `type-empty` and `scope-empty` to severity 1, so `wibble(nonsense): whatever` passes the Commit Lint gate with exit 0.
- CI validates commits with a globally installed `@commitlint/cli@18.6.1` while the husky hook uses the root devDependency, currently resolving to `20.5.0`.
- The coverage chain in `ci-base.yml` passes green when `find` matches zero `coverage-final.json` files, uploading an empty report that looks real.

Issue #1005, opened 2026-06-29
[HF v8→v9] MJS-02 — contracts package: protocol-version orchestration


DarrenOliveiro-Priestnall:
**Workstream:** Midnight.JS — contracts package (midnight-js)
**Part of:** Ledger v8→v9 Hard Fork Migration — [SOW-Q3-10 / product#119](https://github.com/shieldedtech/product/issues/119) ([scope comment](https://github.com/shieldedtech/product/issues/119#issuecomment-4833224770))
**First-pass sizing (to confirm with team):** 6–10 person-days

## Summary
Update the Midnight.JS `contracts` package so an **already-deployed, ledger-8-compiled contract keeps transacting after the fork with no recompilation and no v9-compiled variant** — **without impairing native ledger-9 contract execution**. The package subscribes to the protocol version and, once v9 is active, wraps the unchanged ledger-8 execution in a native ledger-9 transaction; it does **not** dispatch between two compiled contract artifacts.

## Scope
- Subscribe to protocol version from the public state provider and switch behaviour once v9 is active.
- Implement the post-fork **keep-state execution path** for pre-fork contracts (three moves):
  1. **Keep** the deployed ledger-8 artifacts + runtime (no recompile).
  2. **Down-convert** the migrated on-chain state back to the ledger-8 shape for execution — productionize [`island-3/driver/src/downcast.ts`](https://github.com/shieldedtech/spike-dapp-hf/blob/main/island-3/driver/src/downcast.ts) (`downcastV9StateForExecution`, incl. the contract-agnostic `rehashStateValue`).
  3. **Build + prove** a native ledger-9 `ContractCallPrototype` around the resulting POJO transcript — productionize [`island-3/driver/src/assemble.ts`](https://github.com/shieldedtech/spike-dapp-hf/blob/main/island-3/driver/src/assemble.ts) (`assembleCallV9`). Proof version is selected by the resolved verifier-key/IR tag (V2 → migrated `co.v2` slot), **not** hardcoded.
- Route pre-fork (v8-native) contract calls unchanged.
- **The HF work MUST NOT impair native ledger-9 contract execution** — the v9-native call/deploy path stays the default and must remain unaffected by the added dual-version dispatch and v8 keep-state path.

## Acceptance criteria
- A contract deployed under ledger-8 accepts a **new transaction after the fork with no recompilation and no v9 contract variant**.
[...]

PR #1226, opened 2026-09-04
test(midnight-js): capture v9-native non-regression baselines

sp-io:
## Summary

First PR of the #1005 (MJS-02) stack, and deliberately a no-production-change one: it captures the
non-regression baselines for `packages/contracts` **before** any of the era-dispatch work in this
phase touches the package. Test-only, plus one doc — 0 production files changed.

**Golden fixtures for the three deterministic composition stages** — decoded contract state,
guaranteed transcript, and the assembled unproven call — captured from a real `deposit` execution
against the compiled `shielded-map` contract already in the tree. A fourth stage asserts structure
only, because `Transaction.fromPartsRandomized` samples fresh binding randomness per call.

The golden stages read the assembled call back off `tx.intents` rather than rebuilding it. That is
load-bearing: a fixture built by reproducing `createUnprovenLedgerCallTx`'s own argument list would
still pass byte-identically after a change to that list, because it never calls the function it
claims to guard.

**The gates are built so they can actually fail.** Each of these was verified by deliberate
perturbation, not by assumption:

- `Effects` has nine fields and five of them are `Map`s. `JSON.stringify` renders a `Map` as `{}`
  whatever it holds, so a plain-JSON normalizer cannot see them — `claimedUnshieldedSpends` among
  them, the one field `extractUserAddressedOutputs` reads. The normalizer expands `Map`/`Set` into
  tagged entry arrays, and stage 2 asserts the nine-key set by name.
- Stage 2 pins the assembled call's communication commitment, re-derived through the ledger's own
  function. Nothing else on a `ContractCall` observes `input`, `output` or the bound randomness —
  the Debug repr redacts the commitment — so without this, discarding the bound randomness for a
  fresh sample passes every stage.
- The input-hash gate pins its key set to two literals and lives in its own top-level suite, so it
  neither compares an empty set to itself nor gets buried when the stages' shared setup is what
  broke.
- The redaction asserts one match per gas field name before redacting, so a renamed, dropped or
  non-numeric field fails by name rather than the `replace` silently matching nothing.

**The fixture's own inputs are pinned.** The sha256 of the compiled contract and its verifier key are
recorded in the fixture and checked by the suite. Recompiling the contract moves every byte-pinned
value at once, which is indistinguishable from a composition regression unless the inputs are
checked first. Both are regenerated together, in one commit.

**Regeneration is a command, not a hand-edit**: `UPDATE_GOLDEN=1 yarn test
src/test/non-regression-golden.test.ts` records what each stage observed and writes the fixture
back, artifact hashes included. It is opt-in, because a run in that mode will happily record a
genuine regression as the new baseline.

**Gas is shape-checked, not pinned** — but not because it is nondeterministic. Every `RunningCost`
field is documented "(modelled)" and five consecutive runs give byte-identical values. It is
excluded because it moves on any cost-model bump with no change in composition behaviour. Note the
Debug string renders `read_time`/`compute_time` as formatted durations (`255.000us`, `5.834ms`), so
the redaction matches a whole value rather than only its digits.

**Cross-call unshielded aggregation** gets coverage in `ledger-utils.test.ts`, next to the existing
single-call unshielded-offer tests. `createUnprovenLedgerCallTx` aggregates user-addressed outputs
across every call with `calls.flatMap(...)`; every other test there passes a single call, so
reverting to the root call alone was undetectable — and `claimedUnshieldedSpends` is a tuple-keyed
`Map` present in both ledger eras, so it is what an era shim is most likely to convert wrongly.

Rationale, blind spots, the regeneration procedure and the failure-triage decision tree live in
`docs/architecture/contracts-non-regression-golden-baselines.md`, not in source comments.

**Known blind spot, recorded in that doc:** an `input`/`output` transposition is undetectable here,
because `deposit` returns void, so `output` is the empty `AlignedValue` and both orderings hash
alike. `communicationCommitment` is order-sensitive in general; closing this needs a circuit with a
non-empty return value.

Baseline is the integration branch `feat/1004-protocol-dual-ledger`, not `main` — `main`'s hashes
are the wrong baseline for this package.

## Test plan

- [x] Tests written first (TDD), verified they fail before the fix
- [x] All existing tests pass (no regressions) — 24 files / 247 tests
- [x] Lint clean (`eslint` exit 0), typecheck clean (`tsc --noEmit` exit 0), build succeeds

Regenerating after these changes reproduces the previously pinned values **byte for byte**, so this
fixes the gates, not the baseline.

Every gate verified by deliberate perturbation, each failing in exactly one place:

| Perturbation | Result |
|---|---|
| `calls.flatMap` reverted to the root call alone | 2 aggregation tests fail |
| bound commitment randomness discarded for a fresh sample | stage 2 fails (previously passed) |
| guaranteed/fallible transcript slots swapped | stages 2, 3 and 4 fail |
| a `Map` populated in the fixture | stage 2 fails (previously passed) |
| one byte of the state hex flipped | stage 1 only |
| one character of a recorded input hash flipped | input-hash gate only |
| an artifact dropped from `inputHashes.files` | both input tests fail (previously passed vacuously) |


sp-io:
Closed and reopened as #1240 to move the base onto the new MJS-02 integration branch `feat/1005-contracts`. The base could not be changed in place: this PR was a member of GitHub stack 1235, which returns HTTP 422 on a base change and exposes no API to dissolve it.

Same head branch (`feat/1005-contracts-baselines`), same commits, same diff. Nothing was lost — this PR carried no reviews and no human comments.

PR #1236, opened 2026-09-05
feat(midnight-js): decode v8 history lazily and surface versioned records

sp-io:
## Summary

Makes the indexer public-data provider decode **both** ledger eras on the finalized-transaction read path. Before this, a v8-era record was named and refused (`EraUnsupportedError`); it is now decoded and surfaced on the existing `'v8'` arm of `VersionedFinalizedTxData`, which until now no provider produced.

- Per-record era dispatch at both construction sites (`toFinalizedDeployTxData` and the `watchForTxData` path), through a frozen null-prototype decoder table keyed by era.
- The v8 runtime is reached only through `loadLedger8()` from the protocol barrel, so a v9-only session never triggers the v8 WASM. Verified by a spy on the accessor and by the built bundle carrying no `ledger-v8` reference.
- A read's era comes from the `protocolVersion` that read carries; no read issues a network-head request as a side effect. Pinned by a test whose provider rejects on every `query` call.
- New `DecodeVersionMismatchError` (`MIDNIGHT_JS_PR_DECODE_VERSION_MISMATCH`) for a record whose bytes name a different vintage than the block dating it, plus an `isHex` guard on the transaction path so a non-hex payload is diagnosed as malformed encoding rather than silently truncated.
- The 13-field record literal that was duplicated across the two seams in different orders is now built in one place.
- Release documentation for 5.0.0 updated to describe dual decode: six statements that this change falsified now state current behaviour, and `EraUnsupportedError` is documented as what it became rather than dropped.

### Base branch

This PR targets `feat/1005-contracts-breadcrumbs` (#1233), the top of the MJS-02 stack, rather than the integration branch `feat/1004-protocol-dual-ledger`. Two reasons: #1228 relocates `contractStateEnvelopeVersion` out of `indexer/src/codec.ts`, a file this PR edits, and the follow-up task that lifts the v8 refusal at the provider seams has no producer to test against without #1229. It merges into the integration branch once the MJS-02 stack lands.

### Notable decision

`DecodeVersionMismatchError` fires only when the failure carries a `direction` conclusion from the deserialization classifier, not merely a `version-mismatch` classification. The looser test admits corruption: probed against the real v8/v9 runtimes, empty, truncated, garbage and bit-flipped payloads all classify as `version-mismatch`, and a payload truncated mid-body even carries a `receivedVersion` from its intact header tag. Both directions of the guard are pinned by mutation-tested negatives.

## Test plan

- [x] Tests written first (TDD), verified they fail before the implementation

Indexer package: 353 passed / 2 skipped. New `dual-decode.test.ts`: 37 tests, covering both arms round-tripping `toHex(record.tx.serialize())` back to the input hex through the real runtimes, the lazy-acquisition gates, the version-truth invariant on both arms, nine negative decode and encoding shapes, and cause-chain sanitization. All 31 turbo test tasks green.


sp-io:
Closed and reopened as #1241 to move the base onto the new MJS-03 integration branch `feat/1006-providers`, cut from the tip of `feat/1005-contracts-breadcrumbs`. The base could not be changed in place: this PR was a member of GitHub stack 1235, which returns HTTP 422 on a base change and exposes no API to dissolve it.

Same head branch (`feat/1006-provider-dual-decode`), same commits, same diff. Nothing was lost — this PR carried no reviews and no human comments.

PR #1252, opened 2026-09-07
ci: grant nightly-e2e the actions:read its build workflow requires

sp-io:
## Summary

- `nightly-e2e.yml` has been failing with `startup_failure` and **zero jobs** every night since at least 2026-08-26 (14 consecutive runs checked). Root cause: it calls `ci-testkit-js-build.yml`, which requests `actions: read`, and it was the only caller that did not grant it. A reusable workflow cannot be granted more than its caller holds, so GitHub refuses to start the run at all — no jobs, no annotation, nothing in the API to read. `ci.yml` and `ci-testkit-js.yml` both grant it; nightly did not.
- Adds `.github/scripts/check-workflow-permissions.mjs`, which reads the caller/callee relationship statically and names the missing scope.
- Adds `.github/workflows/workflow-permissions.yml`, which runs that gate — see below for why it is a standalone workflow rather than a step in `ci-base.yml`.
- Declares `yaml` as a root devDependency. It was already resolvable transitively at 2.9.0, and `resolutions` already carries a `>=2.8.3` **floor** (not a pin), but a script that imports it directly should not rely on hoisting.

### Why the gate is its own workflow

The first version of this PR ran the gate as a step in `ci-base.yml`, which is reached through `ci.yml` → `ci-midnight-js.yml` → `ci-base.yml` — the very chain it validates. That does not work. A `startup_failure` produces **zero** jobs, so a permission fault anywhere in that chain deletes the step meant to diagnose it: the run that was supposed to carry the message is the run that never starts.

`workflow-permissions.yml` therefore calls no reusable workflow and performs no workspace install. The workspace install needs a GitHub Packages token, and a gate that depends on a secret stops working exactly where an unprivileged run needs it most; instead the step installs the single public, dependency-free `yaml` package under a prefix outside the checkout (an `npm install` at the root of a Yarn workspace tries to resolve the whole monorepo and dies on peer conflicts) and links it in. The whole job runs in seconds.

**This new workflow is not yet a required check** — worth adding in branch protection, otherwise it can go red without blocking.

### What the gate checks

A callee's requirement is the per-scope maximum across its workflow-level `permissions:` block **and every job-level block**. Reading only the workflow-level block is a false negative: a callee whose top level looks modest can still over-ask one job down. Taking the maximum also makes the check sound across a chain — when A calls B and B calls C, C's demands surface in B's requirement, so the A→B pair catches a shortfall only C would have exposed at run time.

Reusable workflows in other repositories (five of them here) cannot be read from this repo, so they are reported as **unchecked** rather than folded into a pass.

### Scope note

This is a real bug worth fixing, but it is **not** a blocker for the fork-crossing e2e work. e2e already runs on every PR through `ci.yml` → `ci-testkit-js.yml` → `ci-testkit-js-e2e.yml`, pinned to the `devnet` image set. Nightly's only added value is rotating the five `testkit-js/env/*.env` sets, and none of those is a fork-capable topology, so it was never going to supply the environment that work needs.

### What the first successful nightly will show

Two matrix entries are expected to go red, by design rather than by regression: `mainnet.env` and `preprod.env` pin node **0.22.x**, and `protocolVersionToLedger` deliberately does not map node 0.x (#1218, `2d349f1b`, spec OQ1 as shipped). Anything reading the head version fails there on purpose. Flagging it here so it is not diagnosed from scratch at 02:00.

## Test plan

- [x] Tests written first (TDD), verified they fail before fix

The gate was run **before** the fix and reported exactly the diagnosed bug, independently of how the diagnosis was reached:

```
::error::nightly-e2e.yml job 'build' grants actions: none but ci-testkit-js-build.yml
requires actions: read -- the run will fail at startup with no jobs and no annotation
```

After the fix: `Checked 21 workflow files; every local reusable-workflow call is granted what it requires.`

The script now has a vitest suite (13 tests, `yarn test:workflow-scripts`, run in `ci-base.yml`). Reverting the job-level-permissions fix fails exactly the two tests that cover it and no others. Four sabotage runs against the final code:

| Sabotage | Result |
|---|---|
| nightly loses `actions: read` | the original diagnosis, verbatim |
| job-level over-ask inside `ci-testkit-js-e2e.yml` | caught, naming **both** callers — this passed silently before |
| `checks: readonly` typo in `ci.yml` | `::error::` annotation, not a stack trace |
| `ci.yml` drops `checks: write` | both affected jobs reported |

The last case is the one that motivated moving the gate out of `ci-base.yml`: it is only reportable from a workflow that survives the fault.

Also verified: the gate runs correctly from any working directory, and the install step works against a clean checkout with no `node_modules`.

---

### DIFF
--- a/.github/scripts/check-workflow-permissions.mjs
+++ b/.github/scripts/check-workflow-permissions.mjs
@@ -20,20 +20,30 @@
 // omitted `actions: read`.
 //
 // This gate reads the same relationship statically and names the missing scope.
+// It is run by `.github/workflows/workflow-permissions.yml`, which deliberately
+// calls no reusable workflow of its own -- see the comment there.
 
 import { readFileSync } from 'node:fs';
 import { readdir } from 'node:fs/promises';
 import path from 'node:path';
 import process from 'node:process';
+import { fileURLToPath } from 'node:url';
 
 import { parse } from 'yaml';
 
-const WORKFLOW_DIRECTORY = '.github/workflows';
+/** Resolved from this file, not from `process.cwd()`, so the gate runs from anywhere. */
+export const WORKFLOW_DIRECTORY = fileURLToPath(new URL('../workflows', import.meta.url));
 
 /** Permission levels, ordered. A caller must hold at least the level its callee asks for. */
 const LEVELS = { none: 0, read: 1, write: 2 };
 
-/** Every scope GitHub can grant, needed to expand the `read-all` / `write-all` shorthands. */
+/**
+ * Every scope GitHub can grant, needed to expand the `read-all` / `write-all`
+ * shorthands. Scopes outside this list still compare correctly against each
+ * other; the list only bounds what the two shorthands are taken to cover, so a
+ * scope GitHub adds later would be over-reported against a `read-all` caller
+ * rather than silently skipped.
+ */
 const ALL_SCOPES = [
   'actions',
   'attestations',
@@ -59,7 +69,7 @@ const ALL_SCOPES = [
  * empty one: absent means "inherit whatever the repository default is", which
  * cannot be resolved statically, while `{}` means "grant nothing".
  */
-const normalisePermissions = (value) => {
+export const normalisePermissions = (value) => {
   if (value === undefined || value === null) {
     return undefined;
   }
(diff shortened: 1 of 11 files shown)


VanessaPC (approved):

DarrenOliveiro-Priestnall (approved):
The only low security issue is in workflow-permissions.yml, where "npm install" should ideally pass in "--ignore-scripts" to avoid the execution of lifecycle scripts from a package being loaded. There aren't any scripts right now though so it isn't a problem, it's more of a guard for the future.

Approving since this isn't an issue today and is low risk, just a suggestion to guard against future malicious scripts being introduced and run

PR #1257, opened 2026-09-07
test(midnight-js): install the framework the way a consumer does (AC6, OQ13)

sp-io:
## Summary

Phase 5, Task 5.1 — the split-topology integration milestone (spec §10 step 5, OQ13), and the isolated-linker install smoke AC6 asks for.

This is the only lane that exercises the framework **as a consumer installs it**: packed tarballs, an isolated linker, outside the workspace. Inside the monorepo every package resolves through one hoisted tree, so the situation this is about cannot even be represented there.

- `packaging/pack-framework.mjs` — packs all 15 publishable workspaces and writes a name → tarball manifest. `yarn pack` resolves `workspace:*` ranges to the concrete versions a release would carry, so the tarball is a faithful stand-in for the published artifact.
- `packaging/personas.mjs` — two dApp personas and the declarations that make them different.
- `packaging/persona-entry.mjs` — runs *inside* the installed persona, where the assertions mean something.
- `packaging/linker-smoke.mjs` — builds, installs and runs each persona × linker combination.
- `ci-packaging.yml`, called from `ci.yml`.

### The experiment is the Compact runtime

A retained-era dApp pins `@midnight-ntwrk/compact-runtime@0.16.0`, because the contract module its pre-fork toolchain emitted opens with:

```js
__compactRuntime.checkRuntimeVersion('0.16.0');
```

and throws otherwise. The framework it installs pins `0.19.0-rc.0` for itself. **Both majors have to resolve in one install** — which a hoisted linker cannot do and pnpm's default linker and Yarn PnP can. That is precisely what the personas assert, from inside the install:

1. the persona's own `@midnight-ntwrk/compact-runtime` is its era's version;
2. the framework's copy, resolved *from* `midnight-js-protocol`, is a different one;
3. the contract module imports successfully — a passing `checkRuntimeVersion` is the assertion, and it is the line that fails loudly if the two copies above collapsed into one;
4. for the retained persona, `loadLedger8()` resolves and `onchain-runtime-v3` resolves to exactly one physical copy (two would be `Ledger8InstanceMismatchError`).

A hoisted linker is deliberately **absent** from the matrix rather than expected-to-fail: it cannot satisfy a retained-era dApp at all, which is a statement for the migration guide rather than a test case.

### Why this had to come before AC0

I originally sequenced the AC0 fork-crossing scenario ahead of this. That was wrong, and the plan's own PR table (1006-E before 1006-G) had it right. AC0 requires **one session** that executes both a retained contract and a current one, so it needs two runtime majors resolvable at once — exactly what this PR establishes. The AC0 PR builds on this.

### Corepack

`pnpm` is pinned as a root devDependency and invoked directly rather than through Corepack, which resolves and verifies its own download — locally that fails outright with `Cannot find matching keyid` on an older Corepack. Pinning also means the version under test is a reviewable line in `package.json` rather than whatever Corepack fetches.

## Test plan

- [x] Tests written first (TDD), verified they fail before fix

**Verified locally up to the registry-auth boundary, and no further — stating that plainly rather than implying a green run.** `pack-framework.mjs` produces all 15 tarballs; `linker-smoke.mjs` builds the persona, invokes the pinned pnpm, and reaches dependency resolution, where it stops on:

```
ERR_PNPM_FETCH_401  GET https://npm.pkg.github.com/@midnight-ntwrk%2Fcompact-runtime: Unauthorized
```

That is my local token lacking read access to the scope, not a defect in the harness — the personas resolve `@midnight-ntwrk/compact-runtime` from GitHub Packages, and neither pnpm nor a PnP install inherits Yarn's scope configuration. The CI job writes the credential where each looks. **The first run of this PR is the real verification of the smoke, and I will confirm it before asking anyone to merge.**

Everything not gated on the registry was verified locally: 15/15 workspaces pack, the manifest is written, persona manifests generate with the full override set (a tarball's own dependency ranges point at unpublished registry versions, so pinning only the named packages is not enough), and `yarn lint` is clean after adding the generated `packaging/.tarballs/**` and `packaging/.personas/**` trees to the eslint ignores — without that, eslint lints the compiler-emitted contract module it copies in.

### Not in this PR

The **Vite laziness fixture** — the other AC6 packaging gate — is not here. It is a browser-bundler fixture with its own build, its own failure modes and its own review surface, and #1239 already flagged that it may go red on a pre-existing eager load of protocol's root (`packages/contracts/src/internal/ledger8-entry.ts` statically imports `loadLedger8Engine`) rather than on anything the bundler does. Bundling it with this would make both harder to review, and it needs that finding resolved first.

## Stacking

Base is `test/1006-counter016-zk-artifacts` (#1256) → `test/1006-v8-era-lane` (#1255) → `feat/1006-fork-e2e-env` (#1253) → `feat/1006-hardening`, the Phase 5 integration branch cut from #1239. That integration branch is not intended to be merged to `main` on its own.

---


sp-io:
Reopening from a correctly-named branch: this PR is the split-topology packaging work, not the AC0 scenario. Superseded by the PR from `test/1006-split-topology`.

PR #1309, opened 2026-09-15
refactor(midnight-js)!: build a call's Zswap offer from the split the composer resolves

sp-io:
## Summary

`LedgerEra.composeCallTx` took the Zswap offer as ready-made bytes but drew the guaranteed/fallible split inside itself. A caller that had to route a coin into the segment its movement belongs to therefore needed the split *before* it could supply the offer — a cycle in the API shape. It was cut by publishing the split step as a fifth `LedgerEra` method, `partitionCallTranscript`, which its own docstring labelled a **PROTOTYPE SEAM** on an interface covered by semantic versioning (finding 8 of the #1218 architecture review).

This inverts the dependency instead of relabelling or hiding the prototype.

- `composeCallTx` now takes a `zswapOffer` **factory** and calls it back with every call's split once the calls are assembled. It answers with `ComposeCallResultPojo` — the transaction bytes plus those same splits.
- An offer can no longer be built *before* the split exists, which is how routing every movement into the guaranteed segment — the defect the wallet reports as `Wallet.InsufficientFunds` — used to be reached: by omission, with no way to do otherwise.
- `partitionCallTranscript` is removed from `LedgerEra` and from `assemble-call`. The seam is back to its four era-symmetric methods, which is what `era-seam.md` claimed all along.
- `assembleCallPrototype` answers with the prototype **and** the pair it was built from, so the split is drawn exactly once per call.
- The retained pipeline hands the composer `kind: 'unpartitioned'` and builds its offer inside the factory.
- `kind: 'partitioned'` **stays**. A caller that already holds a split is a real case (that is how the current era arrives from compact-runtime 0.17); it simply is no longer how the retained pipeline talks to the composer.

One method fewer on a semver-covered seam.

### Breaking changes

- `LedgerEra.composeCallTx` returns `ComposeCallResultPojo` rather than `Uint8Array`
- `ComposeCallOptions.guaranteedZswapOffer` / `.fallibleZswapOffer` replaced by `.zswapOffer`
- `LedgerEra.partitionCallTranscript` and `EraPartitionCallOptions` removed
- `ZswapOfferFactory` and `ComposeCallResultPojo` published in their place

`ComposeDeployOptions` is unchanged: a deploy has no transcript to split.

### One deliberate contract change

A call's offer bytes are now read **after** assembly rather than before, because the factory cannot run earlier. The coded error is unchanged (`ComposeOptionError`, `option: 'zswapOffer'`), and the deploy legs keep the old order. Recorded in `ComposeRefusalOrder`, together with why the trade is worth it: bytes routed into the wrong segment pass every check here and are rejected at balancing time by a diagnosis that names neither the call nor the segment.

Three paragraphs of `compose-refusal-order.md` were deleted rather than rewritten — they existed only to explain why the partitioner's own answer, sent back through the seam dressed as a caller-supplied pair, could not trip the empty-pair refusal. That case no longer exists.

## Follow-up fix (b19438ef)

A review of the commit above found that the reordering had been applied to only one arm, plus a silent path the factory shape opened. Both are fixed here, each driven RED first.

**The two era arms refused in different orders.** The v9 call leg invoked the factory *before* `aggregateUnshieldedOffers`, the v8 leg *after*. Both steps refuse, so a call carrying an unpayable payout **and** unreadable offer bytes was diagnosed as `ComposeOptionError` / `'zswapOffer'` on v9 but `ComposeFailedError` / `'call-unsupported-payout'` on v8 — contradicting `compose-refusal-order.md`, which states both arms refuse in one order. This was a regression against `main`, where both arms read offers up front. The v9 factory call now sits below the unshielded aggregation, so offers really are read last on both arms.

**An era that ignored the factory composed a coin-less transaction, silently.** `zswapOffer` is optional on the seam, so an implementation that never calls back was type-correct — and the retained pipeline builds its offer *only* inside that callback. The result reported both offers as `undefined`, which is also the honest shape of a call that moved no shielded coin, so nothing downstream could tell them apart. The pipeline now holds one `routed` binding that stays `undefined` until the factory runs and refuses via `assertDefined` if it never does, which separates "moved nothing" from "was never asked". A short `partitions` list is refused the same way rather than falling through to the router's `[undefined, undefined]` default.

Docs corrected in the same commit: `README.md` no longer documents the removed `guaranteedZswapOffer` / `fallibleZswapOffer` call options; its usage example routes against every partition rather than destructuring the first (which on a v9 call tree would route the whole transaction against a callee's split); the v8-native "offers are handles" note now distinguishes the deploy leg from the call leg; and `compose-refusal-order.md` / `era-seam.md` no longer claim the factory makes misrouting *unreachable* — being handed the split is what the seam guarantees, using it remains the factory's own responsibility.

## Test plan

- [x] Surface gates driven RED first (`era-load-era`, `dist-laziness`, `protocol-type-acl`), then made green by the implementation
- [x] The #731/#877 regression guards kept at full strength: `routes a coin the partition places in the fallible half` still swaps the real halves, now by intercepting the factory rather than the removed method, and still asserts both directions
- [x] Parity test: both eras hand the factory the same split and answer with it, pinned in both directions so an empty pair cannot satisfy it
- [x] Assembler test: the pair returned is the pair the prototype was built from, proven by recomposing from it and comparing the rendered call
- [x] Orchestration-order tests updated — the offer is now built *inside* the composition (`era.v8.zswapOffer` / `era.v9.zswapOffer`), and the composer receives `kind: 'unpartitioned'`
- [x] **New, RED first:** parity test pinning the unshielded payout refusal ahead of unreadable offer bytes — the one input that can tell the two refusal orders apart. Failed on v9, passed on v8, before the fix
- [x] **New, RED first:** v9 test pinning one partition per call in `calls` order, tied by object identity rather than length, with the factory invoked exactly once per composition. Mutation-checked: `push` -> `unshift` makes it fail
- [x] **New, RED first:** pipeline test pinning the refusal of an era that composes without ever asking for the offer. Failed with "promise resolved instead of rejecting" before the fix
- [x] `yarn lint` clean, `yarn build` succeeds (16/16)
- [x] `yarn turbo test --force` — 31/31 tasks, `Cached: 0 cached`

### CI note

The 7 `Hard fork / *` jobs are red. This branch did not cause it, and — contrary to what the workflow's own comment suggests — it is **not** the lane's intended red state either. It is a live regression, tracked in #1313.

The lane ran **7/7 green on `main`** on 2026-09-11, both at `4b2d912c` (the #1218 merge) and at `98ab4ba7`. From 15:26 UTC that same day every run has been 0/7, across three unrelated branches, with only documentation commits landing on `main` in between.

Root cause is a dependency drift outside this repository's control: `@midnightntwrk/ledger-v9` moved from `1.0.0-rc.4` to `1.0.0-rc.5`, and `@midnight-ntwrk/compact-js` ranges over it with `^1.0.0-rc.3`. The consumer-e2e persona installs into a separate Yarn PnP project where this repository's `resolutions` do not apply, and `consumer-e2e/personas.mjs` carries pins across for `ledger-v8`, `compact-runtime` and `onchain-runtime-v3` but not for `ledger-v9`. The green run's install log contains only `rc.4`; the red runs contain `rc.4` **and** `rc.5`. Two copies means two WASM instantiations, so `partitionAllTranscripts` fails its `instanceof` check with `expected instance of LedgerParameters`.

`main` shows green only because the lane is **skipped** there — it is gated on path filters the nightly schedule does not trip. That is the absence of a run, not a passing run.

Separately, `unshielded.transfer.it.test.ts` was cancelled at its 30-minute timeout while still inside `yarn install`, with node-gyp stalled fetching from nodejs.org. No test ran and no report was produced; that one is genuine infrastructure flake and passes on other branches.

Assisted-by: AI


VanessaPC (approved):

VanessaPC (commented):
Two things from a read of the full diff. The core move is right — inverting the dependency removes a method from a semver-covered interface rather than relabelling it, and it drops a duplicated `QueryContext` + partition that `main` did in the pipeline. The follow-up commit's refusal-order fix is a real regression catch, and the parity test that pins it is the correct minimal input.

Neither comment below blocks; both are about the seam's edges rather than its shape.

_Review analysis was AI-assisted._

packages/protocol/README.md
```diff
@@ -151,7 +151,17 @@ const era = await loadLedgerEra(versionOfRecord(indexerRecord));
...
+const { transaction, partitions } = era.composeCallTx({
+  calls,
+  networkId,
+  ttl,
+  // Called back ONCE, with one `[guaranteed, fallible]` pair per call, in
+  // `calls` order -- cross-contract callees first, the root call last. Route
+  // against all of them: a transaction carries one offer per segment, so
+  // destructuring the first pair alone would route the whole transaction
+  // against a callee's split and drop the root call's.
+  zswapOffer: (partitions) => buildSegmentedOfferBytes(partitions)
```
VanessaPC, line 163: The comment above this line tells callers to "Route against all of them", but the helper this same section names two paragraphs earlier can't do that: `zswapStateToSegmentedOffer` takes a **single** `PartitionedTranscript` as its fourth argument (`packages/contracts/src/internal/utils/zswap-utils.ts:337`), with no list-accepting form.

The repo's own current-era path doesn't do it either — `packages/contracts/src/internal/utils/ledger-utils.ts:198-203` routes against `rootCall.public.partitionedTranscript` alone.

So a reader with a v9 call tree containing a coin-moving callee is instructed to do something there is no supported way to do, and the nearest achievable thing is the one the comment warns against. Given the documented callees-first ordering, what's actually reachable today is the root call's pair — `partitions[partitions.length - 1]`, not `partitions[0]`.

Either describe that, or give `zswapStateToSegmentedOffer` a multi-partition form. Latent for now (`ledger8-pipeline.ts:627` is the only production caller and composes exactly one call), but the doc is currently wrong in the direction that produces the `Wallet.InsufficientFunds` diagnosis it's warning about.

Separately, the snippet shadows `partitions` — the destructured `const` on line 154 vs. the factory parameter here. Legal, but confusing in a doc example.

packages/contracts/src/internal/ledger8-pipeline.ts
```diff
@@ -678,12 +648,58 @@ export const runLedger8CallPipeline = async <TState>(
...
+    }
   });
 
+  // The factory is the ONLY place this pipeline builds its offer, and
+  // `zswapOffer` is optional on the seam -- so an era that never calls back
+  // composes a transaction carrying none of the circuit's coin movements, and
+  // reports offers indistinguishable from a call that legitimately moved none.
+  // The wallet would diagnose that as `Wallet.InsufficientFunds`, naming
+  // neither the call nor the segment. Refused here instead.
+  assertDefined(
```
VanessaPC, line 689: This guard proves the factory was *asked*, not that its answer was *used*, and the gap is reachable the other way round:

- An era that calls the factory, discards the return, and composes with no offer sets `routed` and passes this check. The pipeline then reports `guaranteedZswapOffer` bytes that are not in `txBytes` — the same indistinguishable state this guard was added to prevent.
- An era that calls the factory twice leaves `routed` holding the *last* routing while the transaction may carry the first. Nothing notices.
- The factory routes against its own `partitions[0]` argument, but line 695 reports `composed.partitions[0]`. For both in-repo arms these are the same object, so this only diverges through a third-party `LedgerEra` — but so does the case this guard already covers.

A `calledOnce` flag plus an identity check (`partition === composed.partitions[0]`) would close all three and cost about four lines. Low severity, but the asymmetry stands out given how deliberately the zero-call case is argued in the comment above.

PR #1335, opened 2026-09-17
docs(midnight-js): move design rationale out of API signatures into package docs

### LINKED ISSUE #1334: [Task] Move design rationale out of API signatures and into package docs
### Components

component:midnight-js

### Summary

`packages/contracts/README.md` and `packages/protocol/README.md` both state the rule this task enforces:

> The reasoning behind this package's shape lives in `docs/`, not in the source

Code merged in #1218 and the PRs that followed it does not hold to that rule. A scan of the 106 non-test source files changed between `8545d7a5` and `main` finds:

| Measure | Count |
|---|---|
| Doc-comment blocks in changed sources | 1109 |
| Longer than 12 lines | 226 |
| ... with >= 10 lines of prose before the first `@` tag | 154 |
| ... of those, carrying design rationale (`why this exists`, `rather than`, `by design`) | **49** |
| Prose words in that set | **~7300** |
| Of those 49, already cross-referencing a package doc | **0** |

Length alone is not the defect — much of the 226 is legitimate TSDoc whose length comes from `@param`/`@returns`/`@throws`. The defect is a long rationale preamble standing in front of an API signature, where TypeDoc renders it as the symbol's description.

**Where it sits — packages that already have `docs/` (32 blocks, ~4730 words):**

| File | Blocks | Words | Existing home |
|---|---|---|---|
| `contracts/src/errors.ts` | 8 | 1248 | **none** — no error-taxonomy document exists |
| `contracts/src/internal/ledger8-entry.ts` | 7 | 1078 | partly `keep-state-pipeline.md` |
| `contracts/src/find-deployed-contract.ts` | 2 | 331 | `keep-state-pipeline.md` |
| `contracts/src/ledger8-contract.ts` | 1 | 249 | `keep-state-pipeline.md` |
| `contracts/src/era-results.ts` | 2 | 213 | `overload-typing.md` |
| `protocol/src/lib/v9/operations.ts` | 1 | 241 | **none** |
| `protocol` (4 others) | 4 | 512 | `era-seam.md`, `compose-refusal-order.md` |
| `contracts`/`protocol` remainder | 7 | ~860 | mixed |

**Where the pattern is absent entirely (17 blocks, ~2550 words):** `types` (9), `indexer-public-data-provider` (3), `utils` (3), `fetch-zk-config-provider` (1), `node-zk-config-provider` (1). None of these has a `docs/` directory or `projectDocuments` in its `typedoc.json`.

The two zk-config-provider blocks are the *same* comment duplicated across both packages. The logic they describe belongs to `utils` (`parseZkArtifactRuntimeVersion`), so the rationale goes to `utils/docs/` and both call sites reduce to one line and an `@see`. That removes the duplicate rather than relocating it twice.

### Acceptance criteria / done when

- [ ] `packages/contracts/docs/error-taxonomy.md` exists and carries the rationale currently sitting in the eight `contracts/src/errors.ts` blocks.
- [ ] The blocks in `contracts`/`protocol` whose topic already has a document are folded into that document; the source keeps the API contract plus an `@see`.
- [ ] `packages/protocol/docs/` gains a document for `reexpressOperationsForCurrentEra`.
- [ ] `types`, `utils` and `indexer-public-data-provider` each gain `docs/*.md`, `"projectDocuments": ["docs/*.md"]` in `typedoc.json`, and the README table the other packages use.
- [ ] Every relocated block leaves behind what a caller needs at the signature. Where a comment opens with a contract (for example `ledger8-contract.ts` `signingKey`: *"VALIDATED BEFORE IT IS STORED"*, or `Ledger8DeployUnconfirmedError`: *"DO NOT DEPLOY AGAIN"*), that sentence stays; only the rationale behind it moves.
- [ ] No type, signature, export or runtime behaviour changes. The diff is `.md`, comments, and two configuration lines per package.
- [ ] `yarn lint` and `yarn build` clean; TypeDoc generates the new documents without warnings.
- [ ] `docs/api` is **not** regenerated in this PR — the bot does that in its own `docs: API documentation update` run.

### Context / notes

- Scanned range: `8545d7a5..main` (#1218 merge `4b2d912c` and everything after).
- Established pattern: 17 documents, 3776 lines, ~31.5k words across `packages/{contracts,protocol,midnight-js}/docs/`, wired through `typedoc.json` `projectDocuments` and indexed by a README table.
- `package.json` in these packages declares `files: ["dist/"]`, so `docs/` is not published to npm. Package size is unaffected.
- `types` is touched, but only in comments, `typedoc.json` and `README.md`. No declaration changes, so this is not a breaking change to the interface package.

sp-io:
## Summary

Closes #1334.

`packages/contracts/README.md` and `packages/protocol/README.md` both state the rule this PR enforces:

> The reasoning behind this package's shape lives in `docs/`, not in the source

Code merged in #1218 and the PRs after it did not hold to that rule. This moves the design rationale out of API signatures and into package documents, leaving the API contract — what a symbol does, its parameters, what it returns, what it throws, and what a caller must not do — at the signature.

| Measure | Before | After |
|---|---|---|
| Doc blocks with >= 10 lines of rationale prose ahead of the first `@` tag | 53 | **13** |
| Words of rationale prose in those blocks | ~8 230 | **~1 420** |
| Of those, cross-referencing a package document | 0 | all |

The 13 that remain are blocks already trimmed to their API contract; they still match the crude keyword scan because words like "because" survive in legitimate contract text. Driving the number to zero would mean deleting things a caller needs.

### New documents (8)

| Package | Document | What it carries |
|---|---|---|
| `contracts` | `error-taxonomy.md` | Which class to catch across both eras, which errors register a code and which deliberately do not, the three refusals that fire before a caller pays, the after-submission region the signing key rides through |
| `protocol` | `cross-era-operation-registry.md` | What actually crosses the fork boundary when a dormant retained-era contract is called, and why re-expressing its operations is faithful rather than a substitution |
| `types` | `version-tagged-payloads.md` | Why a retained-era payload crosses a seam as bytes, and the three things the `version` tag deliberately does not establish |
| `types` | `seam-era-declarations.md` | Why `supportedEras` is read up front and still not trusted, why the two era refusals stay distinct, what the lifting adapters exist to avoid |
| `types` | `reading-the-head-era.md` | The caching rule `queryLatestProtocolVersion` imposes, and why it takes no force-refresh option |
| `utils` | `error-vocabulary.md` | Why the framework-code and foreign-code guards are separate, and why foreignness is enforced twice |
| `utils` | `artifact-runtime-version.md` | Why absence of `runtime-version` is a refusal, and why the integrity manifest is consulted before the compiler description |
| `indexer-public-data-provider` | `subscription-shapes.md`, `error-boundaries.md` | What each `contractStateObservable` branch costs on the wire and why the two shapes cannot be collapsed; why `IndexerError` is not exhaustive over a read |

Existing documents extended: `keep-state-pipeline.md`, `overload-typing.md`, `era-dispatch.md`, `breadcrumbs.md`.

### Three stale claims found and corrected

Moving the prose surfaced three places where a docstring or a document had fallen out of step with the code. Each is corrected rather than relocated as-is:

1. `keep-state-pipeline.md` said the attach arm "diverges from the current era in one case". The code diverges in **three** — the doc predated the supplied-key validation and the unreadable-entry rule.
2. `breadcrumbs.md` said "Three decisions get a breadcrumb". There are **four**; `RetainedSigningKeyEntryBreadcrumb` was added later and never reached the document.
3. `ledger8-entry.ts` documented `assertLedger8TxSucceeded` as raising "a bare `Error`", explaining at length why no proper class was available. It has raised `Ledger8CallTxFailedError` for some time — its own `@throws` said so.

### Pattern extended to three more packages

`types`, `utils` and `indexer-public-data-provider` had no `docs/` directory and no `projectDocuments`. Each now gets the same three pieces the other packages have: the documents, one line in `typedoc.json`, and the README table. TypeDoc generates their `documents/` pages accordingly.

The duplicated `getArtifactRuntimeVersion` rationale in `fetch-zk-config-provider` and `node-zk-config-provider` is **removed rather than relocated twice**: the logic belongs to `utils` (`parseZkArtifactRuntimeVersion`), so the reasoning lives there once and both providers keep their transport-specific contract plus a pointer.

### Not in scope

- **`docs/api` is not regenerated here.** The bot does that in its own `docs: API documentation update` run. Generating it locally confirmed all 8 new documents render and TypeDoc exits 0.
- **Pre-existing TypeDoc warning, untouched.** `@see {@link Title}` resolves inside a package's own project but not in the `midnight-js` barrel, which re-documents the same symbols without the sub-packages' `projectDocuments`. This already affects every existing document — `ComposeRefusalOrder`, `VerifierKeys`, `FailClosedDecoding`, `EraSeam` and others all warn on `main` today. The new documents behave identically. Fixing it means restructuring the barrel's TypeDoc project and belongs in its own change.

## Test plan

- [x] No type, signature, export or runtime behaviour changes — the diff is `.md`, comments, and one configuration line per package
- [x] All existing tests pass (no regressions): `yarn test` — 31/31 tasks, every touched package a cache miss, so each actually ran
- [x] Lint clean: `yarn lint` and `eslint` over the changed files explicitly, both exit 0
- [x] Build succeeds: `yarn build` — 16/16
- [x] TypeDoc generates the 8 new documents without error (exit 0)

No tests were written first, because nothing testable changed. The verification for a documentation move is that the behaviour gates stay green and the generated pages render.


VanessaPC (approved):

sp-io:
recheck

sp-io:
recheck

PR #1337, opened 2026-09-21
feat(testkit-js): upgrade wallet-sdk to the 2.0.0 canary and adopt WalletFacade.adoptTransaction

sp-io:
# Overview

Moves `@midnightntwrk/wallet-sdk` from `2.0.0-beta.3` to
`2.0.0-canary.20260917091532-7025e69`, and adopts the API that canary brings:
[`WalletFacade.adoptTransaction(bytes, stage)`](https://github.com/midnightntwrk/midnight-wallet/pull/760).

## What the upgrade actually changes

I diffed the published `.d.ts` of all 15 wallet-sdk packages, beta.3 against the
canary. The complete public-surface delta:

| Package | Change |
| --- | --- |
| `facade` 5.0.0 | **+ `adoptTransaction(bytes, stage): WalletTransaction<TStage>`** |
| `shielded`, `dust-wallet`, `unshielded-wallet` | `OutOfOrderSyncUpdateError` added to the `WalletError` union; `unshielded`'s `applyUpdate`/`applyFailedUpdate` widened with `UtxoNotFoundError` |
| `abstractions`, `hd` | documentation only |
| the other nine | no change |

Ledger dependencies are untouched (`ledger-v9@1.0.0-rc.4`, `ledger-v8@^8.1.0`),
so the root `resolutions` stay as they are. Nothing in this repository switches
exhaustively over `WalletError`, so the widened unions cost nothing. The
upgrade is non-breaking here.

`packages/utils` moves its direct `wallet-sdk-address-format` dependency in the
same step. Left behind it would resolve a beta alongside the canary, putting two
copies of that package — and its WASM — in the tree.

## Where PR 760 applies, and where it does not

`DAppConnectorWalletAdapter` held exactly the "before" that PR removes, in three
methods. It deserialized with the **ledger-v9 codec specifically** and then
stamped the handle with the facade's *active* protocol version — so below the v9
fork the codec and the stamp disagreed. `adoptTransaction` answers both from one
version, which is a fix, not just a tidy-up.

It does **not** apply to `MidnightWalletProvider` or the `adoptVersioned*`
helpers. Those sit at the version-tagged provider seam (ADR 0006, ADR 0014):
the payload names its own era, so the era comes from the tag, never from the
facade's current version. `adoptTransaction` always reads at the facade's
version and would silently misread a `v8`-tagged payload. Left alone
deliberately.

## Removed as dead

`adoptUnbound` and `adoptFinalized` lost their only caller. `unwrapFinalized`
had **no** production caller before this PR — only a test kept it alive. All
three are internal (`wallet-transaction.ts` is not in the package barrel), so
nothing outside the package sees this.

`test/wallet-transaction.ut.test.ts` tested only those three and is deleted; the
versioned seam it sat next to is already covered end-to-end through
`midnight-wallet-provider.ut.test.ts` (both eras, both seams, untagged refusal).

## The eslint commit

Unrelated to the SDK, but it is what let the pre-push hook run rather than be
bypassed. `.gitignore` already excludes `.worktrees`, but eslint walked it:
every checkout carries its own `tsconfig.json`, the type-aware parser saw
several candidate roots, and **every file in the repository** failed to parse —
11031 errors, so `yarn lint` could not pass for anyone using in-repo worktrees.

## Submission Checklist

- [x] Self-reviewed the diff
- [ ] Reviewer requested
- [ ] Update README.md file (if relevant)
- [ ] Update documentation (if relevant)
- [x] No new todos introduced

## Verification

- `yarn ut` in `testkit-js/testkit-js` — **156 passed, 0 failed** (two
  `fixtures-hf` tests fail on an unbuilt `dist/`, on this branch and on `main`
  alike; they pass after `yarn build:testkit`).
- `yarn typecheck:tests` — clean, core and testkit.
- `yarn lint` — clean across the whole repo, after the eslint commit above.
- `yarn constraints` — clean.
- **Not run:** the e2e suite, which needs Docker.

## Note for reviewers

The dependency is a **canary** build. Worth pinning to a release tag before this
reaches anything long-lived.


testkit-js/testkit-js/src/wallet/dapp-connector-wallet-adapter.ts
```diff
@@ -102,28 +100,25 @@ export class DAppConnectorWalletAdapter implements ConnectedAPI {
   }
 
   async balanceUnsealedTransaction(tx: string, options?: { payFees?: boolean }): Promise<{ tx: string }> {
-    const unboundTx = LedgerTransaction.deserialize<SignatureEnabled, Proof, PreBinding>('signature', 'proof', 'pre-binding', fromHex(tx));
     const tokenKindsToBalance = options?.payFees === false ? (['shielded', 'unshielded'] as ('shielded' | 'unshielded')[]) : ('all' as const);
     const recipe = await this.walletProvider.wallet.balanceUnboundTransaction(
-      await adoptUnbound(this.walletProvider.wallet, unboundTx),
+      this.walletProvider.wallet.adoptTransaction(fromHex(tx), 'Unbound'),
```
agronmurtezi: Looks good, but you may want to do this a step earlier and catch/handle the error it throws

sp-io: Done in adad4359, and applied to all three entry points that read dApp bytes rather than just this one — `balanceSealedTransaction` and `submitTransaction` had the same shape.

The read is now its own step in a private `adopt(tx, stage)` helper ahead of the wallet call, and the refusal is caught there. The connector API already has the right vocabulary for it: `ErrorCodes.InvalidRequest` is documented as covering "a malformed transaction", which is exactly what bytes authored on the other side of a protocol boundary are from here. So callers now get an `APIError` (`type: 'DAppConnectorAPIError'`, `code: 'InvalidRequest'`) with the wallet SDK's `WireFormatError` kept as `cause`, instead of the SDK error leaking through.

That also closes a fidelity gap in this test double specifically: PR 760 leaves error mapping to the wallet, so a real wallet would answer `InvalidRequest` here. Without the mapping a dApp could pass against this adapter and fail against a real wallet.

Covered by a table-driven test over the three entry points asserting the code, the preserved cause, and that the wallet is never reached.

agronmurtezi (approved):

PR #1341, opened 2026-09-22
build(midnight-js): adopt compact-js 3.0.0-rc.0, and fix two defects the move exposed

sp-io:
# Overview

Moves `packages/protocol` onto `@midnight-ntwrk/compact-js@3.0.0-rc.0` — the release that adds
era-pinned `/v8` and `/v9` entry points — and fixes two defects the move exposed.

**No era-execution logic moves.** `packages/protocol/src/lib/v8/` is byte-identical to `main`.
That is deliberate and enforced as an acceptance criterion: compact-js cannot yet supply
everything that layer needs, and the four gaps are filed upstream (see *Phase 2*). Retiring
`execute.ts` and `down-convert.ts` is phase 2's job, not this PR's.

Design: `docs/specs/2026-09-22-compact-js-v3-era-adoption-design.md`.
Plan: `docs/plans/2026-09-22-compact-js-v3-adoption-phase-1.md`.
(Both untracked, per this repo's convention for plans and specs.)

## What this changes

**1. The dependency bump, with the installed-tree invariant under test.**
compact-js 3.0.0-rc.0 declares `@midnightntwrk/ledger-v9@^1.0.0-rc.4`; installed on its own that
caret resolves to `rc.5`, while this repo pins `rc.4`. Two physical copies of ledger-v9 in one
process is the fault `lib/v8/instance-guard.ts` exists to detect on a different axis, so the new
`single-instance.test.ts` turns "the root `resolutions` pins should win" into something the build
checks. It asserts one `ledger-v9` at `1.0.0-rc.4`, and **two** `compact-runtime` copies —
`0.16.0` (the ledger-8 era's runtime, via the `compact-runtime-ledger8` alias) and `0.19.0` — so
it pins the deliberate duality while catching a *third*.

**2. Two test assertions that passed for exactly the values they existed to exclude.**
`packages/protocol`'s ledger-era facade promises only plain data crosses it. Two suites asserted
that as `expect(() => structuredClone(x)).not.toThrow()`. That is unsound: `structuredClone` on a
WASM-backed instance does **not** throw — it returns `{ __wbg_ptr: <n> }` with every field gone.
Replaced with a structural own-property walk, plus a test that the new assertion rejects what the
old one accepted.

**3. Circuit arguments were unchecked on the current era's public call options.**
compact-js's `ProvableCircuitId<C, K>` is `K & Brand<'ProvableCircuitId'>`, and
`Contract.CircuitParameters<C, K>` indexing with that intersection degrades to `unknown[]`.
`getProvableCircuitIds()` produces branded ids, so real call sites hit it. Fixed locally by
unbranding the lookup key (`Brand.Brand.Unbranded`) in `call.ts` and `tx-interfaces.ts`;
`circuitId` keeps its brand, which is what still rejects an arbitrary string.

## Breaking change (type-level)

Code that passed a **wrongly-typed circuit argument** through `CallOptionsWithArguments` /
`CallTxOptions` instantiated with a branded circuit id compiled before and does not now. It was
never correct at run time. `CircuitCallTxInterface` — which types `callTx` — keys with the plain
unbranded id and was already precise; it is unchanged.

**Correction to an earlier revision of this description.** It claimed `circuitId` keeps its brand "which is what still rejects an arbitrary string". That is wrong, and review caught it: `Contract.ProvableCircuitId<C>` as used in these signatures is the *namespace* member (`keyof C['provableCircuits'] & string`), not the branded module-level export — so a plain unbranded literal is accepted, and it is the `keyof` constraint that does the rejecting. The brand is incidental to whatever a caller instantiates the parameter with. The fix itself is unaffected; only the explanation was wrong.

No runtime behaviour changes anywhere in this PR. The only production edits are a type alias in
`call.ts` and explicit type arguments plus a retargeted cast in `tx-interfaces.ts`.

## One thing reviewers should weigh

`packages/protocol` re-exports compact-js through bare `export *` in three published subpaths
(`./compact-js`, `./compact-js/effect`, `./compact-js/effect/Contract`). `protocol-acl.test.ts`
asserts only that those subpaths resolve and are non-empty, so a **major** compact-js bump flows
through them unchecked.

This PR demonstrates the surface did move: `overloads.test-d.ts` had to be rewritten because
compact-js 3 makes `Contract.Any` accept a retained-era contract (`Awaitable<A> = A | Promise<A>`)
and `CircuitContext` is now `any`. So "no published surface changed" is true of *this repo's own
declarations*, not of what `@midnight-ntwrk/midnight-js-protocol` publishes downstream. Worth a
decision on whether those subpaths should gain pinned surface tests.

The replaced type-level guard is not a safety loss: era selection is a **runtime** decision in
`resolveArtifactEra`, which reads the artifact's shape and the compact-runtime version its own
artifacts declare, and is covered by `era-dispatch.test.ts` and `era-dispatch-ledger8.test.ts`.

## Phase 2 — blocked upstream

| deferred | blocked on |
|---|---|
| retiring `lib/v8/execute.ts` | `block` / `comIndices` absent from compact-js's public result — midnightntwrk/midnight-sdk#400, with #403 |
| retiring `lib/v8/down-convert.ts` | `CompactRuntime` exports no `StateValue` / `ChargedState` value, and ledger-v8's are a physically distinct WASM instance |
| moving `executeConstructor` | `initialize` also registers verifier keys, needs a `ZKConfiguration` layer `protocol` does not take, and sources the signing key differently |
| dropping `protocol`'s direct ledger-v8 dependency | `Ledger` facade lacks `Transaction` / `ZswapOffer` / `ContractCall` / `CostModel` — midnightntwrk/midnight-sdk#401 |

Upstream issues opened while evaluating this adoption: midnightntwrk/midnight-sdk#400, #401,
#402, #403.

## Known limits, accepted deliberately

- The guard-the-guard test uses a hand-built collapsed-handle shape rather than a live nested WASM
  instance. Faithful given own-property semantics; the test and variable are named so no reader
  concludes otherwise.
- `canonical()` in `clone-assertions.ts` has no `Set` handling, throws on circular references, and
  inherits JSON's `-0` / `NaN` / `undefined` behaviour. Correct at all four call sites; the limits
  are documented on the helper.

## Submission Checklist

- [x] Self-reviewed the diff
- [ ] Reviewer requested
- [x] Update README.md file (if relevant) — not relevant
- [x] Update documentation (if relevant) — spec and plan accompany the branch

## Test plan

Run on the final tree, cache bypassed (`yarn test --force`): **31/31 turbo tasks, 0 cached, 0
failures.** `yarn build`, `yarn lint ./packages` and `yarn typecheck:tests` clean.

`export-surface.test.ts`, `protocol-acl.test.ts` and `protocol-type-acl.test.ts` pass
**unmodified** — they were not touched, which is how "no declared surface changed" is checked
rather than asserted.

## Links

Relates to midnightntwrk/midnight-sdk#387, #388 (the era-packaging work this consumes).


packages/contracts/src/test/typecheck/call-args.test-d.ts
```diff
@@ -0,0 +1,178 @@
...
+    // `CoinReceiver016Coin` is read off the generated circuit signature rather than restated,
+    // so this cannot drift from the artifact.
+    expectTypeOf<
+      CallOptionsWithArguments<CoinReceiver016Contract, 'receive_coin'>['args']
+    >().toEqualTypeOf<[CoinReceiver016Coin]>();
+  });
+
+  it('omits args entirely for a zero-argument circuit', () => {
+    expectTypeOf<CallOptionsWithArguments<Counter016Contract, 'increment'>>().not.toHaveProperty(
+      'args'
```
VanessaPC: This is the one cell of the matrix that doesn't discriminate. The key here is a **literal**, so no brand ever enters the computation — exactly the trap the comment at L63-70 calls out for the other arm. This assertion passes identically with and without `CircuitKey`.

That matters because the file's own header states the failure mode as `args` staying present "on every circuit, **argument-taking or not**" — but every branded assertion lives on the argument-taking side. Revert the fix and `CallOptionsWithArguments<Counter016Contract, BrandedIncrementId>` gains a spurious *required* `args: unknown[]` (since `unknown[] extends []` is false, the conditional takes the wrong arm), and nothing here catches it.

One mirror of L71-75 closes it:

```ts
it('omits args for a zero-argument circuit at the BRANDED id too', () => {
  type BrandedIncrement = ProvableCircuitId<Counter016Contract, 'increment'>;
  expectTypeOf<CallOptionsWithArguments<Counter016Contract, BrandedIncrement>>()
    .not.toHaveProperty('args');
});
```

packages/contracts/src/tx-interfaces.ts
```diff
@@ -47,14 +47,27 @@ export type CircuitCallTxInterface<C extends Contract.Any> = {
 
 /**
  * Creates a {@link CallTxOptions} object from various data.
+ *
+ * `args` is indexed with {@link CircuitKey}-unbranded `PCK`, matching `call.ts`'s
+ * `CallOptionsWithArguments`: by the time a real call site (`circuit()`, `getProvableCircuitIds()`)
```
VanessaPC: This claim doesn't hold, and it contradicts the comment added ~55 lines below it in the same diff.

`createCallTxOptions`'s only in-repo caller is L105, which explicitly instantiates at the **unbranded** `Contract.ProvableCircuitId<C>` — a change this PR makes — and the inline comment at L111 says exactly that. `circuit()` isn't a caller either: the only `.circuit(` in production is `unproven-call-tx.ts:129`, which consumes the options downstream.

The fix is still justified — `createCallTxOptions` is exported from the barrel (`index.ts:157`), so an external consumer instantiating `PCK` with a branded id from `getProvableCircuitIds()` does hit the degradation. That's just a different statement from the one written here. Suggest something like: "a consumer that instantiates `PCK` with a branded id gets `unknown[]` here; this package's own call site passes the unbranded id explicitly."

packages/protocol/src/test/clone-assertions.ts
```diff
@@ -0,0 +1,112 @@
...
+
+const childEntries = (value: object, path: string): readonly PathedEntry[] => {
+  if (Array.isArray(value)) {
+    return value.map((entry, index): PathedEntry => [`${path}[${index}]`, entry]);
+  }
+  if (value instanceof Map) {
+    return Array.from(value).flatMap(([key, entry], index): PathedEntry[] => [
+      [`${path}<map key ${index}>`, key],
+      [`${path}<map value ${index}>`, entry]
+    ]);
```
VanessaPC, line 42: This branch is load-bearing for the same reason the `Set` branch below it is — `Object.entries(new Map([[k, v]]))` is `[]`, so a handle behind a `Map` would pass silently under the catch-all at L50 — but unlike `Set` it has no regression test. Delete these five lines and nothing goes red.

It's the more important of the two, because `Map` is in the shape production actually returns (`claimedUnshieldedSpends`, `verifierKeys` — see `era-parity.test.ts:141-145,343`), while `Set` isn't. So the branch with a test guards a shape that doesn't occur and the branch without one guards a shape that does.

Cloning `era-parity.test.ts:717-727` with `new Map([['k', { __wbg_ptr: 1232224 }]])` (expecting `value.leaked<map value 0>`) plus a handle-as-key case would cover both halves the walk emits.

packages/protocol/src/test/shared-contract-state.test.ts
```diff
@@ -143,15 +144,26 @@ describe('decodeContractStateWith', () => {
...
+  // -- both degrade to `{ __wbg_ptr: <number> }` with every real field gone
+  // (see `clone-assertions.ts`). It rejects a handle by walking the value's own
+  // structure for that pointer property at any depth.
   it('returns plain data a structured clone can carry across a boundary', () => {
     const pojo = decodeContractStateWith(readHexFixture('state-migrated-v9.hex'), 'v9', ledgerV9);
 
     expect(Object.getPrototypeOf(pojo)).toBe(Object.prototype);
-    expect(() => structuredClone(pojo)).not.toThrow();
+    expectStructuredCloneable(pojo);
     for (const entry of pojo.entryPoints) {
```
VanessaPC, line 170: This loop can iterate zero times and assert nothing. The comment above claims the checks reach "the result, and each `entryPoints` member", which holds only for a non-empty array — if the decoder or `state-migrated-v9.hex` regressed to zero entry points, this half is silently vacuous and `expectStructuredCloneable(pojo)` still passes (`[]` clones fine).

The untouched sibling test at L49, on the same fixture and immediately before an identical `for…of`, already guards this with `expect(pojo.entryPoints.length).toBeGreaterThan(0);`. Worth matching here.

VanessaPC (commented):
Review of the type fix and the new test surface. The core mechanic checks out — I verified `Brand.Brand.Unbranded<K>` is an exact identity on an unbranded `K` (no widening) against the pinned `effect@3.22.1`, and the delete-on-red pin at `call-args.test-d.ts:117` genuinely flips. Encapsulation of `CircuitKey` is enforced by `contracts-type-acl.test.ts`, not just intended. Four inline comments below; two further items can't anchor to a diff line:

**1. The "Known limits" bullet in the PR description describes code that was deleted mid-PR.** It documents `canonical()` as having no `Set` handling, throwing on circular references, and inheriting JSON's `-0`/`NaN`/`undefined` behaviour. The shipped `clone-assertions.ts` is 112 lines with an explicit `Set` branch, a `WeakSet` cycle guard, and no content comparison at all — so all three caveats are the opposite of what landed. The helper and its doc comment are fine; it's the description that's stale, and it's the first thing a reviewer reads.

**2. ADR-0010 now contradicts the code.** Rule 2 still reads that the `structuredClone` gate "DOCUMENTS the rule rather than mechanising it… Closing that is its own change; until then the rule rests on review, not on the gate." This PR *is* that change. Left unamended, a future reader is told the gate is decorative and may delete it — and AGENTS.md treats Accepted ADRs as binding.

Also worth a decision, not a blocker: `single-instance.test.ts` reads the root `package.json` `resolutions` and the yarn resolution graph, but neither the root manifest nor `yarn.lock` is in the `test` task's `inputs` or in `globalDependencies`. Whether Turborepo's built-in lockfile hashing already covers this I couldn't settle without an install — but `turbo.json`'s own comment block establishes the convention of declaring root-file gates explicitly and naming each one, so this new gate arguably belongs on that list regardless.

packages/contracts/src/test/typecheck/call-args.test-d.ts
sp-io: Fixed in e3ff187. Added the branded mirror; the literal-keyed one is renamed to say LITERAL, matching its neighbour at L57.

Checked the discrimination both ways: reverting the unbranding in `call.ts` now fails `omits args for a zero-argument circuit at the BRANDED id too` (alongside the two branded assertions that already caught it) while the literal-keyed omission assertion stays green — which was your point exactly.

packages/contracts/src/tx-interfaces.ts
sp-io: Correct, and the contradiction was inside one diff. Rewritten in bf76600.

The doc now says what you suggested: the function is exported from the barrel (`index.ts:157`), so a **consumer** instantiating `PCK` with a branded id from `getProvableCircuitIds()` is who hits the degradation, and this package's own call site passes the unbranded `Contract.ProvableCircuitId<C>` explicitly — with a pointer to the comment 55 lines below that already said so.

packages/protocol/src/test/clone-assertions.ts
sp-io: Confirmed and fixed in e3ff187. Deleting L38-43 left all 42 era-parity tests green, exactly as you said.

Added both halves the branch emits, since it enumerates keys and values: a handle as a map value (`value.leaked<map value 0>`) and one used as a map key (`value.leaked<map key 0>`). Re-ran the deletion check: removing the branch now fails those two tests and nothing else.

packages/protocol/src/test/shared-contract-state.test.ts
sp-io: Fixed in e3ff187 — `expect(pojo.entryPoints.length).toBeGreaterThan(0);` before the loop, matching the sibling at L49, with a one-line comment saying why an empty array would otherwise pass.

sp-io:
The three items that couldn't anchor to a diff line are done too. All four inline threads have replies.

**1. Stale "Known limits" bullet.** You're right that all three caveats describe the deleted `canonical()`, not what shipped. The description now states what `expectStructuredCloneable` actually does and doesn't reach — own enumerable properties plus `Map`/`Set` members and array elements; nothing behind a non-enumerable own property or a prototype getter; no content comparison at all — with a short note recording that the old bullet described a helper that no longer exists, so the correction is visible rather than silent.

**2. ADR-0010.** Amended in bf76600, per AGENTS.md: an appended `## Amendment — the transport rule is now mechanised, not merely documented (2026-09-23)` section, original sections untouched, no `Amended:` header. It quotes the sentence that no longer holds, records what the walk enforces now, keeps the three things that did *not* change (`structuredClone` is still called for its own throw; the rule itself is unchanged; the gate is a test, not a type), and narrows "rests on review, not on the gate" to the shapes the walk genuinely cannot reach rather than deleting it.

**3. turbo inputs for `single-instance.test.ts`.** Settled empirically, and the gate was leaky. Before the change, the `test` task hash for `midnight-js-protocol` was `d66d6b2f43b32228` and stayed at `d66d6b2f43b32228` after editing the root `package.json` — so a `resolutions` edit that was never reinstalled, the exact case that gate exists to catch, would have replayed a cached PASS. Turborepo's lockfile hashing covers `hashOfExternalDependencies`, which is not the same thing.

Both files are now in `globalDependencies` and appear in `globalCacheInputs.files`; the same edit takes the hash to `26433557b6c08128` and back on restore. Named in the comment block with the other root-file gates, per the convention that block sets.

Re-run on the final tree, cache bypassed: **31/31 turbo tasks, 0 cached, 0 failures.** `yarn build`, `yarn lint ./packages` and `yarn typecheck:tests` clean.

Still open from the description, unchanged by this round and worth a separate decision: the three bare `export *` subpaths (`./compact-js`, `./compact-js/effect`, `./compact-js/effect/Contract`) still pass a major compact-js bump through to downstream consumers with only a resolves-and-is-non-empty check behind them.

VanessaPC (approved):

